PR #2, opened 2014-04-28
Added option to overwrite existing stored credentials.

lraccomando:
Added an option to django_orm.Storage locked_put method to allow for overwriting an existing credential record rather than the creation of a new one.

I needed this functionality on a project I'm working on and ended up subclassing Storage in order to achieve it. Thought it may be useful for others as well. The default behavior remains the same.


oauth2client/django_orm.py
```diff
@@ -116,14 +116,19 @@ def locked_get(self):
         credential.set_store(self)
     return credential
 
-  def locked_put(self, credentials):
+  def locked_put(self, credentials, overwrite=False):
     """Write a Credentials to the datastore.
 
     Args:
       credentials: Credentials, the credentials to store.
```
craigcitro, line 123: Could you add a description of the new argument to the docstring?

oauth2client/django_orm.py
```diff
@@ -116,14 +116,19 @@ def locked_get(self):
...
     """
     args = {self.key_name: self.key_value}
-    entity = self.model_class(**args)
+    
+    if overwrite: 
+      entity, is_new = self.model_class.objects.get_or_create(**args)
```
craigcitro: Did you want to use `is_new`? If not, maybe just call it `unused_is_new` or `_`?

PR #6, opened 2014-05-09
Add support for default credentials.

orestica:
This pull request goes hand in hand with the following pull request: https://github.com/google/google-api-python-client/pull/4

Once these two pull requests are accepted and integrated into their respective repositories, a Google API call will be authenticated as simple as building the service like this:

service = build("compute", "v1", credential=GoogleCredential.get_default_credential())

This will work when the code is being run on a VM inside Google Compute Engine, a Google App Engine application or on any other off-fabric platform (in this last case, a credentials file needs to be put together containing either user or service account credentials and an environment variable pointed to it OR a 'gcloud auth login' command needs to be run before using the default credentials).


oauth2client/default_credential.py
```diff
@@ -0,0 +1,215 @@
...
+    if default_credential_file:
+      return get_default_credential_from_file(default_credential_file, scopes)
+    elif well_known_file:
+      return get_default_credential_from_file(well_known_file, scopes)
+    elif env_name in ('GAE_PRODUCTION', 'GAE_LOCAL'):
+      return get_default_credential_GAE(scopes)
+    elif env_name == 'GCE_PRODUCTION':
+      return get_default_credential_GCE(scopes)
+    else:
+      raise Exception('Either GOOGLE_CREDENTIALS_DEFAULT '
```
craigcitro: We want to raise exceptions from this package, not just base `Exception`

(sample comment)

oauth2client/client.py
```diff
@@ -242,6 +257,211 @@ def from_json(cls, s):
     """
     return Credentials()
 
+  @staticmethod
+  def get_default():
+    """Provides default credentials to be used in authenticating
```
anthmgoogle: If this is the starting point for the online docs, I would include a very breif summary what this does, i.e it uses the GAE or GCE built in service accounts, or it uses the environment variable or it uses the credential provided by "gcloud auth logic"

oauth2client/client.py
```diff
@@ -242,6 +257,211 @@ def from_json(cls, s):
...
+      return _get_default_credential_GCE()
+    else:
+      raise Exception('Either GOOGLE_CREDENTIALS_DEFAULT '
+                      'environment variable must be set or you need '
+                      'to run "gcloud auth login"! '
+                      'Please see https://developers.google.com/'
+                      'accounts/docs/default-credential')
+
+
+def _get_google_credentials_default():
```
anthmgoogle: A nit: I found the name of this method confusing at first before i read the code to discover that it is environment variable path. Perhaps call it _get_environment_variable_credentials or something with environment_variable in it.

oauth2client/client.py
```diff
@@ -242,6 +257,211 @@ def from_json(cls, s):
...
+      raise Exception('File ' + default_credential_file +
+                      ' (pointed by GOOGLE_CREDENTIALS_DEFAULT environment'
+                      ' variable) does not exist!')
+
+
+def _get_well_known_file():
+  """Get the well known file produced by command 'gcloud auth login'."""
+
+  WELL_KNOWN_CREDENTIALS_FILE = "credentials_default.json"
+  CLOUDSDK_CONFIG_WORD = "gcloud"
```
anthmgoogle: In the Java version, Brian Eaton suggested I change this to CLOUDSDK_CONFIG_DIRECTORY.

oauth2client/client.py
```diff
@@ -242,6 +257,211 @@ def from_json(cls, s):
...
+      if len(missing_fields) > 0:
+        _raise_exception_for_missing_fields(missing_fields)
+
+      return OAuth2Credentials(
+          None,
+          client_credentials['client_id'],
+          client_credentials['client_secret'],
+          client_credentials['refresh_token'],
+          None,
+          'https://accounts.google.com/o/oauth2/token',
```
anthmgoogle: This must be used elsewhere in the file. Can it be a constant?

oauth2client/client.py
```diff
@@ -242,6 +257,211 @@ def from_json(cls, s):
...
+        missing_fields.append('client_id')
+      if 'client_secret' not in client_credentials:
+        missing_fields.append('client_secret')
+      if 'refresh_token' not in client_credentials:
+        missing_fields.append('refresh_token')
+      if len(missing_fields) > 0:
+        _raise_exception_for_missing_fields(missing_fields)
+
+      return OAuth2Credentials(
+          None,
```
anthmgoogle: Minor: this call might be a good place to use named arguments as many of the arguments are not obvious from context.

oauth2client/client.py
```diff
@@ -242,6 +257,211 @@ def from_json(cls, s):
...
+      if 'private_key' not in client_credentials:
+        missing_fields.append('private_key')
+      if len(missing_fields) > 0:
+        _raise_exception_for_missing_fields(missing_fields)
+
+      return service_account.ServiceAccountCredentials(
+          client_credentials['client_email'],
+          client_credentials['private_key_id'],
+          client_credentials['private_key'],
+          [])
```
anthmgoogle: Named argument just for the scopes might be more readable.

oauth2client/client.py
```diff
@@ -242,6 +257,211 @@ def from_json(cls, s):
...
+      if 'client_email' not in client_credentials:
+        missing_fields.append('client_email')
+      if 'private_key_id' not in client_credentials:
+        missing_fields.append('private_key_id')
+      if 'private_key' not in client_credentials:
+        missing_fields.append('private_key')
+      if len(missing_fields) > 0:
+        _raise_exception_for_missing_fields(missing_fields)
+
+      return service_account.ServiceAccountCredentials(
```
anthmgoogle: In the Java version we do check for an pass in the "client_id" in addition to the email. The Java class does actually have inputs for it, although it does not seem to do anything interesting with them other than make them visible on the object. We should perhaps discuss this in the next meeting.

oauth2client/client.py
```diff
@@ -242,6 +257,211 @@ def from_json(cls, s):
...
+          client_credentials['client_email'],
+          client_credentials['private_key_id'],
+          client_credentials['private_key'],
+          [])
+  except Exception as ex:
+    if google_credentials_default:
```
anthmgoogle: This feels like an implementation detail of the calling method bleeding into the interior method. It may be more maintainble to have this catch in the calling method. This also allows implementing from_stream if we decide that it needs to be public.

oauth2client/service_account.py
```diff
@@ -0,0 +1,130 @@
...
+class ServiceAccountCredentials(AssertionCredentials):
+  """Class representing a service account (signed JWT) credential."""
+
+  GOOGLE_REVOKE_URI = 'https://accounts.google.com/o/oauth2/revoke'
+  GOOGLE_TOKEN_URI = 'https://accounts.google.com/o/oauth2/token'
+
+  MAX_TOKEN_LIFETIME_SECS = 3600 # 1 hour in seconds
+
+  def __init__(self,
+      service_account_email,
```
anthmgoogle: As mentioned, we should discuss whether we shoul add service_account_id here.

oauth2client/client.py
```diff
@@ -242,6 +257,211 @@ def from_json(cls, s):
     """
     return Credentials()
 
+  @staticmethod
```
anthmgoogle: As per the spec, I was expecting this on a new class in this same file called GoogleCredentials inherited from Credentials.

samples/call_compute_service.py
```diff
@@ -0,0 +1,16 @@
+# To be used to test GoogleCredential.GetDefaultCredential()
+# from local machine and GCE.
+
+from googleapiclient.discovery import build
+from oauth2client.client import Credentials
+
+PROJECT = "bamboo-machine-422"  # Provide your own GCE project here
+ZONE = "us-central1-a"          # Put here a zone which has some VMs
+
+service = build("compute", "v1", credentials=Credentials.get_default())
```
anthmgoogle: Expecting this to be GoogleCredentials.get_default()

oauth2client/service_account.py
```diff
@@ -0,0 +1,130 @@
...
+
+from oauth2client import util
+from oauth2client.anyjson import simplejson
+from oauth2client.client import AssertionCredentials
+
+from pyasn1.codec.ber import decoder
+from pyasn1_modules.rfc5208 import PrivateKeyInfo
+
+
+class ServiceAccountCredentials(AssertionCredentials):
```
anthmgoogle: It was expected that this class inherit from a new GoogleCredentials class. If AssertionCredentials can have that injected as a parent and that is not a breaking change in Python, that should do the trick.

oauth2client/client.py
```diff
@@ -242,6 +257,211 @@ def from_json(cls, s):
...
+
+    service = build('discovery', 'v1')
+    resource = service.apis()
+    request = resource.list()
+    response = request.execute()
+
+    print response
+    </code>
+    """
+
```
anthmgoogle: The java implementation caches this in a static variable, and the non-functional requirements indicate that repeated calls should not need to recreate the instance, re-parse files, etc. This also requires simple locking on the fetch. This is non-trivial to test. I Java I did this by putting all the static state into an internal class only visible to the tests, and this also makes the tests thread-safe if run concurrently.

tests/test_oauth2client.py
```diff
@@ -100,12 +101,21 @@ def set(self, key, value, namespace=''):
```
anthmgoogle, line 115: Some potential gaps in the testing: it would be good to test the file format parsing, and exposing and testing _from_stream should enable this, as that part of the call would not touch the file system, environment variables, etc. Coverage of the error cases is important too.

There seems to be no coverage of the environment variable or well known file discovery paths, including errors. These are not straight-forward to test so, it might not be inconsistent with the depth of the testing that exists already to skip or defer these. In Java I did this with an internal class holding both the state and virtual methods to allow mocking things like checking for files or reading environment variables.

oauth2client/client.py
orestica: Done. I made OAuth2Credentials inherit from GoogleCredentials, instead of Credentials. Hopefully this won't break the backwards compatibility.

oauth2client/client.py
orestica: I moved the comment to the newly created GoogleCredentials class' level. I put a link to https://developers.google.com/accounts/docs/default-credentials for further details.

oauth2client/client.py
orestica: I moved _env_name variable at the module level so it is static for all practical purposes.

oauth2client/client.py
orestica: I changed it to _get_environment_variable_file().

oauth2client/client.py
orestica: Done.

oauth2client/client.py
orestica: Done.

oauth2client/client.py
orestica: Done.

oauth2client/client.py
orestica: I added the dummy client_id in the game.
Do you want me to add a getter on the ServiceAccountCredentials class for it?

oauth2client/client.py
orestica: Added for all of them, to be consistent.

oauth2client/client.py
orestica: You are spot on.
I moved it one level up.

oauth2client/service_account.py
orestica: I introduced GoogleCredentials inheriting from Credentials and made OAuth2Credentials inherit from GoogleCredentials. As AssertionCredentials inherits from OAuth2Credentials, we should be good.

oauth2client/service_account.py
orestica: I already introduced it.
Let's discuss whether we need to remove it now. :)

samples/call_compute_service.py
orestica: It is now.

tests/test_oauth2client.py
```diff
@@ -526,7 +562,7 @@ def test_exchange_id_token_fail(self):
     self.assertRaises(VerifyJwtTokenError, self.flow.step2_exchange,
       'some random code', http=http)
 
-  def test_exchange_id_token_fail(self):
+  def test_exchange_id_token_fail_encore(self):
```
orestica: Please expand the code above.
There were two test methods with the same name: test_exchange_id_token_fail.
I thought I would be a good citizen if I differentiated them somehow.

tests/test_oauth2client.py
orestica, line 115: I'll work tomorrow on adding more tests.
And once I do that, I'll push my code so you'll be able to see all the changes I've made at your suggestions.

oauth2client/client.py
```diff
@@ -169,6 +173,21 @@ def apply(self, headers):
       headers: dict, the headers to add the Authorization header to.
     """
     _abstract()
+    
+  def create_scoped_required(self):
+    """Whether this Credentials object is scopeless.
+    
+    create_scoped(scopes) method needs to be called in order to create
+    a proper Credentials object.
```
anthmgoogle: Perhaps "a Credentials object for API calls" instead of "a proper Credentials object".

oauth2client/client.py
```diff
@@ -169,6 +173,21 @@ def apply(self, headers):
       headers: dict, the headers to add the Authorization header to.
     """
     _abstract()
+    
+  def create_scoped_required(self):
```
anthmgoogle: The spec indicates that these should be surfaced on GoogleCredentials, not Credentials. We can discuss when we talk about the surface, but the equivalent for Java is on the Google-specific class.

oauth2client/client.py
```diff
@@ -243,6 +262,246 @@ def from_json(cls, s):
...
+      if ('metadata-flavor' in response and
+          response['metadata-flavor'] == 'Google'):
+        return 'GCE_PRODUCTION'
+      else:
+        return 'UNKNOWN'
+    except httplib2.ServerNotFoundError:
+      return 'UNKNOWN'
+
+
+_env_name = _get_environment()
```
anthmgoogle: This looks like it may have some side effects we might want to prevent. For example, would this do the server call now whenever the Credentials type was touched? It also looks like it might do this query  even if the environment variable or well known file were present. This could be prevented by doing the file checks first, doing this on-demand when get_default is called, and keeping a flag to make sure this query happens only once.

oauth2client/client.py
```diff
@@ -243,6 +262,246 @@ def from_json(cls, s):
...
+    elif _env_name in ('GAE_PRODUCTION', 'GAE_LOCAL'):
+      return _get_default_credential_GAE()
+    elif _env_name == 'GCE_PRODUCTION':
+      return _get_default_credential_GCE()
+    else:
+      raise DefaultCredentialsError('Either GOOGLE_CREDENTIALS_DEFAULT '
+                                    'environment variable must be set or '
+                                    'you need to run "gcloud auth login"! '
+                                    'Please see https://developers.google.com/'
+                                    'accounts/docs/default-credential')
```
anthmgoogle: Note, I changed this url to be plural: default-credentials

oauth2client/client.py
orestica: My intent was to convey the information that a Credentials object of the appropriate class would be returned.
Applied your suggestion.

oauth2client/client.py
orestica: Moved the two new methods to GoogleCredentials class.

oauth2client/client.py
orestica: Good catch.
Done.

oauth2client/client.py
orestica: I modified the code so _get_environment() gets called only once: the very first time the GoogleCredentials.get_default() is called.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+  requires authentication:
+
+  <code>
+  from googleapiclient.discovery import build
+  from oauth2client.client import Credentials
+
+  PROJECT = 'bamboo-machine-422'  # replace this with one of your projects
+  ZONE = 'us-central1-a'          # replace this with the zone you care about
+
+  service = build('compute', 'v1', credentials=Credentials.get_default())
```
anthmgoogle: Should this get GoogleCredentials.get_default()

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+                                    " defined pointing to a file defining the"
+                                    " credentials. See https://developers."
+                                    "google.com/accounts/docs/default-"
+                                    "credentials for details.")
+
+  @staticmethod
+  def from_stream(credential_file):
+    """Create a Credentials object by reading the information from a given file.
+    
+    It returns an object of type _ServiceAccountCredentials/GoogleCredentials.
```
anthmgoogle: I would just say GoogleCredentials here, as the _ServiceAccountCredentials should probably not be referenced in public docs and it should inherit from GoogleCredentials.

oauth2client/client.py
orestica: Yes, it should. Nice catch.
Done.

oauth2client/client.py
orestica: Done.

tests/test_oauth2client.py
```diff
@@ -107,6 +120,314 @@ def test_to_from_json(self):
...
+    self.env_server_software = os.environ.get('SERVER_SOFTWARE', None)
+    self.env_google_credentials_default = (
+        os.environ.get('GOOGLE_CREDENTIALS_DEFAULT', None))
+    self.env_appdata = os.environ.get('APPDATA', None)
+    self.os_name = os.name
+    from oauth2client import client
+    setattr(client, '_computed_env_name', False)
+
+  def tearDown(self):
+    self.reset_env('SERVER_SOFTWARE', self.env_server_software)
```
anthmgoogle, line 151: This pattern will have side-effects that could cause errors if the tests are run multi-threaded or if the there is an error in the tear-down code. This might not be a problem depending on how Python unit tests get run. It may be worth double-checking with Craig Citro/John Asmuth to see if it is OK to have tests that modify the process environment like this. If not, this can be handled by having the product code go through an non-public overridable virtual call for things like checking the environment.

orestica, line 151: Hopefully, Craig will enlighten me how to do this (more) properly, once you give me your OK to get him involved.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+  return _env_name
+
+
+class GoogleCredentials(OAuth2Credentials):
+  """Provides default credentials to be used in authenticating
+  Google APIs calls.
+  
+  The default credentials are being constructed as a function of the environment
+  where the code is being run. More details can be found on this page:
+  https://developers.google.com/accounts/docs/default-credentials .
```
craigcitro: Is there even a draft of this to look at somewhere?

orestica: Yes. Here: https://docs.google.com/a/google.com/document/d/1jvTfUBqBgqVDQI24kzoE_YrAUbl6j4kKUshaSGWw0E0

oauth2client/appengine.py
```diff
@@ -196,6 +197,13 @@ def _refresh(self, http_request):
       raise AccessTokenRefreshError(str(e))
     self.access_token = token
 
+  def create_scoped_required(self):
+    return not bool(self.scope)
```
craigcitro: You can drop the `bool` -- just `not self.scope` works (and is preferred by various style guides).

oauth2client/appengine.py
```diff
@@ -196,6 +197,13 @@ def _refresh(self, http_request):
...
+
+  def create_scoped(self, scopes):
+    return AppAssertionCredentials(scopes,
+                                   **self.kwargs)
```
craigcitro: bump this up one line.

oauth2client/client.py
```diff
@@ -99,6 +99,10 @@ class NonAsciiHeaderError(Error):
   """Header names and values must be ASCII strings."""
 
 
+class DefaultCredentialsError(Error):
+  """Error retrieving the default credentials."""
```
craigcitro: we seem to have a little inconsistency around how we talk about this -- the external docs seem to prefer the slightly unfortunate 'the Default Credentials', so we should probably go with that.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+    _env_name = 'GAE_PRODUCTION'
+  elif server_software.startswith('Development/'):
+    _env_name = 'GAE_LOCAL'
+  else:
+    import httplib2
+    try:
+      if http_request is None:
+        http_request = httplib2.Http().request
+      response, _ = http_request('http://metadata.google.internal')
+      if ('metadata-flavor' in response and
```
craigcitro: Python dictionaries have a `.get()` method that simplifies this:

```
if response.get('metadata-flavor') == 'Google':
```

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
 
-class AssertionCredentials(OAuth2Credentials):
+_computed_env_name = False
+_env_name = None
+
+
+def _get_environment(http_request=None):
+  """Detect the environment the code is being run on."""
+
+  global _computed_env_name
```
craigcitro: Why the second variable? Seems easier to just check for `_env_name` being `None`. (If the thought was "oh performance", remember it's Python.)

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+  if _computed_env_name:
+    return _env_name
+
+  server_software = os.environ.get('SERVER_SOFTWARE', '')
+  if server_software.startswith('Google App Engine/'):
+    _env_name = 'GAE_PRODUCTION'
+  elif server_software.startswith('Development/'):
+    _env_name = 'GAE_LOCAL'
+  else:
+    import httplib2
```
craigcitro: I'd prefer to use `urllib2` in this case, for three reasons:
- First and foremost, I eventually want to drop the dependency on `httplib2` in this library. No point in adding more references.
- As a symptom of why I want to drop it, last week the GCS folks hit some strange issues with `httplib2` in this particular codepath (in apitools, but it's doing the same thing). We didn't dig much, but switching to `urllib2` fixed it.
- Given that we're testing for GCE anyway, I don't see an important use case for the user passing in a custom http instance.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+        _env_name = 'UNKNOWN'
+    except httplib2.ServerNotFoundError:
+      _env_name = 'UNKNOWN'
+
+  _computed_env_name = True
+  return _env_name
+
+
+class GoogleCredentials(OAuth2Credentials):
+  """Provides default credentials to be used in authenticating
```
craigcitro: This should be one line -- I'd say just `Default credentials for use in calling Google APIs.`

oauth2client/appengine.py
orestica: Done.

oauth2client/client.py
craigcitro: Feel free to drop the `.`, which is always awkward with URLs anyway.

oauth2client/appengine.py
orestica: Done. (also in gce.py and account_service.py)

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+  Google APIs calls.
+  
+  The default credentials are being constructed as a function of the environment
+  where the code is being run. More details can be found on this page:
+  https://developers.google.com/accounts/docs/default-credentials .
+
+  Here is an example of how to use the default credentials for a service that
+  requires authentication:
+
+  <code>
```
craigcitro, line 891: I'd rather see this in the samples and have a pointer here; multiple copies of hand-written docs just get out of sync too fast.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+  from googleapiclient.discovery import build
+  from oauth2client.client import GoogleCredentials
+
+  PROJECT = 'bamboo-machine-422'  # replace this with one of your projects
+  ZONE = 'us-central1-a'          # replace this with the zone you care about
+
+  service = build('compute', 'v1', credentials=GoogleCredentials.get_default())
+
+  resource = service.instances()
+  request = resource.list(project=PROJECT, zone=ZONE)
```
craigcitro: I'd at least combine this line with the one above.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+
+  service = build('discovery', 'v1')
+  resource = service.apis()
+  request = resource.list()
+  response = request.execute()
+
+  print response
+  </code>
+
+  Exceptions:
```
craigcitro: This probably belongs on the docstrings of the method(s) that raise it, not here.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+    DefaultCredentialsError: raised when the credentials fail to be retrieved.
+  """
+
+  def __init__(self,
```
craigcitro: Let's pack these a little tighter.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+               client_secret,
+               refresh_token,
+               token_expiry,
+               token_uri,
+               user_agent,
+               revoke_uri=GOOGLE_REVOKE_URI):
+    """Create an instance of GoogleCredentials.
+
+    This constructor is not usually called by the user, instead
+    GoogleCredentials objects are instantiated by the
```
craigcitro: Drop "the".

oauth2client/client.py
orestica: Done. (also in other 5 places in this file)

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+    GoogleCredentials.from_stream() and GoogleCredentials.get_default().
```
craigcitro: silly: and -> or

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+  def create_scoped(self, scopes):
+    """Create a Credentials object for the given scopes.
+    
+    The Credentials type is preserved.
+    """ 
+    return self
+
+  @staticmethod
+  def get_default():
+    """Get the default credentials appropriate for the environment in which the
```
craigcitro: This needs to be a one-liner. Maybe `Fetch the default credentials for this environment`?

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+                                    " Cloud SDK and running 'gcloud auth login'"
+                                    ". Otherwise, the environment variable"
+                                    " GOOGLE_CREDENTIALS_DEFAULT must be"
+                                    " defined pointing to a file defining the"
+                                    " credentials. See https://developers."
+                                    "google.com/accounts/docs/default-"
+                                    "credentials for details.")
+
+  @staticmethod
+  def from_stream(credential_file):
```
craigcitro: This isn't a file, it's a filename.

Relatedly, this would be significantly more useful if it took a file instead of a filename, but that would involve breaking compatibility with the other clients, possibly.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+    Args:
+      credential_file: the path to the file from where the credentials
+        are to be read
+    """
+
+    if credential_file and os.path.isfile(credential_file):
+      try:
+        return _get_default_credential_from_file(credential_file)
+      except (DefaultCredentialsError, ValueError) as error:
+        helping_message = ' (provided as parameter to the from_stream() method)'
```
craigcitro: maybe `helping` -> `help`?

craigcitro: Or `extra_help`?

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+                           ' "gcloud auth login" command)')
+        _raise_exception_for_reading_json(well_known_file,
+                                          helping_message,
+                                          error)
+    elif _env_name in ('GAE_PRODUCTION', 'GAE_LOCAL'):
+      return _get_default_credential_GAE()
+    elif _env_name == 'GCE_PRODUCTION':
+      return _get_default_credential_GCE()
+    else:
+      raise DefaultCredentialsError("The Default Credentials are not available."
```
craigcitro: We should use `'` for consistency here; in addition, I'd vote to start this on the line below, so that we don't use up so much vertical space.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+    if os.path.isfile(default_credential_file):
+      return default_credential_file
+    else:
+      raise DefaultCredentialsError('File ' + default_credential_file +
+                                    ' (pointed by GOOGLE_CREDENTIALS_DEFAULT'
+                                    ' environment variable) does not exist!')
+
+
+def _get_well_known_file():
+  """Get the well known file produced by command 'gcloud auth login'."""
```
craigcitro, line 1050: I have two worries about all these `gcloud` references:
- `gcloud` doesn't cover nearly all the Google APIs (eg Drive); I don't want anyone to think this library isn't useful outside cloud APIs.
- If the command name changes, we have a **lot** of references to update.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+  CLOUDSDK_CONFIG_DIRECTORY = 'gcloud'
+
+  if os.name == 'nt':
+    try:
+      default_config_path = os.path.join(os.environ['APPDATA'],
+                                         CLOUDSDK_CONFIG_DIRECTORY)
+    except KeyError:
+      # This should never happen unless someone is really messing with things.
+      drive = os.environ.get('SystemDrive', 'C:')
+      default_config_path = os.path.join(drive,
```
craigcitro: Will this all fit on one line?

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+                                       '.config',
+                                       CLOUDSDK_CONFIG_DIRECTORY)
+
+  default_config_path = os.path.join(default_config_path,
+                                     WELL_KNOWN_CREDENTIALS_FILE)
+
+  if os.path.isfile(default_config_path):
+    return default_config_path
+  else:
+    return None
```
craigcitro: You don't need this -- falling off the end of a function returns `None`.

oauth2client/client.py
craigcitro, line 1050: And now, reading the body of this function, we're taking quite a dependency on logic about where `gcloud` puts its credentials. Is there a way to depend on this logic over there? (Maybe they could set an env var?)

oauth2client/client.py
orestica: You are right. It's not worth explaining why I chose to use both (it has to do with me learning Python).
I got rid of the _computed_env_name variable.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+def _get_default_credential_from_file(default_credential_file):
+  """Build the default credentials from file."""
+
+  import service_account
+
+  # read the credentials from the file
+  with open(default_credential_file) as default_credential:
+    client_credentials = service_account.simplejson.load(default_credential)
+
+  if ('type' not in client_credentials or
```
craigcitro:
```
if client_credentials.get('type') not in ('authorized_user', 'service_account'):
```

For the record, just depending on strings is going to get us into trouble one day. Let's not worry about that here.

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+      client_credentials['type'] != 'authorized_user' and
+      client_credentials['type'] != 'service_account'):
+    raise DefaultCredentialsError("'type' field should be defined "
```
craigcitro, line 1093: This is a place where `"` is totally legit (because it's a string containing `'`).

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+                                  "(and have one of the 'authorized_user' "
+                                  "or 'service_account' values)")
+
```
craigcitro: Two thoughts:
- Lots of duplication here; any reason we can't collapse it?
- Should some/all of this code live on the classes for the credentials they create?

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+  if client_credentials['type'] == 'authorized_user':
+    missing_fields = []
```
craigcitro: Kudos on the good error message here!

Python has a pretty featureful `set` type; I think it'd help here. For instance:

```
required_fields = set(['client_id', 'client_secret', 'refresh_token'])
missing_fields = required_fields.difference(client_credentials.keys())
if missing_fields:
  ...
```

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+
+def _raise_exception_for_reading_json(credential_file,
+                                      helping_message,
+                                      error):
+  raise DefaultCredentialsError('An error was encountered while reading '
+                                'json file: '+ credential_file +
+                                helping_message + ': ' + str(error))
+
+
+def _get_default_credential_GAE():
```
craigcitro, line 1133: Seems like this could live on the GAE credential class?

oauth2client/client.py
```diff
@@ -820,7 +838,337 @@ def _revoke(self, http_request):
...
+  from oauth2client.appengine import AppAssertionCredentials
+
+  return AppAssertionCredentials([])
```
craigcitro, line 1136: it really doesn't need any args?

oauth2client/gce.py
```diff
@@ -87,4 +88,14 @@ def _refresh(self, http_request):
         raise AccessTokenRefreshError(str(e))
       self.access_token = d['accessToken']
     else:
+      if content.find('Error 404') >= 0:
```
craigcitro: Can't we check against the status code on the response? I don't like grepping error strings unless we have to.

oauth2client/service_account.py
```diff
@@ -0,0 +1,132 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A service account credentials class.
+
+This credentials class is implemented on top of rsa library.
+"""
+
+__author__ = 'orest@google.com (Orest Bolohan)'
```
craigcitro: We don't actually need `__author__` tags anymore.

oauth2client/service_account.py
```diff
@@ -0,0 +1,132 @@
...
+from pyasn1.codec.ber import decoder
+from pyasn1_modules.rfc5208 import PrivateKeyInfo
+
+
+class _ServiceAccountCredentials(AssertionCredentials):
+  """Class representing a service account (signed JWT) credential."""
+
+  MAX_TOKEN_LIFETIME_SECS = 3600 # 1 hour in seconds
+
+  def __init__(self,
```
craigcitro: Let's pack these arguments.

oauth2client/service_account.py
```diff
@@ -0,0 +1,132 @@
...
+    signature = base64.urlsafe_b64encode(rsa.pkcs1.sign(
+        assertion_input, self._private_key, 'SHA-256')).rstrip('=')
+
+    return '%s.%s' % (assertion_input, signature)
+
+  def sign_blob(self, blob):
+    return (self._private_key_id,
+            rsa.pkcs1.sign(blob, self._private_key, 'SHA-256'))
+
+  def get_service_account_email(self):
```
craigcitro: Just make this

```
@property
def service_account_email(self):
  return self._service_account_email
```

and then you have a getter; callers can just say `credential.service_account_email` and it does what you expect.

oauth2client/service_account.py
```diff
@@ -0,0 +1,132 @@
...
+                                      token_uri=self._token_uri,
+                                      revoke_uri=self._revoke_uri,
+                                      **self._kwargs)
+
+def _urlsafe_b64encode(data):
+  return base64.urlsafe_b64encode(
+      simplejson.dumps(data, separators = (',', ':'))\
+          .encode('UTF-8')).rstrip('=')
+
+def _get_pk(private_key_pkcs8_text):
```
craigcitro: let's expand out the name (pk = private key?)

samples/app.yaml
```diff
@@ -0,0 +1,10 @@
+application: bamboo-machine-422
```
craigcitro: actually, why don't we make a folder for the sample GAE app inside `samples/`? it's always frustrating when you don't know which file(s) are part of a bundle like that.

samples/call_compute_service_from_gae.py
```diff
@@ -0,0 +1,24 @@
...
+  resource = service.instances()
+  request = resource.list(project=PROJECT, zone=ZONE)
+  return request.execute()
+
+class MainPage(webapp2.RequestHandler):
+
+  def get(self):
+    self.response.write(get_instances())
+
+app = webapp2.WSGIApplication([
```
craigcitro: one line

setup.py
```diff
@@ -26,6 +26,9 @@
 
 install_requires = [
     'httplib2>=0.8',
+    'pyasn1==0.1.7',
+    'pyasn1_modules==0.0.5',
```
craigcitro, line 30: do we know anything about these modules? taking a dependency on something at version `0.0.5` worries me; have we at least forked them ourselves in case they disappear?

tests/data/publickey_openssl.pem
```diff
@@ -0,0 +1,9 @@
+-----BEGIN PUBLIC KEY-----
```
craigcitro, line 1: is this a public key you have the private key for?

oauth2client/client.py
craigcitro: [Sample code](https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/third_party/storage_apitools/credentials_lib.py#L135) where we use `urllib2` in `gsutil`, for exactly this purpose.

orestica: Done.

oauth2client/client.py
orestica: Indeed, I should have used the get('metadata-flavor', '') instead of checking like this.
In the new implementation (using urllib2 instead of httplib2), it is not needed anymore.

oauth2client/client.py
orestica: Combined the previous two lines into one.

oauth2client/client.py
orestica: I borrowed the idea from the docstring of class AccessTokenCredentials.
I moved it down to get_default() and from_stream() methods.

oauth2client/client.py
orestica: Done.

oauth2client/client.py
orestica: Done.

oauth2client/client.py
orestica: Done.

oauth2client/client.py
orestica: I used " because there is 'gcloud auth login' somewhere there inside the string.
I consolidated the text a little bit.

oauth2client/client.py
orestica: I changed it to credential_filename. Let's leave like this.

oauth2client/client.py
orestica: Renamed it to extra_help.

oauth2client/client.py
orestica, line 1050: We plan to announce the 'gcloud auth login' in the public go to page for education purposes.
I sent you an internal link (via email) regarding where I borrowed the code from.

oauth2client/client.py
orestica: Yes.
I moved everything on a single line.

oauth2client/client.py
orestica: Got rid of it.

oauth2client/client.py
orestica, line 1093: This was my rationale behind the previous ", as well.

oauth2client/client.py
orestica: Thanks.
Python style is more concise, indeed.

oauth2client/client.py
orestica: I restructured it a little bit.
I would rather keep the validation of fields' existence here, than pushing it down into the credentials classes.

oauth2client/client.py
orestica: This is more elegant, indeed. Thank you.

oauth2client/client.py
orestica, line 1133: Moving it as a public method to appengine.AppAssertionCredentials class would increase the new surface we introduce. We don't want that.
Moving it as a private method over there wouldn't gain us much: we'd still need an import followed by a one line to call this method.
I would rather let it here.

oauth2client/client.py
orestica, line 1136: It needs only scopes. Passing an empty list of scopes defers them for later filling.

oauth2client/gce.py
orestica: Done.

oauth2client/service_account.py
orestica: Dropped it.

oauth2client/service_account.py
orestica: Done.

oauth2client/service_account.py
orestica: Neat.
Thank you.

oauth2client/service_account.py
orestica: Done.

samples/app.yaml
orestica: Done.

samples/call_compute_service_from_gae.py
orestica: Done.

setup.py
orestica, line 30: Should I get rid of the versions?

tests/data/publickey_openssl.pem
orestica, line 1: This public key was generated from the certificate from the existing publickey.pem file, by running:

openssl x509 -pubkey -noout -in publickey.pem

This public key goes hand in hand with the private key from various privatekey files from the same directory (as you'll see in test_service_account.py).

tests/test_oauth2client.py
craigcitro, line 151: A few thoughts:
- The tests are never run in any sort of parallel/multithreaded environment, so there's unlikely to be a problem on that front.
- Most of the env stuff is harmless enough, but modifying anything in `os` is more worrisome.
- There's definitely a cleaner way to handle this in python, though I'm not sure how much you want to do. The "right" answer here is to write a [context manager](https://docs.python.org/2/library/contextlib.html); roughly speaking, you'd write something that looks like (where you're welcome to change the names):
  
  ```
  @contextlib.contextmanager
  def SetupEnv(desired_os, desired_server_software):
    original_server_software = os.environ.get('SERVER_SOFTWARE')
    os.environ['SERVER_SOFTWARE'] = desired_server_software
    ... <more env setup> ...
    yield
    os.environ['SERVER_SOFTWARE'] = original_server_software
    ... <more resetting> ...
  ```
  
  Then you can just say
  
  ```
  with SetupEnv('nt', 'DEV_APPSERVER'):
    ... <tests for that env> ...
  ```

tests/test_oauth2client.py
```diff
@@ -107,6 +120,314 @@ def test_to_from_json(self):
...
+    self.reset_env('GOOGLE_CREDENTIALS_DEFAULT',
+                   self.env_google_credentials_default)
+    self.reset_env('APPDATA', self.env_appdata)
+    os.name = self.os_name
+
+  def reset_env(self, env, value):
```
craigcitro, line 157: I think you could simplify this using `os.environ.pop`, but it's not deeply important.

tests/test_oauth2client.py
```diff
@@ -107,6 +120,314 @@ def test_to_from_json(self):
...
+    self.assertEqual(None, credentials.access_token)
+    self.assertEqual('123', credentials.client_id)
+    self.assertEqual('secret', credentials.client_secret)
+    self.assertEqual('alabalaportocala', credentials.refresh_token)
+    self.assertEqual(None, credentials.token_expiry)
+    self.assertEqual(GOOGLE_TOKEN_URI, credentials.token_uri)
+    self.assertEqual('Python client library', credentials.user_agent)
+
+  def get_a_google_credentials_object(self):
+    return GoogleCredentials(None,
```
craigcitro: Let's compact these

tests/test_oauth2client.py
```diff
@@ -107,6 +120,314 @@ def test_to_from_json(self):
...
+        self.get_a_google_credentials_object().create_scoped_required())
+
+  def test_create_scoped(self):
+    credentials = self.get_a_google_credentials_object() 
+    self.assertEqual(credentials, credentials.create_scoped(None))
+    self.assertEqual(credentials,
+                     credentials.create_scoped(['dummy_scope']))
+
+  def test_get_environment_gae_production(self):
+    os.environ['SERVER_SOFTWARE'] = 'Google App Engine/XYZ'
```
craigcitro, line 195: Probably want to reset this after.

craigcitro, line 195: (and below)

tests/test_oauth2client.py
```diff
@@ -107,6 +120,314 @@ def test_to_from_json(self):
...
+    environment_variable_file = datafile(
+        os.path.join('gcloud', 'credentials_default.json'))
+    os.environ['GOOGLE_CREDENTIALS_DEFAULT'] = environment_variable_file
+    self.assertEqual(environment_variable_file,
+                     _get_environment_variable_file())
+
+  def test_get_environment_variable_file_error(self):
+    nonexistent_file = datafile('nonexistent')
+    os.environ['GOOGLE_CREDENTIALS_DEFAULT'] = nonexistent_file
+    try:
```
craigcitro, line 247: You should check out `self.assertRaises`, which will let you clean this up.

tests/test_oauth2client.py
craigcitro: Yes! I also spotted this the other day and fixed it, but apparently hadn't pushed to this repo; it's now e48d951.

tests/test_service_account.py
```diff
@@ -0,0 +1,121 @@
...
+import rsa
+import time
+import unittest
+
+from http_mock import HttpMockSequence
+from oauth2client.anyjson import simplejson
+from oauth2client.service_account import _ServiceAccountCredentials
+
+
+def datafile(filename):
```
craigcitro, line 33: The right way to do this is with `pkgutil.get_data`; we do this in the [bq code](https://code.google.com/p/google-bigquery-tools/source/browse/bq/bigquery_client.py#471), but it also requires a little work in `setup.py` to make sure you've got it right. You can leave a `TODO` on this one if you'd like.

tests/test_oauth2client.py
orestica, line 157: Is this what you had in mind?

tests/test_oauth2client.py
orestica: Done.

tests/test_oauth2client.py
orestica, line 195: setUp() and tearDown() take care of this.

tests/test_oauth2client.py
orestica, line 247: self.assertRaises() looks nice if I just want to make sure a DefaultCredentialsError exception is being raised. But could I test its message the way I do it here?

tests/test_service_account.py
orestica, line 33: I got inspired from test_jwt.py, which uses this very pattern.
I put a TODO.

oauth2client/appengine.py
```diff
@@ -164,6 +164,7 @@ def __init__(self, scope, **kwargs):
           unspecified, the default service account for the app is used.
     """
     self.scope = util.scopes_to_string(scope)
+    self.kwargs = kwargs
```
craigcitro: Any reason to leave this public?

(I could _swear_ I already made this comment, but don't see it; if I'm repeating myself, just tell me I'm getting senile.)

oauth2client/client.py
```diff
@@ -820,7 +838,304 @@ def _revoke(self, http_request):
...
+    _env_name = 'GAE_PRODUCTION'
+  elif server_software.startswith('Development/'):
+    _env_name = 'GAE_LOCAL'
+  else:
+    import urllib2
+    try:
+      if urllib2_urlopen is None:
+        urllib2_urlopen = urllib2.urlopen
+      response = urllib2_urlopen('http://metadata.google.internal')
+      if 'Metadata-Flavor: Google\r\n' in response.info().headers:
```
craigcitro: It might be worth calling `.rstrip()` on the headers and checking without the `\r\n` -- that seems like it might fail painfully one day.

craigcitro: or, alternately:

```
if any('Metadata-Flavor: Google' in h for h in response.info().headers):
```

oauth2client/client.py
```diff
@@ -820,7 +838,304 @@ def _revoke(self, http_request):
...
+      else:
+        _env_name = 'UNKNOWN'
+    except urllib2.URLError:
+      _env_name = 'UNKNOWN'
+
+  return _env_name
+
+
+class GoogleCredentials(OAuth2Credentials):
+  """Provides Default Credentials to be used in authenticating
```
craigcitro: This still needs to be shortened to one line.

oauth2client/client.py
craigcitro, line 1050: OK; I'd ultimately like to figure out a way to have gcloud tell us this file location, but that doesn't have to happen right now.

oauth2client/client.py
```diff
@@ -820,7 +838,304 @@ def _revoke(self, http_request):
...
+
+  if (client_credentials.get('type') not in ('authorized_user',
+                                             'service_account')):
+    raise DefaultCredentialsError("'type' field should be defined "
+                                  "(and have one of the 'authorized_user' "
+                                  "or 'service_account' values)")
+
+  if client_credentials['type'] == 'authorized_user':
+    required_fields = set(['client_id', 'client_secret', 'refresh_token'])
+  else: # client_credentials['type'] == 'service_account'
```
craigcitro: nit: python-style wants two spaces before the comment (so `else:  # ...`).

craigcitro: also, in this case, why not just combine this with the previous if:

```
credentials_type = client_credentials.get('type')
if credentials_type == 'authorized_user':
  ...
elif credentials_type == 'service_account':
  ...
else:
  raise DefaultCredentialsError(...)
```

oauth2client/client.py
```diff
@@ -820,7 +838,304 @@ def _revoke(self, http_request):
...
+def _get_default_credential_from_file(default_credential_filename):
+  """Build the Default Credentials from file."""
+
+  import service_account
+
+  # read the credentials from the file
+  with open(default_credential_filename) as default_credential:
+    client_credentials = service_account.simplejson.load(default_credential)
+
+  if (client_credentials.get('type') not in ('authorized_user',
```
craigcitro: can you give these two strings (`'authorized_user'` and `'service_account'`) names that we use throughout?

setup.py
craigcitro, line 30: I'm less worried about the versions, and more worried about the prospect of those packages disappearing from the world.

tests/data/publickey_openssl.pem
craigcitro, line 1: aha -- I forgot there was already a key in this repo. no worries.

tests/test_oauth2client.py
craigcitro, line 157: Yep. Can you add a docstring for this function?

tests/test_oauth2client.py
craigcitro, line 247: Yes -- also check out `assertRaisesRegexp`.

tests/test_service_account.py
craigcitro, line 33: TODO works.

oauth2client/appengine.py
orestica: No reason. I just wanted to preserve the look of the surrounding instance variable.
I made it private.

oauth2client/client.py
orestica: Chose the 'if 'Metadata-Flavor: Google\r\n' in response.info().headers:' option.

oauth2client/client.py
orestica: I remember seeing this suggestion in the email, but it did not show up in https://github.com/google/oauth2client/pull/6/files
Have you experienced inconsistencies with github's reviewing tools?

Done.

oauth2client/client.py
orestica, line 1050: I left a TODO.

oauth2client/client.py
orestica: Done.

oauth2client/client.py
craigcitro: Yes, I think sometimes comments are hard to find unless you trawl through all the commits. It's sub-par.

oauth2client/client.py
orestica: Done.

setup.py
orestica, line 30: Replied to you on email.

tests/data/publickey_openssl.pem
orestica, line 1: Ack.

tests/test_oauth2client.py
orestica, line 157: Done.

tests/test_oauth2client.py
orestica, line 247: Done.
Thank you.

Issue #4, opened 2014-05-04
failing to get private key in crypt.py from .p12 file


iditgolden:
when running line 131 with a .p12 file:

```
pkey = crypto.load_pkcs12(key, password).get_privatekey()
```

the following error occurs: 
OpenSSL.crypto.Error: [('asn1 encoding routines', 'ASN1_D2I_READ_BIO', 'not enough data')]

should be:

```
pkey = crypto.load_pkcs12(file(key,"rb").read(),str(password)).get_privatekey()
```

PR #8, opened 2014-05-12
fix the private key file openning and getting the private key

iditgolden:
Related to issue: https://github.com/google/oauth2client/issues/4

Fixed the private key file loading.

### DIFF
--- a/oauth2client/crypt.py
+++ b/oauth2client/crypt.py
@@ -128,7 +128,7 @@ try:
       if parsed_pem_key:
         pkey = crypto.load_privatekey(crypto.FILETYPE_PEM, parsed_pem_key)
       else:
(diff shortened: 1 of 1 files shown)


oauth2client/crypt.py
```diff
@@ -128,7 +128,7 @@ def from_string(key, password='notasecret'):
       if parsed_pem_key:
         pkey = crypto.load_privatekey(crypto.FILETYPE_PEM, parsed_pem_key)
       else:
-        pkey = crypto.load_pkcs12(key, password).get_privatekey()
+        pkey = crypto.load_pkcs12(file(key,"rb").read(),str(password)).get_privatekey()
```
skelterjohn: I would have thought "open(key)" instead of file. Also, you'd need to do, eg
with open(key) as key_file:
  ... key_file.read() ...

PR #9, opened 2014-05-16
Coerce password to str so assertion credentials can be refreshed.

skelterjohn:
Fixes http://code.google.com/p/google-api-python-client/issues/detail?id=328

Still needs a unit test.


oauth2client/crypt.py
```diff
@@ -128,7 +128,7 @@ def from_string(key, password='notasecret'):
       if parsed_pem_key:
         pkey = crypto.load_privatekey(crypto.FILETYPE_PEM, parsed_pem_key)
       else:
-        pkey = crypto.load_pkcs12(key, password).get_privatekey()
+        pkey = crypto.load_pkcs12(key, str(password)).get_privatekey()
```
craigcitro, line 131: should we be worried about the case of actual non-ascii characters in password?

skelterjohn, line 131: I'd say that that decision has been made for us by the crypto module.

PR #18, opened 2014-06-19
Support the login_hint parameter

jay0lee:
Add support for setting the login_hint parameter which can simplify the authorization flow if the application knows the email address or domain the user should authenticate with. The parameter is defined at: https://developers.google.com/accounts/docs/OAuth2WebServer#formingtheurl

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -1504,11 +1504,11 @@ class OAuth2WebServerFlow(Flow):
   @util.positional(4)
   def __init__(self, client_id, client_secret, scope,
                redirect_uri=None,
-               login_hint=None,
                user_agent=None,
                auth_uri=GOOGLE_AUTH_URI,
                token_uri=GOOGLE_TOKEN_URI,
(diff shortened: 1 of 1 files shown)


oauth2client/client.py
```diff
@@ -1656,7 +1660,7 @@ def step2_exchange(self, code, http=None):
 
 @util.positional(2)
 def flow_from_clientsecrets(filename, scope, redirect_uri=None,
-                            message=None, cache=None):
+                            login_hint=None, message=None, cache=None):
```
skelterjohn: A new parameter needs to go on the end, in case people were not naming their args when using this function.

oauth2client/client.py
```diff
@@ -1504,6 +1504,7 @@ class OAuth2WebServerFlow(Flow):
   @util.positional(4)
   def __init__(self, client_id, client_secret, scope,
                redirect_uri=None,
+               login_hint=None,
```
skelterjohn: A new parameter needs to go on the end.

PR #26, opened 2014-07-15
GCE VM needs the Compute API enabled

orestica:
Added a comment emphasizing that the GCE VM needs to have the Compute API enabled.


samples/call_compute_service.py
```diff
@@ -1,5 +1,6 @@
 # To be used to test GoogleCredentials.get_application_default()
 # from local machine and GCE.
+# The GCE virtual machine needs to have the Compute API enabled.
```
craigcitro: I think we need to at _least_ say "service account" here; better yet would be a pointer to some docs.

orestica: Done.

PR #27, opened 2014-07-17
Introduce the save_to_well_known_file() method.

orestica:


oauth2client/client.py
```diff
@@ -985,6 +985,16 @@ def create_scoped(self, scopes):
     """
     return self
 
+  @property
+  def serialization_data(self):
+    """Get the fields and their values identifying the current credentials."""
+    credentials_data = {}
```
craigcitro: it's not a huge difference, but i'd rather see this as an inline literal, and probably even just return it -- something like

```
return {
    'type': 'authorized_user',
    'client_id': self.client_id,
    ...
}
```

oauth2client/client.py
```diff
@@ -1064,6 +1076,28 @@ def from_stream(credential_filename):
           'method should point to a file.')
 
 
+def save_to_well_known_file(credentials, test_well_known_file=None):
```
craigcitro: why not just call it `well_known_file`?

oauth2client/appengine.py
```diff
@@ -197,6 +197,10 @@ def _refresh(self, http_request):
       raise AccessTokenRefreshError(str(e))
     self.access_token = token
 
+  @property
+  def serialization_data(self):
+    raise NotImplementedError('No credentials data for App Engine.')
```
craigcitro: maybe `Cannot serialize credentials on App Engine`? actually, that's not true ... maybe `Cannot serialize credentials for AppEngine service accounts`?

(same for GCE below)

oauth2client/client.py
craigcitro: also, should we put this in `tools`?

oauth2client/client.py
orestica: Done.

oauth2client/client.py
orestica: I hope you were referring to the name of the parameter, not the name of the method.
I had wanted to emphasize that it should be used for testing only.
I changed it as you suggested.

orestica: Regarding, the move to tools, it can't be done because of the argparse which is imported in tools.py and doesn't play nicely with Python 2.6: https://travis-ci.org/orestica/oauth2client/builds/30218894

oauth2client/appengine.py
orestica: Done.

oauth2client/client.py
craigcitro: re: tools, that makes sense -- we need to fix the argparse thing (it can be optionally installed in 2.6), so maybe leave a TODO?

craigcitro: re: naming, yes, the parameter.

PR #37, opened 2014-08-18
More import/lint cleanup.

craigcitro:
This is a follow-on to the last CL (which dropped the `json` wrapper); it's mostly import cleanups, along with a few related things I saw while doing import cleanup.

PTAL @skelterjohn

### DIFF
--- a/oauth2client/tools.py
+++ b/oauth2client/tools.py
@@ -25,21 +25,14 @@ __all__ = ['argparser', 'run_flow', 'run', 'message_if_missing']
 
 import BaseHTTPServer
 import argparse
-import httplib2
 import logging
-import os
 import socket
 import sys
+import urlparse
 import webbrowser
 
 from oauth2client import client
-from oauth2client import file
 from oauth2client import util
-
-try:
-  from urlparse import parse_qsl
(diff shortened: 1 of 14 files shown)


oauth2client/tools.py
```diff
@@ -84,26 +77,25 @@ class ClientRedirectHandler(BaseHTTPServer.BaseHTTPRequestHandler):
...
-    s.wfile.write("<html><head><title>Authentication Status</title></head>")
-    s.wfile.write("<body><p>The authentication flow has completed.</p>")
-    s.wfile.write("</body></html>")
+    self.send_response(200)
+    self.send_header("Content-type", "text/html")
+    self.end_headers()
+    query = self.path.split('?', 1)[-1]
+    query = dict(urlparse.parse_qsl(query))
+    self.server.query_params = query
+    self.wfile.write("<html><head><title>Authentication Status</title></head>")
```
skelterjohn, line 93: """<html>..... """ ?

craigcitro, line 93: i'm going to do some multiline string cleanup in a few places as a follow-on.

Issue #46, opened 2014-08-29
Use redirect_uri from client_secrets.json


bitle:
Client_secrets.json contains redirect_uri and one doesn't need to be provided separately.

```
# client_secrets.json
{
    "web":
    {
        "client_id":"blah-blah",
        "client_secret":"blah-blah",
        "client_email":"blah-blah",
        "redirect_uris":["https://example.com/oauth2callback"],
    }
}
```

PR #48, opened 2014-08-31
Use redirect_uri from client_secrets.json #46

bitle:
Clientsecrets file downloaded from Google APIs project already contains redirect_uri and one doesn't need to provide it separately.

This PR keeps old functionality, when `redirect_uri` can be provided as an argument to `flow_from_clientsecrets`.

Given client_secrets.json that contains redirect uri:

```
# client_secrets.json
{
    "web":
    {
        "client_id":"blah-blah",
        "client_secret":"blah-blah",
        "client_email":"blah-blah",
        "redirect_uris":["https://example.com/oauth2callback"],
    }
}
```

The following two lines of code have the same behavior:

```
redirect_uri = 'https://example.com/oauth2callback
# old way:
flow = flow_from_clientsecrets('client_secrets.json', scopes, redirect_uri=redirect_uri)
# new way:
flow = flow_from_clientsecrets('client_secrets.json', scopes)
```

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -1864,7 +1864,6 @@ def flow_from_clientsecrets(filename, scope, redirect_uri=None,
     client_type, client_info = clientsecrets.loadfile(filename, cache=cache)
     if client_type in (clientsecrets.TYPE_WEB, clientsecrets.TYPE_INSTALLED):
       constructor_kwargs = {
-          'redirect_uri': redirect_uri,
           'auth_uri': client_info['auth_uri'],
           'token_uri': client_info['token_uri'],
           'login_hint': login_hint,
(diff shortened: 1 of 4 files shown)


oauth2client/client.py
```diff
@@ -1874,9 +1873,11 @@ def flow_from_clientsecrets(filename, scope, redirect_uri=None,
         constructor_kwargs['revoke_uri'] = revoke_uri
       if device_uri is not None:
         constructor_kwargs['device_uri'] = device_uri
+      if not redirect_uri and len(client_info['redirect_uris']) > 0:
+          redirect_uri = client_info['redirect_uris'][0]
```
craigcitro, line 1877: it seems odd to only use the first `redirect_uri` here -- and in particular might lead to some confusing failure modes ... do we even know anything about the order they're returned?

craigcitro, line 1877: also -- rather than passing it as an extra arg, might as well just do

```
constructor_kwargs['redirect_uri'] = ...
```

bitle, line 1877: Very valid comment!
Looks like it's possible to define multiple redirect URIs in APIs project console. This may explain why one has to provide redirect_uri separately.
URIs are returned in the same order they are defined in the API project console.

I see a few possible ways to move forward:
1. Close this work, since it may cause confusion
2. Raise an exception if there is more than one redirect URIs and one needs to be provided explicitly
3. Log a warning message that the first uri was used

craigcitro, line 1877: i'm torn -- i really like the idea of not needing to type the `redirect_uri` in the case that there's exactly one, but i wonder if it just leads to too many error cases if things go wrong.

if you're still energetic, i'd say it's worth at least playing with (2) and see how easy it is to add helpful/sensible errors.

PR #51, opened 2014-09-09
Support X509 pem cert and signature verification using PyCrypto.

liujin-google:


oauth2client/crypt.py
```diff
@@ -167,7 +168,10 @@ def verify(self, message, signature):
         return PKCS1_v1_5.new(self._pubkey).verify(
             SHA256.new(message), signature)
       except:
-        return False
+        lsignature = long(signature.encode('hex'), 16)
```
craigcitro: let's move this out of the `except` clause.

tests/test_jwt.py
```diff
@@ -189,7 +189,7 @@ class PEMCryptTestsPyCrypto(CryptTests):
   def setUp(self):
     self.format = 'pem'
     self.signer = crypt.PyCryptoSigner
-    self.verifier = crypt.OpenSSLVerifier
+    self.verifier = crypt.PyCryptoVerifier
```
skelterjohn, line 192: Does this mean that the previous cert types will not be tested? I'd think making this addition would be one or more _new_ tests, rather than changing an existing one.

oauth2client/crypt.py
liujin-google: Reverted the changes here since the current PKCS1_v1_5 turns out working well.

tests/test_jwt.py
liujin-google, line 192: The previous openssl verifier is tested in PEMCryptTestsOpenSSL tests. The existing assignment of crypt.OpenSSLVerifier to PEMCryptTestsPyCrypto.verifier is just a workaround to avoid the "X509 not supported" exception in PyCryptoVerifier. With this patch, that workaround is no longer needed.

PR #52, opened 2014-09-11
[python3]fix bytes-str conversion

nazo:


oauth2client/client.py
```diff
@@ -1429,7 +1429,7 @@ def _urlsafe_b64decode(b64string):
   # Guard against unicode strings, which base64 can't handle.
   if isinstance(b64string, six.text_type):
     b64string = b64string.encode('ascii')
-  padded = b64string + '=' * (4 - len(b64string) % 4)
+  padded = b64string + '='.encode('ascii') * (4 - len(b64string) % 4)
```
methane: `b'='`

PR #54, opened 2014-09-21
Don't send scopes on token request

jay0lee:
The docs at:
https://developers.google.com/accounts/docs/OAuth2InstalledApp#handlingtheresponse

don't mention that scopes should be sent for the token request. Doing so seems to work fine when only requesting a few scopes but when many scopes are being requested, a invalid_scope error is returned by the Google servers.


nathanielmanistaatgoogle:
Have you yet signed Google's Contributor License Agreement? We can't accept your contribution without your having done so.

dhermes:
@jay0lee Are you planning on updating and re-sending?

Issue #64, opened 2014-11-03
Keep the raw value of the OAuth2 id_token


avivien:
I've created a path to keep the raw value of the id_token in the token_reponse attribute of the OAuth2Credentials object.
This value is needed in some case, for example for the Amazon AWS Web identity federation.

PR #65, opened 2014-11-03
Patch to keep the raw value of the OAuth2 id_token

avivien:
Path to keep the raw value of the id_token in the token_reponse attribute of the OAuth2Credentials object.
This value is needed in some case, for example for the Amazon AWS Web identity federation.
See issue #64

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -1838,14 +1838,14 @@ class OAuth2WebServerFlow(Flow):
 
       extracted_id_token = None
       if 'id_token' in d:
-        extracted_id_token = _extract_id_token(d['id_token'])
+        d['extracted_id_token'] = _extract_id_token(d['id_token'])
 
       logger.info('Successfully retrieved access token')
       return OAuth2Credentials(access_token, self.client_id,
                                self.client_secret, refresh_token, token_expiry,
                                self.token_uri, self.user_agent,
                                revoke_uri=self.revoke_uri,
-                               id_token=extracted_id_token,
+                               id_token=d.get('extracted_id_token', None),
                                token_response=d)
     else:
       logger.info('Failed to retrieve access token: %s', content)
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -1836,15 +1836,16 @@ class OAuth2WebServerFlow(Flow):
         token_expiry = datetime.datetime.utcnow() + datetime.timedelta(
             seconds=int(d['expires_in']))
 
+      extracted_id_token = None
       if 'id_token' in d:
-        d['id_token'] = _extract_id_token(d['id_token'])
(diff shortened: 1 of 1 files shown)


oauth2client/client.py
```diff
@@ -1836,15 +1836,16 @@ def step2_exchange(self, code=None, http=None, device_flow_info=None):
         token_expiry = datetime.datetime.utcnow() + datetime.timedelta(
             seconds=int(d['expires_in']))
 
+      extracted_id_token = None
       if 'id_token' in d:
-        d['id_token'] = _extract_id_token(d['id_token'])
+        extracted_id_token = _extract_id_token(d['id_token'])
```
craigcitro: this seems wise, but now i wonder: is anyone already depending on getting the extracted token back in the `token_response` field below?

maybe add it into the response as `d['extracted_id_token']`?

avivien: Yes it's a good idea, I can do that, it will avoid any problems if someone uses the d['id_token'] as the extracted_id_token.
Do you want I submit an other Pull request?

craigcitro: no need for a new one -- just push to the same branch and it'll show up as part of this PR.

skelterjohn: Regarding on repending on the old token_response value, if someone did that they were already in for a bad time. BOTH forms will appear in the serialized credential store, depending on whether or not a refresh has ever happened.

skelterjohn: I only know this because a few days ago I was trying to figure out why something unrelated wasn't refreshing properly, and seeing the mismatched token_response in the before/after for the credential store was a bit of a red herring.

craigcitro: ah, so are you saying we _shouldn't_ keep the un-extracted version at all?

skelterjohn: I'm saying that a field named "token_response" should always have the same structure. having some extra info there doesn't bother me, but it seems like limited utility and we might be able to safely omit it.

craigcitro: i think i'm more just being wary about throwing it away on the off chance that people are using it -- though, as you point out, it could be that people are shooting themselves in the foot with it. ;)

let's go ahead and drop it for now, and i'll add a note in the `CHANGELOG` (so that someone who was looking for the old value can pipe up if it breaks them).

avivien: I agree with you John, and in my initial fix I didn't modify the "token_response" structure, I've just let the token id in its original format, because in the current code, the token id of the "token_response" is overwritten.
And like I said in the Pull request comment, there is an utility, the OAuth un-extracted token id is needed for the Amazon AWS Web identity federation.
http://docs.aws.amazon.com/STS/latest/UsingSTS/web-identity-federation.html
Maybe there is a better way to do that?

PR #71, opened 2014-11-13
Handle a missing argparse more gracefully.

craigcitro:
Currently, if argparse isn't installed, we'll fail when trying to import
`oauth2client.tools`. Given that this module can still be useful (say for
someone using `gflags`), it seems friendlier to simply skip the one top-level
operation requiring `argparse`.

PTAL @skelterjohn

### DIFF
--- a/oauth2client/tools.py
+++ b/oauth2client/tools.py
@@ -24,7 +24,6 @@ from __future__ import print_function
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 __all__ = ['argparser', 'run_flow', 'run', 'message_if_missing']
 
-import argparse
 import BaseHTTPServer
 import logging
 import socket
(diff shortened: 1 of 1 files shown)


oauth2client/tools.py
```diff
@@ -51,17 +50,25 @@
...
-                       help='Hostname when running a local web server.')
-argparser.add_argument('--noauth_local_webserver', action='store_true',
-                       default=False, help='Do not run a local web server.')
-argparser.add_argument('--auth_host_port', default=[8080, 8090], type=int,
-                       nargs='*', help='Port web server should listen on.')
-argparser.add_argument('--logging_level', default='ERROR',
-                       choices=['DEBUG', 'INFO', 'WARNING', 'ERROR',
-                                'CRITICAL'],
-                       help='Set the logging level of detail.')
+argparser = _CreateArgumentParser()
```
skelterjohn, line 53: oops :) need to assign this after the function is defined

craigcitro, line 53: ... something tells me we also need more tests.

PR #83, opened 2014-12-05
Refactoring _get_environment.

dhermes:
- Removes use of "global" by putting the env name in another
  namespace.
- Splits logic into digestible pieces.
- Also makes slight change to GoogleCredentials docstring.

This is the first chunk of #80. I'm happy to wait until #82 is in and rebase.

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -33,7 +33,6 @@ from six.moves import urllib
 
 import httplib2
 from oauth2client import clientsecrets
-from oauth2client import client_settings
 from oauth2client import GOOGLE_AUTH_URI
 from oauth2client import GOOGLE_DEVICE_URI
 from oauth2client import GOOGLE_REVOKE_URI
@@ -90,6 +89,10 @@ ADC_HELP_MSG = (
 # The access token along with the seconds in which it expires.
 AccessTokenInfo = collections.namedtuple(
     'AccessTokenInfo', ['access_token', 'expires_in'])
+
+DEFAULT_ENV_NAME = 'UNKNOWN'
+SETTINGS = collections.namedtuple(
+  'Settings', ['env_name'])(env_name=None)
 
 
 class Error(Exception):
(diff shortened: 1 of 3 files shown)


oauth2client/client.py
```diff
@@ -952,37 +972,19 @@ class GoogleCredentials(OAuth2Credentials):
   Here is an example of how to use the Application Default Credentials for a
   service that requires authentication:
 
-  <code>
```
craigcitro, line 955: oh -- were these `<code>` tags in there for epydoc? (_sigh_ epydoc)

oauth2client/client.py
```diff
@@ -952,37 +972,19 @@ class GoogleCredentials(OAuth2Credentials):
...
+      service = build('compute', 'v1', credentials=credentials)
 
-  print(response)
-  </code>
+      PROJECT = 'bamboo-machine-422'
+      ZONE = 'us-central1-a'
+      request = service.instances().list(project=PROJECT, zone=ZONE)
+      response = request.execute()
 
-  A service that does not require authentication does not need credentials
```
craigcitro, line 972: any reason not to keep this one?

dhermes, line 972: Yes. It is in a credentials class, so discussing a service that doesn't require credentials is off-topic.

oauth2client/client.py
dhermes, line 955: I don't know anything about epydoc, so it's unclear. I hope not.

craigcitro, line 955: can you try a quick `make docs` and see how they look?

oauth2client/client.py
craigcitro, line 972: s'fair.

oauth2client/client.py
dhermes, line 955: http://google.github.io/oauth2client/doc/oauth2client.client.GoogleCredentials-class.html
is the place it gets spit out and it puts the code block in a `<pre>` tag, which is a literal environment, so the tag does nothing (as you can see).

On re-build, the content is still in a `<pre>` environment.

PR #84, opened 2014-12-05
Refactor GoogleCredentials.get_application_default

dhermes:
This should have a diffbase of #83 but we can't specify those on GitHub.

As in #83 (after #82), I prefer to rebase this after #83 is merged.

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -1037,18 +1037,24 @@ class GoogleCredentials(OAuth2Credentials):
     }
 
   @staticmethod
-  def get_application_default():
-    """Get the Application Default Credentials for the current environment.
+  def _get_implicit_credentials():
+    """Gets credentials implicitly from the environment.
+
+    Checks environment in order of precedence:
+    - Google App Engine (production and testing)
+    - Environment variable GOOGLE_APPLICATION_CREDENTIALS pointing to
+      a file with stored credentials information.
+    - Stored "well known" file associated with `gcloud` command line tool.
+    - Google Compute Engine production environment.
 
     Exceptions:
       ApplicationDefaultCredentialsError: raised when the credentials fail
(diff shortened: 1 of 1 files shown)


oauth2client/client.py
```diff
@@ -1037,49 +1037,123 @@ def serialization_data(self):
...
+
+    if not credentials_filename:
+      return
+
+    try:
+      return _get_application_default_credential_from_file(credentials_filename)
+    except (ApplicationDefaultCredentialsError, ValueError) as error:
+      _raise_exception_for_reading_json(credentials_filename, extra_help, error)
+
+  def _get_implicit_credentials():
```
craigcitro, line 1116: why does this need to be a separate function?

dhermes, line 1116: Good point. More future creep.

However, I think it's fine as-is since the plan is to add the 3LO extensibility in to `_get_implicit_credentials` as an optional code path.

craigcitro, line 1116: i'm happy with cleaning up/refactoring the logic here.

i guess i'm confused about where we want this to go -- i thought we wanted to add a new function that amounted to

```
creds = GoogleCredentials.get_application_default()
if creds is None:
  # try 3LO
raise NoCreds("can't get any credentials")
```

i don't know why we need to refactor this to do that?

i can see one good argument, namely that my first call above will `raise` right now. but that doesn't seem to be what we're fixing here.

dhermes, line 1116: The idea is that `_implicit_credentials_from_files` would actually handle the 3LO (in `checkers` below).

This is done in #80, but it is super hacky because the files are in a different format than the JSON files used by `flow_from_clientsecrets`. This is due to inheriting the `gcloud` format. I also had to hack around `run_flow` requiring storage, when none was needed.

PR #95, opened 2014-12-18
Clean up test envs.

craigcitro:
We have a number of extra envs due to checking two versions of pyopenssl (0.13
and 0.14); this isn't actually necessary, since 0.14 is the one we really care
about for the external repo.

This cleans up the cases and names (and drastically shortens `tox.ini`), but
keeps the old envs available for manual testing.

PTAL @dhermes (why does github fail to autocomplete your name?)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,18 +1,15 @@
 [tox]
-envlist = py26openssl13, py26openssl14,
-          py27openssl13, py27openssl14,
-                         py33openssl14,
-                         py34openssl14,
-          pypyopenssl13, pypyopenssl14,
-          cover
+envlist = py26,py27,py33,py34,pypy,cover
 
 [testenv]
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -1,18 +1,15 @@
...
-       nose
+basedeps = keyring
+           mox3
+           pycrypto==2.6
+           django>=1.5,<1.6
+           webtest
+           nose
+deps = {[testenv]basedeps}
+       pyopenssl==0.14
 setenv = PYTHONPATH=../google_appengine
```
dhermes, line 13: Note this still isn't being tested on Travis. It'd be nice to download the SDK and unzip the directory somewhere.

craigcitro, line 13: totally agree -- filing an issue.

PR #98, opened 2014-12-18
Adding automated update to docs on merges to master.

dhermes:
See https://travis-ci.org/dhermes/oauth2client/builds/44409213 for this in action.

@craigcitro You'll need to set some environment variables for the project

```
travis login --github-token=$(cat travis.token)
travis env set GH_OAUTH_TOKEN $(cat travis.token) --repo "google/oauth2client"
travis env set GH_OWNER google --repo "google/oauth2client"
travis env set GH_PROJECT_NAME oauth2client --repo "dhermes/oauth2client"
```

(More detailed instructions for Travis stuff on https://github.com/GoogleCloudPlatform/gcloud-python-wheels/blob/master/README.md if you're not familiar)


scripts/update_docs.sh
```diff
@@ -0,0 +1,59 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ev
+
+###########################################
+# Install dependencies for building docs. #
+###########################################
+sudo apt-get install -y graphviz
```
craigcitro, line 22: oh noes -- this won't work with travis+docker, right?

**sigh** epydoc.

dhermes, line 22: Ha that is correct. Quickly googling around I'm not sure what to do, it seems to have a ton of deps to install from source.

Installing a deb without root:
http://askubuntu.com/questions/339/how-can-i-install-a-package-without-root-access

Debs for download:
http://www.graphviz.org/Download_linux_ubuntu.php

FWIW the images it creates are not particularly useful.

craigcitro, line 22: making a quick pass at a conversion to sphinx. `;)`

dhermes, line 22: I like it.

Issue #93, opened 2014-12-16
Detecting GCE sometimes takes forever.


craigcitro:
on some networks (@dhermes has one at home), doing a lookup on a bogus URL takes ~20-30s. this means that the default way of discovering the current environment (say for fetching the application default credential) can be insanely slow.

craigcitro:
@aaronwinter in this case, we know what IP the hostname resolves to, so i think we *don't* want to go the twisted/dnspython route. 

i think the best solution on the table is a three-pronged attack:
 * refactor the code a bit so that we can avoid the GCE check in more cases,
 * possibly switch to using the IP with a **short** timeout, and 
 * add an env var that says "i'm definitely not on gce, don't bother checking".

craigcitro:
The first step (switch to the IP for the metadata service) is bf9d46e.

aaronwinter:
Ok, thanks for the directions! Do you want the env var to be set to ~"no need to check" once a negative check has been done or something that is by default to "not check gce" and that the user will need to modify to enable gce checking?

dhermes:
@craigcitro Care to close this out?

craigcitro:
@dhermes i figured we'd leave it until all the steps above were done, but maybe we should track those separately ...

dhermes:
Which steps remain? You referring to #111?

I like to use check-marks to track a plan of work in progress (see GoogleCloudPlatform/gcloud-python#451 for example).

aaronwinter:
@dhermes I think @craigcitro is referring to this one:

> refactor the code a bit so that we can avoid the GCE check in more cases,

For future reference:

- [  ] refactor the code a bit so that we can avoid the GCE check in more cases,
- [x] possibly switch to using the IP with a short timeout 
- [x] add an env var that says "i'm definitely not on gce, don't bother checking".

theacodes:
Thank you for creating this issue, however, this project is deprecatedand we will only be addressing critical security issues. You can read moreabout this deprecation [here](http://google-auth.readthedocs.io/en/latest/oauth2client-deprecation.html).

If you need support or help using this library, we recommend that you ask yourquestion on StackOverflow.

If you still think this issue is relevant and should be addressed, pleasecomment and let us know!

PR #101, opened 2014-12-19
Add a timeout for GCE detection.

craigcitro:
This is a temporary workaround for #93, though may end up being the final fix.

I did some test cleanup while I was here, switching to `mock`.

PTAL @dhermes

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -26,6 +26,7 @@ import datetime
 import json
 import logging
 import os
+import socket
 import sys
 import time
 import six
@@ -930,12 +931,22 @@ def _detect_gce_environment(urlopen=None):
           Compute Engine.
   """
   urlopen = urlopen or urllib.request.urlopen
-
+  # Note: the explicit `timeout` below is a workaround. The underlying
+  # issue is that resolving an unknown host on some networks will take
+  # 20-30 seconds; making this timeout short fixes the issue, but
+  # could lead to false negatives in the event that we are on GCE, but
+  # the metadata resolution was particularly slow. The latter case is
+  # "unlikely".
   try:
-    response = urlopen('http://metadata.google.internal')
+    response = urlopen('http://metadata.google.internal/', timeout=1)
     return any('Metadata-Flavor: Google' in header
                for header in response.info().headers)
-  except urllib.error.URLError:
+  except socket.timeout:
+    logger.info('Timeout attempting to reach GCE metadata service.')
+    return False
+  except urllib.error.URLError as e:
+    if isinstance(getattr(e, 'reason', None), socket.timeout):
+      logger.info('Timeout attempting to reach GCE metadata service.')
     return False
 
 
--- a/tests/test_oauth2client.py
+++ b/tests/test_oauth2client.py
@@ -25,13 +25,11 @@ __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 import base64
 import datetime
 import json
-try:
-  from mox3 import mox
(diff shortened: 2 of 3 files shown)


oauth2client/client.py
```diff
@@ -930,12 +931,22 @@ def _detect_gce_environment(urlopen=None):
           Compute Engine.
   """
   urlopen = urlopen or urllib.request.urlopen
-
+  # Note: the explicit `timeout` below is a potentially-short-term
```
dhermes: Pretty silly to call it short-term and not give a date. I'd suggest either not mentioning that it's short term or giving it a timestamp (I know a `git blame` or other could locate the time, but people won't do that).

craigcitro: s'fair. dropped "potentially-short-term".

oauth2client/client.py
```diff
@@ -930,12 +931,22 @@ def _detect_gce_environment(urlopen=None):
...
+  # workaround. The underlying issue is that resolving an unknown host
+  # on some networks will take 20-30 seconds; making this timeout
+  # short fixes the issue, but could lead to false negatives in the
+  # event that we are on GCE, but the metadata resolution was
+  # particularly slow. The latter case is "unlikely".
   try:
-    response = urlopen('http://metadata.google.internal')
+    response = urlopen('http://metadata.google.internal/', timeout=1)
```
dhermes, line 941: It just occurred to me that only headers are necessary. Worth filing a bug to make this a `HEAD` request?

tox.ini
```diff
@@ -4,6 +4,7 @@ envlist = py26,py27,py33,py34,pypy,cover
 [testenv]
 basedeps = keyring
            mox3
+           mock
```
dhermes, line 7: Mox and mock? At the very least you should file an admittedly aspirational bug to choose one ~~ring~~ mocking library to rule them all.

tests/test_oauth2client.py
```diff
@@ -208,37 +206,19 @@ def test_get_environment_gae_local(self):
...
-    urllib2_urlopen.__call__(('http://metadata.google.internal'
-                             )).AndReturn(mockResponse)
-
-    m.ReplayAll()
-
-    self.assertEqual('GCE_PRODUCTION', _get_environment(urllib2_urlopen))
-
-    m.UnsetStubs()
-    m.VerifyAll()
+    with mock.patch.object(urllib.request, 'urlopen') as urlopen:
```
dhermes, line 209: Sorry for being a weenie but why are these necessary? What's wrong with `mox`?

oauth2client/client.py
craigcitro, line 941: we could -- but the response would only be large in the case of something like a captive portal, where i'm downloading a whole page, right? in that case, might be more work than it's worth.

dhermes, line 941: Yes after writing I thought "probably more work than it's worth" :)

tox.ini
craigcitro, line 7: #102

tests/test_oauth2client.py
craigcitro, line 209: my brain can also only keep one mocking library loaded, and `mock` seems to be the future (it's in the stdlib as of 3.3). might as well start moving that way.

dhermes, line 209: It's your mess, but that seems like a terrible idea without a clear plan.

craigcitro, line 209: going to push a change to drop mox shortly. otherwise anything else on this one?

PR #107, opened 2015-01-07
Explicitly fail for python3 versions < 3.3.

craigcitro:
we just need to add a check in `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,8 +21,11 @@ from __future__ import print_function
 
 import sys
 
-if sys.version_info <= (2, 6):
-  print('oauth2client requires python version >= 2.6.', file=sys.stderr)
+if sys.version_info < (2, 6):
+  print('oauth2client requires python2 version >= 2.6.', file=sys.stderr)
+  sys.exit(1)
+if (3, 1) <= sys.version_info < (3, 3):
+  print('oauth2client requires python3 version >= 3.3.', file=sys.stderr)
   sys.exit(1)
 
 from setuptools import setup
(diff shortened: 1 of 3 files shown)


craigcitro:
PTAL @dhermes if you have a sec?

dhermes:
@nicholaschen @craigcitro 

We have an [explicit section][1] explaining why we only support 3.3+.

In reality, enterprise didn't ever (still hasn't) really touch Py3. By forcing 3.3+, it's actually encouraging the community to move forward. See our reasoning for more. (And Craig it may be worth adding a similar section here.)

[1]: https://github.com/GoogleCloudPlatform/gcloud-python/blob/87273a15d9be2b56c7d51a3bea0b5884c2313af0/README.rst#supported-python-versions

setup.py
```diff
@@ -22,7 +22,10 @@
 import sys
 
 if sys.version_info <= (2, 6):
```
dhermes: So you've never support Python 2.6? I am surprised by this but SGTM.

dhermes:
LGTM

setup.py
craigcitro: oh, this actually does the right thing -- the version tuple for `2.6` will always have a third component.

that said, i agree it reads awkwardly, so i'm tweaking.

craigcitro:
OK, added some notes, submitting.

thanks again @nicholaschen for asking -- i wouldn't have thought to add a note. and thanks @dhermes for giving me boilerplate to steal. `;)`

PR #111, opened 2015-01-09
Avoid GCE checking if NO_GCE_CHECK is True

erwanor:
Following issue #93 there is now a variable NO_GCE _CHECK set by default to False that will skip any GCE check if set to True.


craigcitro:
@anthmgoogle can you take a look?

anthmgoogle:
Looks good to me.

aaronwinter:
Great!
@craigcitro I have signed the contributor license agreement by the way.

craigcitro:
sweet! @aaronwinter you answered my question before i even asked it. `;)`

i'm thinking we should maybe switch from a global to an env var? (or have the global default to the env var?)

aaronwinter:
@craigcitro what do you think of ac6142310a9edc47b1c7c6b591b18a7276bcad62 ?

aaronwinter:
@craigcitro I have pulled the recent updates made to insure backward compatibility. I believe that it is ready to be merged if you are OK with that?

oauth2client/client.py
```diff
@@ -970,7 +974,7 @@ def _get_environment(urlopen=None):
     SETTINGS.env_name = 'GAE_PRODUCTION'
   elif server_software.startswith('Development/'):
     SETTINGS.env_name = 'GAE_LOCAL'
-  elif _detect_gce_environment(urlopen=urlopen):
+  elif NO_GCE_CHECK is not 'True' and _detect_gce_environment(urlopen=urlopen):
```
craigcitro: for strings, we want to use `==` instead of `is`.

craigcitro:
one small nit, otherwise looks good.

aaronwinter:
@craigcitro Done! :+1:

craigcitro:
awesome -- thanks @aaronwinter!

PR #115, opened 2015-01-14
Adding protected method to convert PKCS12 key to PEM.

dhermes:
@craigcitro LMK if you think this should be a standalone function (possibly in `crypt.py`) instead of a protected method on `SignedJwtAssertionCredentials`.

### DIFF
--- a/tests/test_crypt.py
+++ b/tests/test_crypt.py
@@ -1,0 +1,73 @@
+# Copyright 2014 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import mock
+import os
+import sys
+import unittest
+
+try:
+  reload
+except NameError:
+  # For Python3 (though importlib should be used, silly 3.3).
+  from imp import reload
+
+from oauth2client.client import HAS_OPENSSL
+from oauth2client.client import SignedJwtAssertionCredentials
+from oauth2client import crypt
+
+
+def datafile(filename):
+  f = open(os.path.join(os.path.dirname(__file__), 'data', filename), 'rb')
+  data = f.read()
+  f.close()
+  return data
+
+
+class Test_pkcs12_key_as_pem(unittest.TestCase):
+
+  def _make_signed_jwt_creds(self, private_key_file='privatekey.p12',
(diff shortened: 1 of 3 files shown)


craigcitro:
+1, i'd rather see it as a separate function (which i suspect could be used elsewhere).

dhermes:
@craigcitro I factored out and made it public in `oauth2client.crypt`. I'm not sure how I feel about

- Putting the tests in `test_jwt` (but there is no `test_crypt`).
- Raising an `ImportError` at runtime. I can "resolve" this by just having the function convert PKCS12 instead of trying to also handle existing PEM keys.

craigcitro:
* for the tests, why not just make a `test_crypt.py`?
* i agree about the `ImportError`, but i don't see an obvious clean answer. i'm fine with (1) doing either an `ImportError` or `NotImplementedError` and (2) filing an issue to deal with the openssl dependency. (ultimately it'd be nice to just drop it.)

dhermes:
OpenSSL is a conditional dependency, that's the whole point of crypt.py (graceful degradation).

If I tear our the PEM part, then I can define this conditionally as OpenSSLVerifier and OpenSSLSigner are done.

(I have talked myself out of the ImportError.)

craigcitro:
wfm

dhermes:
@craigcitro PTAL

dhermes:
@craigcitro PTAL

tests/test_crypt.py
```diff
@@ -0,0 +1,73 @@
...
+  # For Python3 (though importlib should be used, silly 3.3).
+  from imp import reload
+
+from oauth2client.client import HAS_OPENSSL
+from oauth2client.client import SignedJwtAssertionCredentials
+from oauth2client import crypt
+
+
+def datafile(filename):
+  f = open(os.path.join(os.path.dirname(__file__), 'data', filename), 'rb')
```
craigcitro, line 32: why not just `pkgutil.get_data`?

dhermes, line 32: This is copy pasta from `test_jwt`. Also `pkgutil.get_data` would not accomplish anything close to what this method does.

This runs from `tests/` and opens files in `tests/data/` via the relative path given by `filename`.

craigcitro, line 32: pretty sure this whole function is just `pkgutil.get_data(__name__, os.path.join('data', filename))`. go ahead and give it a try before you decide it's unrelated. `;)`

not deeply important though, this code all needs some spring cleaning.

dhermes, line 32: Interesting. My bad.

dhermes:
Thanks Craig!

In the interest of closing GoogleCloudPlatform/gcloud-python#537, when do you think a new release will be cut?

craigcitro:
i should be able to cut a release tomorrow.

craigcitro:
actually, let's make that friday -- so that i can be teaching two other folks to do it at the same time.

craigcitro:
new release is out!

dhermes:
Thanks! I noticed: https://github.com/GoogleCloudPlatform/gcloud-python/pull/559#issuecomment-70327114

Issue #114, opened 2015-01-13
Proxies should never be used when querying GCE metadata server


thobrla:
Similar changes to https://codereview.appspot.com/187380044/ should be made.

dhermes:
I filed GoogleCloudPlatform/gcloud-python#757 to make sure we are also doing the right thing.

AFAICT we get around autodetected proxies (and `urllib` magic overhead) by directly using the `httplib` library:

```python
from six.moves.http_client import HTTPConnection
import socket

host = '169.254.169.254'
uri_path = '/computeMetadata/v1/project/project-id'
headers = {'Metadata-Flavor': 'Google'}
connection = HTTPConnection(host, timeout=0.1)

try:
    connection.request('GET', uri_path, headers=headers)
    response = connection.getresponse()
    if response.status == 200:
        return response.read()
except socket.error:  # socket.timeout or socket.error(64, 'Host is down')
    pass
finally:
    connection.close()
```

[1]: https://github.com/GoogleCloudPlatform/gcloud-python/blob/d34a15ed8cefe722233cd183866bcd9eb325439e/gcloud/_helpers.py#L146-L159

PR #120, opened 2015-01-22
Force GCE metadata querying to be done without proxies

erwanor:
Following issue #114


aaronwinter:
If my understanding is completing the test is failing because urlopen should be called at least once (see [test_oauth2client.py](https://github.com/google/oauth2client/blob/master/tests/test_oauth2client.py#L217). That being said according to [urllib2.py](http://www.opensource.apple.com/source/python/python-3/python/Lib/urllib2.py) urlopen ends up making a call to open anyway. Is that ok @craigcitro ?

craigcitro:
@aaronwinter i suspect it's just that the call doesn't match the expected one in the mock; can you look at updating it? (it's probably just that it needs the args to match.)

oauth2client/client.py
```diff
@@ -938,7 +938,8 @@ def _detect_gce_environment(urlopen=None):
   # the metadata resolution was particularly slow. The latter case is
   # "unlikely".
   try:
-    response = urlopen('http://169.254.169.254/', timeout=1)
+    no_proxy_conf = urllib.request.build_opener(urllib.request.ProxyHandler({}))
```
craigcitro: we can also pull this line out of the `try` block.

aaronwinter:
@craigcitro I have updated the arguments passed to open but the test is still failing. I am not very well versed with unittest. Can you take a quick look at it?

tests/test_oauth2client.py
```diff
@@ -209,21 +209,23 @@ def test_get_environment_gae_local(self):
 
   def test_get_environment_gce_production(self):
     os.environ['SERVER_SOFTWARE'] = ''
+    no_proxy_conf = urllib.request.build_opener(urllib.request.ProxyHandler({}))
     response = MockResponse({'Metadata-Flavor': 'Google'})
-    with mock.patch.object(urllib.request, 'urlopen',
+    with mock.patch.object(no_proxy_conf, 'open',
                            return_value=response,
-                           autospec=True) as urlopen:
+                           autospec=True) as open:
```
craigcitro: `open` is a dangerous name here (since it overrides the builtin); maybe `no_proxy_open`?

tests/test_oauth2client.py
```diff
@@ -209,21 +209,23 @@ def test_get_environment_gae_local(self):
...
       self.assertEqual('GCE_PRODUCTION', _get_environment())
-      urlopen.assert_called_once_with(
+      no_proxy_conf.open.assert_called_once_with(
```
craigcitro: i suspect you just want to use `open` here, since that's your mock. (or, you know, whatever you rename it to based on the above. `;)`)

nathanielmanistaatgoogle:
Please squash commits and push a new draft that passes tests?

dhermes:
@nathanielmanistaatgoogle I'm closing this since #293 handles it. @aaronwinter Feel free to re-open if you feel like my change misses something.

ASIDE: It'd be nice to get it in before cutting `1.5.0`.

PR #122, opened 2015-01-30
Added b prefix for string used in BaseHTTPRequestHandler wfile.write

kcs:
In Python 3 the BaseHTTPRequestHandler need bytes streams to write to the output, with simple strings the following exception occurs:

```
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 43494)
  File "/usr/lib/python3.4/socketserver.py", line 305, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/usr/lib/python3.4/socketserver.py", line 331, in process_request
    self.finish_request(request, client_address)
  File "/usr/lib/python3.4/socketserver.py", line 344, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.4/socketserver.py", line 665, in __init__
    self.handle()
  File "/usr/lib/python3.4/http/server.py", line 398, in handle
    self.handle_one_request()
  File "/usr/lib/python3.4/http/server.py", line 386, in handle_one_request
    method()
  File "/usr/local/lib/python3.4/dist-packages/oauth2client/tools.py", line 101, in do_GET
    self.wfile.write("<html><head><title>Authentication Status</title></head>")
  File "/usr/lib/python3.4/socket.py", line 391, in write
    return self._sock.send(b)
TypeError: 'str' does not support the buffer interface
----------------------------------------
```

Using b"" prefix for these strings is safe because the response is plain ASCII and it will also be compatible with Python 2.6 and 2.7

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -1,0 +1,30 @@
+"""Unit tests for oauth2client.tools."""
+
+import unittest
+from oauth2client import tools
+from six.moves.urllib import request
+import threading
+
+class TestClientRedirectServer(unittest.TestCase):
+    """Test the ClientRedirectServer and ClientRedirectHandler classes."""
+
+    def test_ClientRedirectServer(self):
+        # create a ClientRedirectServer and run it in a thread to listen
+        # for a mock GET request with the access token
+        # the server should return a 200 message and store the token
+        httpd = tools.ClientRedirectServer(('localhost', 0), tools.ClientRedirectHandler)
+        code = 'foo'
+        url = 'http://localhost:%i?code=%s' % (httpd.server_address[1], code)
+        t = threading.Thread(target = httpd.handle_request)
+        t.setDaemon(True)
+        t.start()
+        f = request.urlopen( url )
+        self.assertTrue(f.read())
+        t.join()
+        httpd.server_close()
+        self.assertEqual(httpd.query_params.get('code'),code)
+
+
+if __name__ == '__main__':
+    unittest.main()
+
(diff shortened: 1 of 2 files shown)


nathanielmanistaatgoogle:
Pleased to meet you and thank you for your change proposal! Two things:
(1) Have you signed the CLA (http://google.github.io/oauth2client/contributing.html#contributor-license-agreements)?
(2) Please add a test that fails without this change and passes with this change.

cskertesz:
(1) yes, I did
(2) I just realized that there is no unit test for the tools module. Should I add one for it with just a test for the ClientRedirectHandler class?

craigcitro:
@cskertesz yeah, making a new file would be awesome. thanks!

cskertesz:
I have added also a test for the ClientRedirectHandler class, which runs a redirect server, send a query with an access code and check if there is a valid response and if the stored access code matches the queried one.
Ran it with tox on python 2.7 and 3.4 (right now only have these), the test passes both with my change from previous commit, and fails on python 3.4 without it

tests/test_tools.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def test_ClientRedirectServer(self):
+        # create a ClientRedirectServer and run it in a thread to listen
+        # for a mock GET request with the access token
+        # the server should return a 200 message and store the token
+        httpd = tools.ClientRedirectServer(('localhost', 0), tools.ClientRedirectHandler)
+        code = 'foo'
+        url = 'http://localhost:%i?code=%s' % (httpd.server_address[1], code)
+        t = threading.Thread(target = httpd.handle_request)
+        t.setDaemon(True)
```
nathanielmanistaatgoogle, line 19: Why mark the thread as daemon?

kcs, line 19: I think I was just too overcautious, normally the thread crashes if there is an error, but I wanted to avoid hanging. Of course this doesn't make too much sense because of the join later on.

kcs, line 19: Thinking of it a little more, it helps if the main thread crashes before finishing connection to the handler thread, in which case the test app will not wait forever for the thread to close.

nathanielmanistaatgoogle:
Thank you for the contribution!

PR #129, opened 2015-02-11
Fall back to refreshing credentials if open_and_lock fails with ENOSYS o...

thobrla:
...r ENOLCK


nathanielmanistaatgoogle:
Thank you!

PR #133, opened 2015-02-16
Fix example to be Python3 compatible, use format()

RandomInsano:
Both print() and format() are compatible from 2.6. Also, format() is much nicer to use for internationalization since you can define the location of your substitutions. It works similarly to Java and .net's format() as well. Great stuff!

Should I tackle the other examples as well, or is piece meal all right?


RandomInsano:
CLA signed.

nathanielmanistaatgoogle:
Thank you for the contribution. I think piecemeal is just fine.

PR #134, opened 2015-02-17
Fall back to using an access_token for revocation.

### LINKED ISSUE #132: Revoking with the access token throws a TokenRevokeError
The program tries to revoke the access and refresh tokens of a Google Analytics service.
The error occurs in credentials.revoke(), in the following snippet of code:
```py
@require_GET
def logout(request):
    if request.app_user:
        from oauth2client.client import Credentials
        credentials = Credentials.new_from_json(request.app_user['credentials_json'])
        if not credentials.refresh_token:
            try:
                credentials.revoke(httplib2.Http())
            except Exception:
                # If credentials.revoke() doesn't work for some reason
                # use a plain http request as a fallback
                import requests
                response = requests.get(
                    credentials.revoke_uri + '?token=' + credentials.access_token)
        auth.logout(request)
        return HttpResponse()
    else:
        return HttpResponse(status=401)
```

The credentials.access_token is set and valid and the credentials.refresh_token is set to None. 
The credentials.revoke() causes an TokenRevokeError but when I want to do the same thing using the requests library, it works perfectly. 
```py
requests.get(credentials.revoke_uri + '?token=' + credentials.access_token)
``` 

To better understand what I'm saying, here is a screenshot of when the above snippet gets executed:
oauth2client-revoke-error

craigcitro:
According to the [OAuth2 docs](https://developers.google.com/accounts/docs/OAuth2WebServer#tokenrevoke), we can use either the refresh token or access token when revoking a token. If we've lost the refresh token for some reason, we should fall back to revoking via access token. (Note that if the access token has expired, this will still raise, which is the correct behavior.)

Fixes #132.

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -811,16 +811,16 @@ class OAuth2Credentials(Credentials):
       raise AccessTokenRefreshError(error_msg)
 
   def _revoke(self, http_request):
-    """Revokes the refresh_token and deletes the store if available.
+    """Revokes this credential and deletes the stored copy (if it exists).
 
     Args:
       http_request: callable, a callable that matches the method signature of
         httplib2.Http.request, used to make the revoke request.
     """
-    self._do_revoke(http_request, self.refresh_token)
+    self._do_revoke(http_request, self.refresh_token or self.access_token)
 
   def _do_revoke(self, http_request, token):
-    """Revokes the credentials and deletes the store if available.
+    """Revokes this credential and deletes the stored copy (if it exists).
 
     Args:
       http_request: callable, a callable that matches the method signature of
--- a/tests/test_oauth2client.py
+++ b/tests/test_oauth2client.py
@@ -560,6 +560,14 @@ class BasicCredentialsTests(unittest.TestCase):
         self, '400', revoke_raise=True,
         valid_bool_value=False, token_attr='refresh_token')
 
+  def test_token_revoke_fallback(self):
+    original_credentials = self.credentials.to_json()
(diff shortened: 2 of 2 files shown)


craigcitro:
PTAL @nathanielmanistaatgoogle

oauth2client/client.py
```diff
@@ -817,7 +817,8 @@ def _revoke(self, http_request):
       http_request: callable, a callable that matches the method signature of
         httplib2.Http.request, used to make the revoke request.
     """
-    self._do_revoke(http_request, self.refresh_token)
+    token = self.refresh_token or self.access_token
```
nathanielmanistaatgoogle: I don't think this field justifies its existence; just use "self.refresh_token or self.access_token" on the next line.

oauth2client/client.py
```diff
@@ -811,13 +811,13 @@ def _do_refresh_request(self, http_request):
       raise AccessTokenRefreshError(error_msg)
 
   def _revoke(self, http_request):
-    """Revokes the refresh_token and deletes the store if available.
+    """Revokes this token and deletes the stored copy (if available).
```
nathanielmanistaatgoogle: Is "this token" the right wording? Why not "these Oauth2Credentials"? You are referring to "this object", right, and the object is an Oauth2Credentials?

craigcitro: well, i don't think we have a context where a given `OAuth2Credentials` object would ever be mapped to more than one refresh token, right?

but in any event, i'm happy to tweak.

PR #136, opened 2015-02-21
Switched headers to be normalized as strings, not bytes, in keeping with httplib2.

foozlevazquez:
It appears that earlier contributors made the attempt to keep everything as bytes, mostly deducing from the test code. This doesn't work with the Python3 version of `httplib2`, and it is exactly this issue (`bytes` being passed to the Python3 version of `httplib2` instead of `str`) that causes [corrupt headers](https://github.com/GoogleCloudPlatform/gcloud-python/issues/653).  `httplib2` is [very clear](https://github.com/jcgregorio/httplib2/blob/df1cca48b204fc20fde1e49d7ad763394b01d931/python3/README#L8-L11) about [what it expects](https://github.com/jcgregorio/httplib2/wiki/Examples-Python3).

Given the dependency on `httplib2`, in `oauth2client`, it makes sense to me that `oauth2client` should follow its lead and use whatever version of `str` for headers makes sense in the version of Python it is using.
- Fixed header strings to use strings and not bytes.
- Fixed tests to work with Python3

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -397,9 +397,32 @@ class Storage(object):
     finally:
       self.release_lock()
 
+# Since strings ARE unicode in Python3, "cleaning" the string doesn't mean
+# just converting to str.
+#
+#  As per the httplib2 documentation:
+#
+#    ** THE RESPONSE HEADERS ARE STRINGS, BUT THE CONTENT BODY IS BYTES **
+#
+def _clean_header(s):
+  """Always returns something of type str.  Raise UnicodeEncodeError if
+  unconvertable to ascii."""
+
+  if not isinstance(s, str):
+    # str(b'foo') will return "b'foo'" in Py3, not what we want.
+    if isinstance(s, bytes):
+      # Binary string in Py3
+      s = s.decode('utf-8')
+    else:
+      s = str(s)
+
+  # We're trying to generate the UnicodeEncodeError here if the string is
+  # unconvertable, AND keep this a string in both Py2 and Py3.
+  s = str((s.encode('ascii')).decode('utf-8'))
+  return s
 
 def clean_headers(headers):
-  """Forces header keys and values to be strings, i.e not unicode.
+  """Forces header keys and values to be strings suitable for httplib2.
 
   The httplib module just concats the header keys and values in a way that may
   make the message header a unicode string, which, if it then tries to
@@ -414,8 +437,8 @@ def clean_headers(headers):
   clean = {}
   try:
     for k, v in six.iteritems(headers):
-      clean_k = k if isinstance(k, bytes) else str(k).encode('ascii')
-      clean_v = v if isinstance(v, bytes) else str(v).encode('ascii')
+      clean_k = _clean_header(k)
+      clean_v = _clean_header(v)
       clean[clean_k] = clean_v
   except UnicodeEncodeError:
     raise NonAsciiHeaderError(k + ': ' + v)
--- a/tests/test_appengine.py
+++ b/tests/test_appengine.py
@@ -29,7 +29,11 @@ import os
 import time
 import unittest
 import urllib
-import urlparse
+
+try:
+  import urlparse
+except ImportError:
+  from urllib.parse import urlparse
 
 import dev_appserver
 dev_appserver.fix_sys_path()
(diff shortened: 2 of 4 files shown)


craigcitro:
isn't this the same issue as https://github.com/google/oauth2client/pull/104 ?

foozlevazquez:
@Craigcitro #104 is the same root problem, but the conclusion that the bug lies in httplib2 is mistaken:  httplib2 clearly expects strings and not bytes in headers, so this is not a bug with httplib2.

tests/test_appengine.py
```diff
@@ -29,7 +29,11 @@
 import time
 import unittest
 import urllib
-import urlparse
+
+try:
+  import urlparse
+except ImportError:
+  from urllib.parse import urlparse
```
dhermes, line 36: You should use `six.moves`.

oauth2client/client.py
```diff
@@ -397,9 +397,32 @@ def delete(self):
...
 
+# Since strings ARE unicode in Python3, "cleaning" the string doesn't mean
+# just converting to str.
+#
+#  As per the httplib2 documentation:
+#
+#    ** THE RESPONSE HEADERS ARE STRINGS, BUT THE CONTENT BODY IS BYTES **
+#
+def _clean_header(s):
+  """Always returns something of type str.  Raise UnicodeEncodeError if
```
dhermes, line 408: Stay under 80 chars.

oauth2client/client.py
```diff
@@ -397,9 +397,32 @@ def delete(self):
     finally:
       self.release_lock()
 
+# Since strings ARE unicode in Python3, "cleaning" the string doesn't mean
+# just converting to str.
+#
+#  As per the httplib2 documentation:
+#
+#    ** THE RESPONSE HEADERS ARE STRINGS, BUT THE CONTENT BODY IS BYTES **
+#
```
dhermes, line 406: This comment should be in the docstring.

dhermes:
I would argue #104 is "on the right track" and that `httplib2` should handle this.

cganterh:
Should I expect this to be solved soon or should I find a workaround?
This is delaying my project. If I could help please say so.

dhermes:
@cganterh This is only needed for Python 3, if that helps.

cganterh:
@dhermes my whole project is based on Python 3 and is strongly dependent on Python 3 libraries. Is it a good idea to fork oauth2client or httplib2 and apply the pull requests myself so I can go on while this issue is being fixed?

nathanielmanistaatgoogle:
@cganterh: Why fork httplib2? What change would you want to make to httplib2 that hasn't been accepted by the httplib2 maintainers?

cganterh:
Correct me if im wrong, but I think that #136, #104 and jcgregorio/httplib2#291 are trying to solve the same issue. However I haven't read the pull requests in depth.
In jcgregorio/httplib2#291 they describe exactly the same error I'm getting in my code. The reason I posted here is that this pull request was more active than the others.
Sorry if I'm completely wrong, I'm using oauth2client for the first time.

nathanielmanistaatgoogle:
Okay, I suspected that they were all the same thing. It looks like getting a fix into httplib2 is blocked on writing test coverage for the fix, which is something that anyone can do.

cganterh:
Yes, I know. But I think I'm not the most appropriate person to do this, since I have little experience on writing tests and it is my first time using the library. So I thought it was a bad idea to do it myself and maybe a better one was to just use some of the pull requests while the problem was actually being fixed. Now I think it doesn't make sense to fork the whole library rather than using the fork that the author of the pull request already made.

dhermes:
@anthmgoogle I don't think @jcgregorio is actively maintaining `httplib2`, so it may be best to just get the support in `oauth2client`. Maybe I am mistaken though.

cganterh:
@dhermes I think it's a good idea, but I don't know what is stopping #136 and #104 to be merged. On the other hand, fixing httplib2 would be better because it may be an improvement for other libraries that depend on httplib2.

nathanielmanistaatgoogle:
All that's stopping `httplib2` from having the change is that the change has to come with unit test coverage. That's not-at-all an unmaintained project. :-)

cganterh:
So I finally decided to write the unit test for jcgregorio/httplib2#291.
@nathanielmanistaatgoogle you were right, it was actually easy! (If I did it right ...)

cganterh:
My pull request was accepted. Someone should see if this solves #136 or #104.

nathanielmanistaatgoogle:
What action remains to be taken in this pull request now that the fix in httplib2 has been made?

dhermes:
@nathanielmanistaatgoogle I am going through this to see if there is anything we missed in #250

PR #138, opened 2015-02-25
fix bytearray conversion without encoding

obayemi:
fix bytearray conversion without encoding in validate_token l.117

### DIFF
--- a/oauth2client/xsrfutil.py
+++ b/oauth2client/xsrfutil.py
@@ -24,6 +24,7 @@ __authors__ = [
 import base64
 import hmac
 import time
+import sys
 
 import six
 from oauth2client import util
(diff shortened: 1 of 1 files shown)


nathanielmanistaatgoogle:
Thanks for signing the CLA.

Please amend your commit message to be in line with current guidance on commit metadata: http://chris.beams.io/posts/git-commit/#seven-rules.

What about test coverage for this fix?

nathanielmanistaatgoogle:
Please rewrite your commit log message to both conform to http://chris.beams.io/posts/git-commit/#seven-rules and to describe the defect being fixed by this change.

Please also look into this change's failure on Travis and draft a new version that passes existing tests.

Also what about test coverage for this fix itself? Are you able to write a test that fails without this change and passes with this change?

dhermes:
Closing due to staleness.

@obayemi Feel free to re-open if you get a chance. I am interested in when this change was needed. It seems the token will always be ascii (since it's mostly just base64) so I'm unclear on when that'd be needed.

PR #143, opened 2015-03-10
Adds support for Google Developer Shell session credentials

aalexand:


oauth2client/devshell.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def _SendRecv():
+  """Communicate with the Developer Shell server socket."""
+
+  port = int(os.getenv(DEVSHELL_ENV, 0))
+  if port == 0:
+    raise NoDevshellServer()
+
+  # pylint:disable=g-import-not-at-top, Delay for performance.
```
skelterjohn: pylint comments can go

oauth2client/devshell.py
```diff
@@ -0,0 +1,139 @@
...
+    * Index 1 - default project ID. None if the project context is not known.
+    * Index 2 - OAuth2 access token. None if there is no valid auth context.
+  """
+
+  def __init__(self, json_string):
+    """Initialize the response data from JSON PBLite array."""
+    pbl = json.loads(json_string)
+    if not isinstance(pbl, list):
+      raise ValueError('Not a list: ' + str(pbl))
+    self.user_email = pbl[0] if len(pbl) else None
```
skelterjohn: error if len(pbl) != 3?

skelterjohn:
LGTM. regarding CLA, have you registered as a github googler?

oauth2client/devshell.py
aalexand: It's in the proto spirit to ignore extra fields so I did that.

aalexand:
> have you registered as a github googler?

I think I have now.

oauth2client/devshell.py
aalexand: Done

nathanielmanistaatgoogle:
(I'll be taking a look at this at some point.)

oauth2client/devshell.py
nathanielmanistaatgoogle: Define a local field rather than computing len(pbl) three times, please?

nathanielmanistaatgoogle: Use an explicitly numeric comparison to match the following two lines.

oauth2client/devshell.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def _SendRecv():
+  """Communicate with the Developer Shell server socket."""
+
+  port = int(os.getenv(DEVSHELL_ENV, 0))
+  if port == 0:
+    raise NoDevshellServer()
+
+  import socket
```
nathanielmanistaatgoogle, line 72: What the reason for having this import not at top-level at the beginning of the file?

oauth2client/devshell.py
```diff
@@ -0,0 +1,138 @@
...
+
+  s = socket.socket()
```
nathanielmanistaatgoogle: "Avoid single-character names except for counters or iterators": https://google-styleguide.googlecode.com/svn/trunk/pyguide.html?showone=Naming#Naming

oauth2client/devshell.py
```diff
@@ -0,0 +1,138 @@
...
+  s.connect(('localhost', port))
+
+  data = CREDENTIAL_INFO_REQUEST_JSON
+  msg = '%s\n%s' % (len(data), data)
+  s.sendall(msg.encode())
+
+  resp_buffer = ''
```
nathanielmanistaatgoogle: The value assigned on this line doesn't appear to ever be used. Is this dead code that can simply be struck?

nathanielmanistaatgoogle: Please avoid abbreviations when you can. Should this field be named "response_buffer"?

oauth2client/devshell.py
```diff
@@ -0,0 +1,138 @@
...
+  resp_1 = s.recv(6).decode()
```
nathanielmanistaatgoogle: "resp_1" is way too weird a name - unless that's the name used by the protocol for the datum stored within the field?

Why not just "first_six_response_bytes"?

oauth2client/devshell.py
```diff
@@ -0,0 +1,138 @@
...
+  if '\n' not in resp_1:
+    raise CommunicationError('saw no newline in the first 6 bytes')
+  nstr, extra = resp_1.split('\n', 1)
+  resp_buffer = extra
+  n = int(nstr)
```
nathanielmanistaatgoogle: Single-letter identifier.

oauth2client/devshell.py
```diff
@@ -0,0 +1,138 @@
...
+  data = CREDENTIAL_INFO_REQUEST_JSON
+  msg = '%s\n%s' % (len(data), data)
+  s.sendall(msg.encode())
+
+  resp_buffer = ''
+  resp_1 = s.recv(6).decode()
+  if '\n' not in resp_1:
+    raise CommunicationError('saw no newline in the first 6 bytes')
+  nstr, extra = resp_1.split('\n', 1)
+  resp_buffer = extra
```
nathanielmanistaatgoogle: It looks like you're making this assignment too early. What about moving this assignment to an "else" block of the "if" on lines 88-89?

oauth2client/devshell.py
```diff
@@ -0,0 +1,138 @@
...
+        None,  # access_token, initialized below
+        None,  # client_id
+        None,  # client_secret
+        None,  # refresh_token
+        None,  # token_expiry
+        None,  # token_uri
+        user_agent)
+    self._refresh(None)
+
+  def _refresh(self, http):
```
nathanielmanistaatgoogle: This smells like a bug... that turns out not to matter but should probably still be fixed.

DevshellCredentials inherits the method Oauth2Credentials.refresh which takes an httplib2.Http object and the method Oauth2Credentials._refresh which takes "a callable that matches the method signature of httplib2.Http.request" and is named "http_request". In this method definition you've taken the method name of one and the parameter name of the other. Which do you actually mean to override?

(Unless I've made a mistake, which is entirely possible.)

oauth2client/devshell.py
```diff
@@ -0,0 +1,138 @@
...
+    self.devshell_response = _SendRecv()
+    self.access_token = self.devshell_response.access_token
```
nathanielmanistaatgoogle, line 117: Where does this instance-scope field get used?

nathanielmanistaatgoogle:
Please squash your commits.

oauth2client/devshell.py
aalexand, line 72: To defer loading the socket package if this kind of credentials is detected unavailable and NoDevshellServer is raised.

oauth2client/devshell.py
aalexand: Renamed to 'header'.

oauth2client/devshell.py
aalexand: Renamed.

oauth2client/devshell.py
aalexand: 'extra' contains the beginning of the message, so we copy that here.  Maybe I'm missing the point.  The code has changed a bit anyway, please re-comment once I update the PR in a bit.

oauth2client/devshell.py
aalexand, line 117: It is used by clients of OAuth2Credentials class.

oauth2client/devshell.py
aalexand: It is supposed to override the impl _refresh(http_request) method.  Renamed the parameter.

oauth2client/devshell.py
aalexand: Renamed.

oauth2client/devshell.py
skelterjohn, line 72: The original source of this code, inside gcloud, necessitated delayed imports for performance reasons. Most programs will have a running time that dwarfs initialization: not so with gcloud. importing socket can take some time.

nathanielmanistaatgoogle:
I'm satisfied (thank you for the changes). Will anyone else be reviewing this before it goes in?

skelterjohn:
changes LGTM as well. let me know if you consider this ready to merge.

nathanielmanistaatgoogle:
I consider this ready to merge.

Last call for obstacles, blockers, and objections?

aalexand:
I think it should be merged now.

PR #148, opened 2015-03-25
Making save_to_well_known_file() save with 0600 permission.

### LINKED ISSUE #144: Well known file is created with inapproriate permission
When I run "gcloud auth login", the file ${HOME}/.config/gcloud/application_default_credentials.json is created with a permission 0644, which seems inappropriate in my opinion. I think it should be 0600 or something stricter.

dhermes:
Fixes #144.

H/T to http://stackoverflow.com/a/5624691/1068170 and
       http://stackoverflow.com/a/5337329/1068170

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -28,7 +28,9 @@ import logging
 import os
 import socket
 import sys
+import tempfile
 import time
+import shutil
 import six
 from six.moves import urllib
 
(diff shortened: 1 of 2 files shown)


anthmgoogle:
LGTM.

Do you need someone to pull this request in?

dhermes:
Yes I don't have merge privileges. But the build failed
https://travis-ci.org/google/oauth2client/jobs/55736796

I need to update so that the correct octal syntax is used for Python 3: `0o0600`

dhermes:
@anthmgoogle I just updated the PR with the fix.

Waiting on new build: https://travis-ci.org/google/oauth2client/builds/55836779

anthmgoogle:
There is a manual test step that should be done here also:
1. Take a sample using the get_application_default auth feature to call a Google API.
2. Use the Google Cloud SDK and run gcloud_auth_login and confirm the prompt.
3. Verify that this enables the sample to work. It should pick up the credential file saved gcloud auth login.
4. Patch this change into the copy of this code in the Cloud SDK.
5. Repeat steps 2-3 to verify there is no regression.

dhermes:
Good call! I discovered it doesn't over-write permissions of an existing file. Will update and create a corresponding unit test.

dhermes:
@anthmgoogle I edited the version of `oauth2client.client` in the SDK and confirmed this all works as expected. New unit test and code change in the second commit just pushed.

anthmgoogle:
Revision LGTM.

dhermes:
Great @craigcitro care to merge?

nathanielmanistaatgoogle:
@dhermes Squash commits? It's not clear why there should be two commits for this change; it looks like it's only an artifact of the trajectory this code took through review.

dhermes:
SGTM @nathanielmanistaatgoogle 

I like the way you think! (I try to be the "fewer commits" stickler, but usually get shot down.)

PR #149, opened 2015-03-25
Using fake ~/.config/gcloud directory for failing defaults test.

### LINKED ISSUE #123: Mock out implicit behavior so that test_get_application_default_environment_not_set_up always passes
Fails on some Ubuntu 14.04 LTS machines for both py27 and py34.

dhermes:
- Partially fixes #123.
- Still doesn't address the case of running in GAE (unlikely) or
  GCE (fairly likely). Could also address these by mocking the
  environment detectors.


anthmgoogle:
LGTM.

anthmgoogle:
I don't have permission to merge this. @craigcitro or @nathanielmanistaatgoogle should make the call.

PR #153, opened 2015-03-29
Adding system tests to make sure token exchange works as expected.

dhermes:
Fixes #127

PR #156, opened 2015-03-30
Paring down unused keys in JSON credentials files.

dhermes:
Previously the User and Service Account credentials files had the some contents even though they have different dictionary keys used for the credentials properties.

/cc @nathanielmanistaatgoogle 

Spinning off from previous PR based from https://github.com/google/oauth2client/pull/153#issuecomment-87793610


anthmgoogle:
LGTM if all the tests pass. This is much closer to how the files appear in real scenarios.

nathanielmanistaatgoogle:
Tests pass; merging based on @anthmgoogle's approval.

Issue #150, opened 2015-03-27
issue with AppAssertionCredentials and Python3


ptone:
seem to be a few edge cases for Python3 - seems similar to #106 

https://gist.github.com/ptone/7e45323e91097ed3cc2f

theacodes:
Thank you for creating this issue, however, this project is deprecatedand we will only be addressing critical security issues. You can read moreabout this deprecation [here](http://google-auth.readthedocs.io/en/latest/oauth2client-deprecation.html).

If you need support or help using this library, we recommend that you ask yourquestion on StackOverflow.

If you still think this issue is relevant and should be addressed, pleasecomment and let us know!

PR #152, opened 2015-03-28
py3 fix for AppAssertionCredentials on GCE

ptone:
fixes #150

PR #157, opened 2015-03-31
utf-8 decode bytes inputs for python3 compat + python3 test fix

saaros:
- json loading: always utf-8 decode `bytes` inputs
- tests: use six.moves.urllib for py2/3 compatibility

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -272,7 +272,7 @@ class Credentials(object):
       An instance of the subclass of Credentials that was serialized with
       to_json().
     """
-    if six.PY3 and isinstance(s, bytes):
+    if isinstance(s, bytes):
       s = s.decode('utf-8')
     data = json.loads(s)
     # Find and call the right classmethod from_json() to restore the object.
@@ -612,7 +612,7 @@ class OAuth2Credentials(Credentials):
     Returns:
       An instance of a Credentials subclass.
     """
-    if six.PY3 and isinstance(s, bytes):
+    if isinstance(s, bytes):
       s = s.decode('utf-8')
     data = json.loads(s)
     if (data.get('token_expiry') and
@@ -780,7 +780,7 @@ class OAuth2Credentials(Credentials):
     logger.info('Refreshing access_token')
     resp, content = http_request(
         self.token_uri, method='POST', body=body, headers=headers)
-    if six.PY3 and isinstance(content, bytes):
+    if isinstance(content, bytes):
       content = content.decode('utf-8')
     if resp.status == 200:
       d = json.loads(content)
@@ -840,6 +840,8 @@ class OAuth2Credentials(Credentials):
     query_params = {'token': token}
     token_revoke_uri = _update_query_params(self.revoke_uri, query_params)
     resp, content = http_request(token_revoke_uri)
+    if isinstance(content, bytes):
+      content = content.decode('utf-8')
     if resp.status == 200:
       self.invalid = True
     else:
@@ -907,7 +909,7 @@ class AccessTokenCredentials(OAuth2Credentials):
 
   @classmethod
   def from_json(cls, s):
-    if six.PY3 and isinstance(s, bytes):
+    if isinstance(s, bytes):
       s = s.decode('utf-8')
     data = json.loads(s)
     retval = AccessTokenCredentials(
@@ -1494,6 +1496,8 @@ class SignedJwtAssertionCredentials(AssertionCredentials):
 
   @classmethod
   def from_json(cls, s):
+    if isinstance(s, bytes):
+      s = s.decode('utf-8')
     data = json.loads(s)
     retval = SignedJwtAssertionCredentials(
         data['service_account_name'],
@@ -1558,8 +1562,10 @@ def verify_id_token(id_token, audience, http=None,
 
   resp, content = http.request(cert_uri)
 
+  if isinstance(content, bytes):
+    content = content.decode('utf-8')
   if resp.status == 200:
-    certs = json.loads(content.decode('utf-8'))
+    certs = json.loads(content)
     return crypt.verify_signed_jwt_with_certs(id_token, certs, audience)
   else:
     raise VerifyJwtTokenError('Status code: %d' % resp.status)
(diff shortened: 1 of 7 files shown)


oauth2client/client.py
```diff
@@ -272,7 +272,7 @@ def new_from_json(cls, s):
       An instance of the subclass of Credentials that was serialized with
       to_json().
     """
-    if six.PY3 and isinstance(s, bytes):
```
nathanielmanistaatgoogle, line 275: I'm not sure I understand this change - why drop the six.PY3 conjunct? Under what circumstances will s be of type bytes? Should the method's doc string be updated to explicitly allow bytes to be passed? As a lifelong opponent of ad hoc polymorphism, help me understand why explicit support for s being either a string or bytes is a good thing?

saaros, line 275: json input should always be an unicode string, this just makes sure we always utf-8 decode the input before passing it to json.loads instead of having json.loads (on python 2) deal (and possibly fail) with the encoding.

I'd say the code should (mostly) just handle unicode strings, but for compatibility it's easier to allow utf-8 encoded bytestring inputs.  But we should probably implement a utf-8-bytes-accepting json decoding function in util.py and use it everywhere instead of duplicating the type check/decode/json load code everywhere.

saaros:
This PR fixes #150 which also had a proposed but slightly broken fix in #152.

oauth2client/client.py
dhermes, line 275: `six.binary_type` is typically used over `bytes`

saaros, line 275: `bytes` is supported in Python 2.6 and newer, `six.binary_type` only makes sense if older versions need to be supported.

saaros:
Added a commit to convert the last remaining uses of `six.binary_type` to `bytes` which was already being used in most places.

sullivanmatt:
+1

Are there any other considerations around this PR that are keeping it from being merged?  I'm completely unable to use Python 3 in my infrastructure solely due to this issue.

dhermes:
@sullivanmatt Have you updated `httplib2`? Is it still an issue with the latest `httplib2` (which now does support Python 3 correctly, and fixed many issues here)?

sullivanmatt:
I'm using httplib2 0.9.1 (latest).

Here's the error I believe this would resolve (haven't checked out the branch to test, I admit) -

```
access_token = get_access_token_from_cache()
credentials = AccessTokenCredentials(access_token, user_agent="httpclient/1.0")
http = httplib2.Http()
http = credentials.authorize(http)


  File "/opt/project-service/project_service/gproject_service/__init__.py", line 160, in _get_history
    labelId="INBOX").execute(http=http)
  File "/home/msulliv/mb-project-sdk-py3/lib/python3.4/site-packages/oauth2client/util.py", line 137, in positional_wrapper
  File "/home/msulliv/mb-project-sdk-py3/lib/python3.4/site-packages/googleapiclient/http.py", line 722, in execute
    body=self.body, headers=self.headers)
  File "/home/msulliv/mb-project-sdk-py3/lib/python3.4/site-packages/oauth2client/util.py", line 137, in positional_wrapper
  File "/home/msulliv/mb-project-sdk-py3/lib/python3.4/site-packages/oauth2client/client.py", line 549, in new_request
    self.apply(headers)
  File "/home/msulliv/mb-project-sdk-py3/lib/python3.4/site-packages/oauth2client/client.py", line 615, in apply
    headers['Authorization'] = 'Bearer ' + self.access_token
TypeError: Can't convert 'bytes' object to str implicitly
```

dhermes:
@sullivanmatt Your stacktrace doesn't seem to match the code snippet.

After running `pip3 install -U oauth2client` (which also updates the dependencies) and running a related snippet with both `bytes` and `unicode` I can reproduce the failure

```python
$ python3
Python 3.4.0 (default, Apr 11 2014, 13:05:11) 
[GCC 4.8.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from oauth2client.client import AccessTokenCredentials
>>> 
>>> 
>>> creds_unicode = AccessTokenCredentials(u'foo', user_agent='UA')
>>> headers = {}
>>> creds_unicode.apply(headers)
>>> print(headers)
{'Authorization': 'Bearer foo'}
>>> 
>>> creds_bytes = AccessTokenCredentials(b'bar', user_agent='UA')
>>> headers = {}
>>> creds_bytes.apply(headers)
  File "/usr/local/lib/python3.4/dist-packages/oauth2client/client.py", line 615, in apply
    headers['Authorization'] = 'Bearer ' + self.access_token
TypeError: Can't convert 'bytes' object to str implicitly
```

sullivanmatt:
For whatever reason I didn't realize this until I saw your reproduction above, but simply doing `access_token.decode('utf-8')` was sufficient to get around the issue for the time being.

dhermes:
@sullivanmatt Do you mind if we close out. A lot of the changes are just `bytes` "grammar" changes.

But there are some places where you add explicit decodes:

```python
    resp, content = http_request(token_revoke_uri)
    if isinstance(content, bytes):
      content = content.decode('utf-8')
...
  def from_json(cls, s):
    if isinstance(s, bytes):
      s = s.decode('utf-8')
...
  if isinstance(content, bytes):
    content = content.decode('utf-8')
  if resp.status == 200:
    # WAS certs = json.loads(content.decode('utf-8'))
    certs = json.loads(content)
```

How did you go about determining these were needed?

sullivanmatt:
I am not the author of this PR (@saaros is), so I can't comment on his methodology.

dhermes:
D'oh! The passage of time has made me weak and dumb :)

sullivanmatt:
We've all been there man.  Thanks!

saaros:
It's to make sure the input to json.loads() is always an unicode string; it's not always clear where the input comes from and whether it's `bytes` or unicode.  Having `if isinstance(s,bytes): s = s.decode("utf-8")` in every place that loads json isn't pretty, though.  A function in util.py to do it would probably look cleaner.

dhermes:
I agree with you there. Do you mind breaking this into two PRs: one with the `bytes` grammar fix and the other with the actual code path fixes. (Or I can do it and CC you on the PRs.)

Also, how did you locate all the code paths you edited?

saaros:
I just ran `git grep` to find them after I ran into various problems using this library on Python 3.  Do you want the commit 3584b074674a with the `six.binary_type` -> `bytes` change in a PR of its own leaving the other two ("fix urllib tests on py3" and "utf8-decode before json.loads") in this PR?

dhermes:
- The `bytes` grammar stuff I'm referring to is removing `six.PY3` and just using `bytes` instead of `six.binary_type`.
- The "fix urllib tests on py3" is somewhat moot since App Engine only supports Py27
- utf8-decode before `json.loads` is fine in this PR but seems worth just closing and opening a new one given your suggestion for moving the functionality to `utils`.

dhermes:
@nathanielmanistaatgoogle I am going through this to see if there is anything we missed in #250

dhermes:
Closing since #272 is in

PR #158, opened 2015-04-03
Check whether a credential loaded from a store is expired before using it.

mattmoor:


nathanielmanistaatgoogle:
What hope is there for test coverage for this change?

mattmoor:
Technically you were already covering it with bad tests. ;-)

I fixed the break and added testing of an actual refresh.

mattmoor:
ping?

nathanielmanistaatgoogle:
Please tweak your commit message to be in accordance with http://chris.beams.io/posts/git-commit/#seven-rules.

Code looks reasonable to me, but @anthmgoogle or @soltanmm, would you mind being a second pair of eyes and also reviewing?

mattmoor:
Should be done.

nathanielmanistaatgoogle:
Commit message is much better; thanks.

thobrla:
Created a updated pull request based on this logic at https://github.com/google/oauth2client/pull/173 that also covers the case where the token in the store is about to expire.

craigcitro:
this was actually included in #173 (i think we needed to mention that in the commit to magically close this?).

PR #159, opened 2015-04-08
Deferring OpenSSL import until usage.

dhermes:
This is to speed up import times. In OpenSSL 0.14 the import takes
0.5 seconds due to cffi on-demand build of extensions in the
cryptography library.

For classes and functions which are conditionally defined based on
the existence of OpenSSL.crypto, we check that the module
exists (but don't import it) using imp.find_module.

See pyca/pyopenssl#137 for context (originally discovered by me because of google/apitools#14).

### DIFF
--- a/oauth2client/crypt.py
+++ b/oauth2client/crypt.py
@@ -16,8 +16,10 @@
 """Crypto-related routines for oauth2client."""
 
 import base64
+import imp
 import json
 import logging
+import os
 import sys
 import time
 
@@ -37,7 +39,9 @@ class AppIdentityError(Exception):
 
 
 try:
-  from OpenSSL import crypto
+  _, _package_dir, _ = imp.find_module('OpenSSL')
+  if not os.path.isfile(os.path.join(_package_dir, 'crypto.py')):
+    raise ImportError('No module named OpenSSL')
 
   class OpenSSLVerifier(object):
     """Verifies the signature on a message."""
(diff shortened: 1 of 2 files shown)


dhermes:
Is there anyone to review this?

/cc @nathanielmanistaatgoogle @anthmgoogle

anthmgoogle:
I took a look, but I don't think I know this part of the code or the language subtleties well enough to be the sole approver. I would suggest  @nathanielmanistaatgoogle, but possibly also @craigcitro or @jcgregorio.

oauth2client/crypt.py
```diff
@@ -37,7 +39,10 @@ class AppIdentityError(Exception):
 
 
 try:
-  from OpenSSL import crypto
+  _, package_dir, _ = imp.find_module('OpenSSL')
```
craigcitro: maybe just call it `_package_dir`, and not worry about deleting it?

(in particular, i think the variable hangs around in the event of the exception two lines down -- so if you want to leave it public, maybe `del` there, too?)

craigcitro:
one little thing, otherwise LGTM

oauth2client/crypt.py
dhermes: I'm going to go with `_package_dir`. Will ping you when rebased.

dhermes:
@craigcitro I `git push -f`ed with suggested change.

Feel free to take a second look.

I don't have merge privileges so someone will need to use your LGTM for me.

craigcitro:
@dhermes yeah, i don't think we have an easy way to give non-googlers push rights on this repo.

merging now.

PR #161, opened 2015-04-10
DO NOT MERGE; Add JWT header auth

soltanmm:
cc @tbetbetbe
@anthmgoogle Is this implementation correct? Is it organized for what we want out of Python? What should the workflow be for user code to get instances of JWT-header-credentials? What should I be testing for it?


oauth2client/service_account.py
```diff
@@ -122,6 +122,82 @@ def create_scoped(self, scopes):
                                       **self._kwargs)
 
 
+class _ServiceAccountJwtHeaderCredentials(AssertionCredentials):
+  """Class representing a service account (signed JWT) credential."""
+
+  MAX_TOKEN_LIFETIME_SECS = 3600 # 1 hour in seconds
+
+  def __init__(self, service_account_id, service_account_email, private_key_id,
+               private_key_pkcs8_text, scopes, user_agent=None,
```
anthmgoogle: For the JwtHeader credential, I would expect that scopes should not be here either as an argument to the type or as a field in the Jwt.

anthmgoogle:
Looking good. I would expect:
- Scopes should not be present as an argument or in the JWT.
- Needs unit testing. The small decrease in coverage is probably related to that.

soltanmm:
What's the workflow for the user to actually *use* the implementation, though? As written, it's entirely inaccessible.

oauth2client/service_account.py
```diff
@@ -122,6 +122,81 @@ def create_scoped(self, scopes):
...
+
+    now = int(time.time())
+    jwt_token = {
+        'aud': self._token_uri,
+        'iat': now,
+        'exp': now + _ServiceAccountCredentials.MAX_TOKEN_LIFETIME_SECS,
+        'iss': self._service_account_email,
+        'sub': self._service_account_email
+    }
+    signed_jwt_token = self.sign_blob(_urlsafe_b64encode(jwt_token))
```
soltanmm: What is the proper order of token -> encode -> sign -> place-in-header -> encode?

Currently the encoding procedure complains about invalid start bytes, and I imagine that has to do with signing making the token an arbitrary byte string rather than proper unicode.

anthmgoogle:
The scenario for use is with gRPC initially. If the transport is gRPC it should be able to use this credential explicitly, and the One Platform entry points should be able to accept this credential as authorization without needing other authorization. The pattern we are encouraging is that if you get a service account credential with missing or empty scopes (that would otherwise fail) the gRPC transport should "promote" to this credential type that does not need scopes and should have better performance and security properties.

soltanmm:
Closing this as it's old, stale, and was implemented based on a misunderstanding of the control flow.

cc #253

PR #162, opened 2015-04-10
Allowing custom directory for well-known file.

### LINKED ISSUE #151: application default credentials don't respect CLOUDSDK_CONFIG
application default credentials should either be written to "$CLOUDSDK_CONFIG/application_default_credentials.json", with CLOUDSDK_CONFIG defaulting to "~/.config/gcloud".

gcloud promises that it won't write any credential information outside of $CLOUDSDK_CONFIG.

dhermes:
Also raising exception if the directory does not exist.

Fixes #151.

### DIFF
--- a/tests/test_devshell.py
+++ b/tests/test_devshell.py
@@ -125,6 +125,11 @@ class DevshellCredentialsTests(unittest.TestCase):
       self.assertEqual('sometoken', creds.access_token)
 
   def test_refuses_to_save_to_well_known_file(self):
-    with _AuthReferenceServer():
-      creds = DevshellCredentials()
-      self.assertRaises(NotImplementedError, save_to_well_known_file, creds)
+    ORIGINAL_ISDIR = os.path.isdir
+    try:
+      os.path.isdir = lambda path: True
+      with _AuthReferenceServer():
+        creds = DevshellCredentials()
+        self.assertRaises(NotImplementedError, save_to_well_known_file, creds)
+    finally:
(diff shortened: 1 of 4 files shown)


tests/test_devshell.py
```diff
@@ -125,6 +125,11 @@ def test_handles_ignores_extra_fields(self):
       self.assertEqual('sometoken', creds.access_token)
 
   def test_refuses_to_save_to_well_known_file(self):
-    with _AuthReferenceServer():
-      creds = DevshellCredentials()
-      self.assertRaises(NotImplementedError, save_to_well_known_file, creds)
+    ORIGINAL_ISDIR = os.path.isdir
+    try:
+      os.path.isdir = lambda path: True
```
soltanmm, line 130: Was there any particular reason for reassigning `os.path.isdir` as opposed to changing the cloud SDK environment variable (say, to a temporary directory)?

dhermes, line 130: I figured it was easier than trying to find a directory which really does exist.

Also it fit in with other changes where we needed a particular directory to not exist (also harder to find on a generic system, at least harder than make `os.path.isdir` return `False`)

anthmgoogle:
LGTM.

dhermes:
Thanks for review. Blocking on testing remark or no? Either way I don't have merge privileges.

anthmgoogle:
I don't have write permission either. @nathanielmanistaatgoogle or @craigcitro may be able to help.

dhermes:
Thanks

PR #167, opened 2015-04-20
In OAuth2Credentials, only pass the first argument as positional.

ryanli:
So that it doesn't print an warning about extra positional arguments if a request takes at most 1 positional argument (e.g. one that is already authorized by OAuth2Credentials.authorize).


dhermes:
Thanks for your contribution!

This was addressed (in a different way) in #196.

Sorry we missed this before #196 came through.

PR #171, opened 2015-05-03
Final iteration to make system tests pass.

dhermes:
Two main things happened here:
1. Manually encrypted files instead of using `travis encrypt-file`
2. Removed OAUTH2CLIENT_TEST_USER_KEY_PATH fallback.

The first was because `travis encrypt-file` was failing mysteriously.
To make up, I ran

```
openssl enc -d -a -md sha1 -aes-256-cbc -nosalt -p
```

and typed random noise into the keyboard to create a candidate KEY and
IV (H/T http://superuser.com/a/471524). After doing this, they were set
via

```
travis env set OAUTH2CLIENT_KEY "..KEY.." --repo "google/oauth2client"
travis env set OAUTH2CLIENT_IV "..IV.." --repo "google/oauth2client"
```

To actually do the encryption locally:

```
openssl aes-256-cbc -K "..KEY.." \
    -iv "..IV.." \
    -in file-to-encrypt \
    -out file-to-encrypt.enc -e
```

As for the second change, OAUTH2CLIENT_TEST_USER_KEY_PATH was previously
allowed to fallback to client._get_well_known_file() in system tests, but this
is a problem since that function throws an OSError when the directory does not
exist. (It does not exist on Travis.)


dhermes:
@nathanielmanistaatgoogle I gave up on `travis encrypt-file` and just did the encryption manually.

Merged this so we can make sure it works on merged requests.

dhermes:
Houston we are a go!
https://travis-ci.org/google/oauth2client/builds/61075708

Success.

nathanielmanistaatgoogle:
Hooray!

dhermes:
But what the heck `travis encrypt-file`?

PR #173, opened 2015-05-11
Avoid using expired tokens from credential storage.

thobrla:
This covers the case where credential storage contains an expired (but different) token, as well as the case where the token read from storage is almost expired and results in a 401 when resending the original request.


craigcitro:
if you say the magic "Closes #158" this will do the needful when i merge.

thobrla:
Closes https://github.com/google/oauth2client/pull/158

thobrla:
Will look into the failures tomorrow.

PR #174, opened 2015-05-12
When refreshing credentials with an stream body, rewind the stream be…

thobrla:
…fore re-sending the original request.

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -554,6 +554,11 @@ class OAuth2Credentials(Credentials):
         else:
           headers['user-agent'] = self.user_agent
 
+      body_stream_position = None
+      if all(getattr(body, stream_prop, None) for stream_prop in
+             ('read', 'seek', 'tell')):
+        body_stream_position = body.tell()
+
       resp, content = request_orig(uri, method, body, clean_headers(headers),
                                    redirections, connection_type)
 
@@ -567,6 +572,9 @@ class OAuth2Credentials(Credentials):
                     refresh_attempt + 1, max_refresh_attempts)
         self._refresh(request_orig)
         self.apply(headers)
+        if body_stream_position is not None:
+          body.seek(body_stream_position)
+
         resp, content = request_orig(uri, method, body, clean_headers(headers),
                                      redirections, connection_type)
 
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -32,12 +32,15 @@ import tempfile
 import unittest
 
 from .http_mock import HttpMockSequence
+import six
+
 from oauth2client import file
 from oauth2client import locked_file
 from oauth2client import multistore_file
 from oauth2client import util
 from oauth2client.client import AccessTokenCredentials
 from oauth2client.client import OAuth2Credentials
+from six.moves import http_client
 try:
   # Python2
   from future_builtins import oct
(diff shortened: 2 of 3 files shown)


oauth2client/client.py
```diff
@@ -554,6 +554,11 @@ def new_request(uri, method='GET', body=None, headers=None,
         else:
           headers['user-agent'] = self.user_agent
 
+      body_stream_position = None
+      if all(hasattr(body, stream_prop) for stream_prop in ('read',
```
nathanielmanistaatgoogle: See http://doc.bazaar.canonical.com/developers/code-style.html#hasattr-and-getattr for a brief lesson on why hasattr is fundamentally broken and should never be used.

tests/test_file.py
```diff
@@ -178,6 +180,33 @@ def test_token_refresh_good_store(self):
...
+    s.put(credentials)
+    credentials = s.get()
+    new_cred = copy.copy(credentials)
+    new_cred.access_token = 'bar'
+    s.put(new_cred)
+
+    valid_access_token = '1/3w'
+    token_response = {'access_token': valid_access_token, 'expires_in': 3600}
+    http = HttpMockSequence([
+        ({'status': '401'}, b'Initial token expired'),
```
nathanielmanistaatgoogle: httplib.UNAUTHORIZED rather than the literal value 401 please.

tests/test_file.py
```diff
@@ -178,6 +180,33 @@ def test_token_refresh_good_store(self):
...
+        ({'status': '401'}, b'Store token expired'),
+        ({'status': '200'}, json.dumps(token_response).encode('utf-8')),
```
nathanielmanistaatgoogle: httplib.OK rather than the literal value 200.

tests/test_file.py
```diff
@@ -178,6 +180,33 @@ def test_token_refresh_good_store(self):
...
+        ({'status': '200'}, 'echo_request_body')
+    ])
+
+    body = six.StringIO('streaming body')
+
+    credentials.authorize(http)
+    _, content = http.request('https://example.com', body=body)
+    self.assertEquals(content, 'streaming body')
```
nathanielmanistaatgoogle: Always leave the optional ess off the end of "assertEqual".

nathanielmanistaatgoogle:
Please update the commit log message to conform to http://chris.beams.io/posts/git-commit/#seven-rules.

thobrla:
It looks like the most recent round is failing because system tests at https://github.com/google/oauth2client/blob/master/scripts/run_system_tests.py#L99 expect environment variables set here https://github.com/google/oauth2client/blob/master/scripts/run_system_tests.sh#L21 - perhaps if those environment variables aren't set, that script shouldn't run?

I think the failures are unrelated to my commit.

nathanielmanistaatgoogle:
@dhermes, please take a look at the system test failures in this pull request.

dhermes:
Re-run the build and see if it fails again. (I don't have privileges to re-start on travis-ci.org)

----

No idea why 

```
  if [[ "${TRAVIS_BRANCH}" == "master" ]] && \
         [[ "${TRAVIS_PULL_REQUEST}" == "false" ]]; then
```

would have fallen through. Let's hope it's an intermittent failure. (I've never seen this in 1500+ builds in `gcloud-python`.)

Or maybe it was caused by `"${TRAVIS}" == "true"` being set (also a strange failure)?

tests/test_file.py
craigcitro: actually, we've now switched from `str` to `int`. are the tests OK with an `int`? (should they be?)

craigcitro:
i agree with travis that the travis failures are unrelated to travis's code. (sorry, couldn't resist.)

i tried re-running the tests, to no avail. let's merge this and move the "integration tests are failing" to a new issue?

tests/test_file.py
nathanielmanistaatgoogle: Good point; my comment was erroneous. Bad case of Tuesday afternoon brain. I now think these changes should have been to "str(httplib.symbolic_constant)" rather than to "httplib.symbolic_constant".

Does the test pass with ints instead of strings?

thobrla:
Tests pass with int or strings; regardless, I'll push the string-conversion change.

dhermes:
@craigcitro Can you check which environment variables are set in the Travis settings:
https://travis-ci.org/google/oauth2client/settings/env_vars

I can't see because I am not an admin. This still doesn't explain why

```
  if [[ "${TRAVIS_BRANCH}" == "master" ]] && \
         [[ "${TRAVIS_PULL_REQUEST}" == "false" ]]; then
```

falls through.

craigcitro:
@dhermes those vars seem to be set (at least, there are a handful of relevant-looking names set on that page).

let's maybe throw a handful of debugging info into the script (just print some travis-related env vars?) and start a new issue on that one?

craigcitro:
@thobrla can you also file an issue to follow up on whether tests should use string or int for status codes? (it'd be fantastic if we could just clean it all up to use the symbolic names in `httplib`.)

dhermes:
UPDATE: Our builds have also started failing (after 1500+ builds never failing): https://travis-ci.org/GoogleCloudPlatform/gcloud-python/builds/62434732

It seems the Travis-CI people have changed something (either on purpose or not).

I'm going to debug in `gcloud-python` (where I am an admin) and report back. Do you guys want to create a tracking bug for this?

nathanielmanistaatgoogle:
@dhermes yes please let's move the why-are-tests-failing discussion out of this unrelated pull request thread. :-)

craigcitro:
@nathanielmanistaatgoogle can you merge if you're good with this PR?

nathanielmanistaatgoogle:
Oh right, this whole discussion was actually about making a change... :-P

PR #176, opened 2015-05-13
Adding passenv to tox.ini.

dhermes:
As of tox 2.0, no environment variables are passed through by
default. Some environment variables are needed for running
system tests.

/cc @craigcitro @nathanielmanistaatgoogle This should fix the issue from #174


dhermes:
See https://testrun.org/tox/latest/changelog.html for 2.0 RE: `passenv`

PR #179, opened 2015-05-14
Allow the config dir to be missing for read.

craigcitro:
As part of #162, we added a check that the directory containing the well-known
file exists. In the case that we're about to write the file, this nicely
solves the issue raised in #151. However, in the case that we're entirely
indifferent to the existence of the well-known file, this is an overzealous
check.

This PR simply tweaks the code to only fail in the case that the directory
doesn't exist when writing the well-known file.

PTAL @dhermes @anthmgoogle

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -1260,6 +1260,10 @@ def save_to_well_known_file(credentials, well_known_file=None):
   if well_known_file is None:
     well_known_file = _get_well_known_file()
 
+  config_dir = os.path.dirname(well_known_file)
+  if not os.path.isdir(config_dir):
+    raise OSError('Config directory does not exist: %s' % config_dir)
+
   credentials_data = credentials.serialization_data
   _save_private_file(well_known_file, credentials_data)
 
@@ -1301,9 +1305,6 @@ def _get_well_known_file():
       default_config_dir = os.path.join(os.path.expanduser('~'),
                                         '.config',
                                         _CLOUDSDK_CONFIG_DIRECTORY)
-
-  if not os.path.isdir(default_config_dir):
-    raise OSError('Config directory does not exist', default_config_dir)
 
   return os.path.join(default_config_dir, WELL_KNOWN_CREDENTIALS_FILE)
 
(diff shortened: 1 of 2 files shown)


oauth2client/client.py
```diff
@@ -1303,7 +1307,7 @@ def _get_well_known_file():
                                         _CLOUDSDK_CONFIG_DIRECTORY)
 
   if not os.path.isdir(default_config_dir):
-    raise OSError('Config directory does not exist', default_config_dir)
+    logging.info('Config directory does not exist: %s', default_config_dir)
```
dhermes: This library is already obnoxious with logging (or is that just `google-api-python-client`?). Why add to it?

tests/test_oauth2client.py
```diff
@@ -305,6 +297,18 @@ def test_save_to_well_known_file_service_account(self):
     self.assertEqual('ABCDEF', d['private_key_id'])
     os.remove(temp_credential_file)
 
+  def test_save_well_known_file_with_non_existent_config_dir(self):
+    credential_file = datafile(
+        os.path.join('gcloud', 'application_default_credentials.json'))
+    credentials = _get_application_default_credential_from_file(
+        credential_file)
```
dhermes, line 304: Why does this matter? Why not just use

``` python
credentials = object()
```

dhermes:
Overall looks good except for tiny comments. Easy to review.

oauth2client/client.py
craigcitro: `info` shouldn't be on by default, but you're right that it probably doesn't serve much purpose. dropping for now.

tests/test_oauth2client.py
craigcitro, line 304: well, in principle `save_to_well_known_file` could do things like validate the credential itself before looking for the file location. given that it's easy to pass a valid credential, might as well, right?

craigcitro:
@anthmgoogle unless you have any thoughts/objections, i'll probably submit tomorrow and roll a new release.

anthmgoogle:
LGTM. Thanks.

PR #182, opened 2015-05-15
Simplify OpenSSL checks in crypt.py.

craigcitro:
Previously, when confirming OpenSSL is installed, we checked that the module
was installed, and verified that `crypto.py` was contained in the package
directory.

However, this can cause issues in cases of non-standard installs. In addition,
the `imp.find_module` call already raises an `ImportError` in the case that
the module is missing, so this second check isn't needed.

Instead, we just drop the explicit check for `crypto.py` in the `OpenSSL`
directory.

PTAL @dhermes

### DIFF
--- a/oauth2client/crypt.py
+++ b/oauth2client/crypt.py
@@ -39,9 +39,7 @@ class AppIdentityError(Exception):
 
 
 try:
-  _, _package_dir, _ = imp.find_module('OpenSSL')
-  if not os.path.isfile(os.path.join(_package_dir, 'crypto.py')):
-    raise ImportError('No module named OpenSSL')
+  _ = imp.find_module('OpenSSL')
 
   class OpenSSLVerifier(object):
     """Verifies the signature on a message."""
--- a/tests/test_crypt.py
+++ b/tests/test_crypt.py
@@ -58,15 +58,17 @@ class Test_pkcs12_key_as_pem(unittest.TestCase):
     self.assertTrue(pem_contents in [pkcs12_key_as_pem, alternate_pem])
 
   def test_without_openssl(self):
-    import os
-    path_isfile = os.path.isfile
+    import imp
+    imp_find_module = imp.find_module
+    def find_module(module_name):
+      raise ImportError('No module named %s' % module_name)
     try:
-      os.path.isfile = lambda value: False
+      imp.find_module = find_module
       reload(crypt)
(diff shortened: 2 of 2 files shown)


dhermes:
@craigcitro I looked at #159 and didn't find any discussion, but I'm fairly certain that `imp.find_module` **does not** throw an `ImportError` in all cases, so your proposed fix doesn't work. I'm going to dig around and report back.

dhermes:
I'm fairly certain I used http://stackoverflow.com/q/14050281/1068170 as my reference and also recall using `pkgutil` before deciding on `imp`.

It seems the most reasonable explanation for

```python
if not os.path.isfile(os.path.join(_package_dir, 'crypto.py')):
```

is to verify that the removed line

```python
from OpenSSL import crypto
```

still works. Testing out in `0.13` instead of `0.14` it seems that `OpenSSL.crypto` comes via `crypto.so` instead of `crypto.py`:

```bash
~/Desktop $ virtualenv venv
~/Desktop $ source venv/bin/activate
(venv) ~/Desktop $ pip install pyOpenSSL==0.13
Collecting pyOpenSSL==0.13
...
(venv) ~/Desktop $ ls -1 venv/local/lib/python2.7/site-packages/OpenSSL/
crypto.so*
__init__.py
__init__.pyc
rand.so*
SSL.so*
test/
tsafe.py
tsafe.pyc
version.py
version.pyc
(venv) ~/Desktop $ cat venv/local/lib/python2.7/site-packages/OpenSSL/version.py
# Copyright (C) AB Strakt
# Copyright (C) Jean-Paul Calderone
# See LICENSE for details.

"""
pyOpenSSL - A simple wrapper around the OpenSSL library
"""

__version__ = '0.13'
```

dhermes:
@craigcitro WDYT of all that?

craigcitro:
@dhermes I guess I'm confused - the python docs explicitly say `find_module` will throw if the module isn't found, and that SO link only mentions a problem for dotted modules.

where will this fail?

dhermes:
Yeah you can disregard my first comment. It's safe(ish, mostly) to say that it will throw when it doesn't exist.

More WDYT of the check of `OpenSSL.crypto` and the original intent of the code.

craigcitro:
do we think there's a case where someone could have `OpenSSL` installed by **not** have an `OpenSSL.crypto` module?

dhermes:
No I don't think so. The flipside: we could check for `crypto.py` **OR** `crypto.so`.

PR #185, opened 2015-05-19
Fixing environment detection behavior on GAE Managed VMs.

theacodes:
When running on Managed VMs, the evironment variables are set for "SERVER_SOFTWARE" but the user doesn't necessarily have to be using a appengine-compat runtime such as `google/appengine-python`, they could be using a general purpose one such as `google/python-runtime`. Because the generic one doesn't have the app engine sdk, it would crash when trying to obtain default credentials. This fixes the problem by checking if the app engine sdk is importable.

### DIFF
--- a/tests/test_oauth2client.py
+++ b/tests/test_oauth2client.py
@@ -23,9 +23,11 @@ Unit tests for oauth2client.
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 
 import base64
+import contextlib
 import datetime
 import json
 import os
+import sys
 import time
 import unittest
 
@@ -144,6 +146,22 @@ class MockResponse(object):
     return Info(self._headers)
 
 
+@contextlib.contextmanager
+def mock_module_import(module):
+  """Place a dummy objects in sys.modules to mock an import test."""
+  parts = module.split('.')
+  entries = ['.'.join(parts[:i+1]) for i in range(len(parts))]
+  for entry in entries:
+    sys.modules[entry] = object()
+  
+  try:
+    yield
+
+  finally:
+    for entry in entries:
+      del sys.modules[entry]
+
+
 class GoogleCredentialsTests(unittest.TestCase):
 
   def setUp(self):
@@ -200,12 +218,14 @@ class GoogleCredentialsTests(unittest.TestCase):
                      credentials.create_scoped(['dummy_scope']))
 
   def test_get_environment_gae_production(self):
-    os.environ['SERVER_SOFTWARE'] = 'Google App Engine/XYZ'
-    self.assertEqual('GAE_PRODUCTION', _get_environment())
+    with mock_module_import('google.appengine'):
+      os.environ['SERVER_SOFTWARE'] = 'Google App Engine/XYZ'
+      self.assertEqual('GAE_PRODUCTION', _get_environment())
 
   def test_get_environment_gae_local(self):
-    os.environ['SERVER_SOFTWARE'] = 'Development/XYZ'
-    self.assertEqual('GAE_LOCAL', _get_environment())
+    with mock_module_import('google.appengine'):
+      os.environ['SERVER_SOFTWARE'] = 'Development/XYZ'
+      self.assertEqual('GAE_LOCAL', _get_environment())
 
   def test_get_environment_gce_production(self):
     os.environ['SERVER_SOFTWARE'] = ''
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -992,11 +992,18 @@ def _get_environment(urlopen=None):
   # None is an unset value, not the default.
   SETTINGS.env_name = DEFAULT_ENV_NAME
 
-  server_software = os.environ.get('SERVER_SOFTWARE', '')
-  if server_software.startswith('Google App Engine/'):
-    SETTINGS.env_name = 'GAE_PRODUCTION'
-  elif server_software.startswith('Development/'):
-    SETTINGS.env_name = 'GAE_LOCAL'
+  try:
+    import google.appengine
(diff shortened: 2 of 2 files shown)


tests/test_oauth2client.py
```diff
@@ -200,12 +201,20 @@ def test_create_scoped(self):
                      credentials.create_scoped(['dummy_scope']))
 
   def test_get_environment_gae_production(self):
+    sys.modules['google'] = object()
+    sys.modules['google.appengine'] = object()
     os.environ['SERVER_SOFTWARE'] = 'Google App Engine/XYZ'
     self.assertEqual('GAE_PRODUCTION', _get_environment())
```
nathanielmanistaatgoogle: If this assertion fails, won't the sys.modules space be polluted by this test and possibly interfere with other tests?

tests/test_oauth2client.py
```diff
@@ -200,12 +201,20 @@ def test_create_scoped(self):
...
+    del sys.modules['google']
+    del sys.modules['google.appengine']
 
   def test_get_environment_gae_local(self):
+    sys.modules['google'] = object()
+    sys.modules['google.appengine'] = object()
     os.environ['SERVER_SOFTWARE'] = 'Development/XYZ'
     self.assertEqual('GAE_LOCAL', _get_environment())
```
nathanielmanistaatgoogle: Same question here.

tests/test_oauth2client.py
theacodes: Possibly, and it could lead to making it difficult to determine which test is causing this. I can re-work this as a context manager if desired.

nathanielmanistaatgoogle: It is strongly desired.

jonparrott:
@nathanielmanistaatgoogle context manager added.

tests/test_oauth2client.py
```diff
@@ -144,6 +146,23 @@ def get(self, key, default=None):
...
 
+@contextlib.contextmanager
+def mock_module_import(module):
+  """Place a dummy object in sys.modules to mock an import test"""
+  parts = module.split('.')
+  for n in range(len(parts)):
+    path = '.'.join(parts[:n+1])
+    sys.modules[path] = object()
+  
+  try:
```
nathanielmanistaatgoogle, line 157: I don't understand why yield is wrapped in a try/finally?

tests/test_oauth2client.py
```diff
@@ -144,6 +146,23 @@ def get(self, key, default=None):
...
+    yield
+
+  finally:
+    for n in range(len(parts)):
```
nathanielmanistaatgoogle: In the first half of the function create a sequence of the dummy entries added so that you don't have to recompute them to remove them here in the second half.

tests/test_oauth2client.py
```diff
@@ -144,6 +146,23 @@ def get(self, key, default=None):
     return Info(self._headers)
 
 
+@contextlib.contextmanager
+def mock_module_import(module):
+  """Place a dummy object in sys.modules to mock an import test"""
+  parts = module.split('.')
+  for n in range(len(parts)):
```
nathanielmanistaatgoogle: If I say that I prefer "i" for an index and "n" to represent a quantity, am I being too fussy?

nathanielmanistaatgoogle:
Please squash commits with each round of code review back-and-forth; the successive stages of progress here shouldn't necessarily become a part of canonical project history.

tests/test_oauth2client.py
theacodes, line 157: To prevent exceptions from causing leaks into sys.modules. See [here](https://docs.python.org/2/library/contextlib.html#contextlib.closing)

tests/test_oauth2client.py
theacodes: I'll accept it. :smile:

jonparrott:
Commit squashed and the context manager was updated with your feedback.

tests/test_oauth2client.py
```diff
@@ -144,6 +146,22 @@ def get(self, key, default=None):
     return Info(self._headers)
 
 
+@contextlib.contextmanager
+def mock_module_import(module):
+  """Place a dummy object in sys.modules to mock an import test"""
+  parts = module.split('.')
+  entries = ['.'.join(parts[:i+1]) for i in range(len(parts))]
+  for n in entries:
```
nathanielmanistaatgoogle: Please don't use "n" for values that aren't integers.

I think "entry" or "fully_qualified_module_name" are both fine names for this value.

tests/test_oauth2client.py
```diff
@@ -144,6 +146,22 @@ def get(self, key, default=None):
     return Info(self._headers)
 
 
+@contextlib.contextmanager
+def mock_module_import(module):
+  """Place a dummy object in sys.modules to mock an import test"""
```
nathanielmanistaatgoogle: Nitpick: missing period at end of text.

tests/test_oauth2client.py
nathanielmanistaatgoogle, line 157: Wow, today I learned that until now I've been writing dangerously defective code with contextlib.contextmanager. :-( Thanks for the education!

theacodes, line 157: Haha!
marvin

jonparrott:
Got the nits taken care of.

nathanielmanistaatgoogle:
Looks good to me. @craigcitro or @soltanmm, please be a second reviewer on this?

anthmgoogle:
LGTM.

nathanielmanistaatgoogle:
Thank you @anthmgoogle.

jonparrott:
Thanks, everyone!

PR #186, opened 2015-05-19
Fall back to importing for OpenSSL detection.

craigcitro:
Previously, we'd attempted to detect OpenSSL without actually importing, in
order to avoid a runtime penalty for importing `oauth2client`. However, in
some exotic situations, `imp.find_module` can fail even though the import
would be successful.

We tweak to fall back to importing, and add some gross to the corresponding
test.

PTAL @dhermes

### DIFF
--- a/oauth2client/crypt.py
+++ b/oauth2client/crypt.py
@@ -38,8 +38,33 @@ class AppIdentityError(Exception):
   pass
 
 
+def _TryOpenSslImport():
+  """Import OpenSSL, avoiding the explicit import where possible.
+
+  Importing OpenSSL 0.14 can take up to 0.5s, which is a large price
+  to pay at module import time. However, it's also possible for
+  ``imp.find_module`` to fail to find the module, even when it's
+  installed. (This is the case in various exotic environments,
+  including some relevant for Google.) So we first try a fast-path,
+  and fall back to the slow import as needed.
+
+  Args:
+    None
+  Returns:
+    None
+  Raises:
+    ImportError if OpenSSL is unavailable.
+
+  """
+  try:
+    _ = imp.find_module('OpenSSL')
+    return
+  except ImportError:
+    import OpenSSL
+
+
 try:
-  _ = imp.find_module('OpenSSL')
+  _TryOpenSslImport()
 
   class OpenSSLVerifier(object):
     """Verifies the signature on a message."""
--- a/tests/test_crypt.py
+++ b/tests/test_crypt.py
@@ -60,15 +60,22 @@ class Test_pkcs12_key_as_pem(unittest.TestCase):
   def test_without_openssl(self):
     import imp
     imp_find_module = imp.find_module
+    orig_sys_path = sys.path
     def find_module(module_name):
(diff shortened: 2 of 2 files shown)


oauth2client/crypt.py
```diff
@@ -38,8 +38,32 @@ class AppIdentityError(Exception):
...
+  Args:
+    None
+  Returns:
+    None
+  Raises:
+    ImportError if OpenSSL is unavailable.
+  """
+  try:
+    _ = imp.find_module('OpenSSL')
+    return
```
dhermes, line 61: You don't need the `_ =` bit. Sorry I didn't mention this on previous PR.

craigcitro, line 61: this quiets the linter (explicitly telling it we're ignoring the return value)

dhermes, line 61: SGTM. :(

I hate being bossed around by linters.

oauth2client/crypt.py
```diff
@@ -38,8 +38,32 @@ class AppIdentityError(Exception):
...
+  except ImportError:
+    pass
+  import OpenSSL
```
dhermes: Just do this instead:

``` python
  try:
    _ = imp.find_module('OpenSSL')
  except ImportError:
    import OpenSSL
```

oauth2client/crypt.py
```diff
@@ -38,8 +38,32 @@ class AppIdentityError(Exception):
...
 
 
+def _TryOpenSslImport():
+  """Import OpenSSL, avoiding the explicit import where possible.
+
+  Importing OpenSSL can take up to 0.5s, which is a large price to pay
+  at module import time. However, it's also possible for
+  ``imp.find_module`` to fail to find the module, even when it's
+  installed. So we first try a fast-path, and fall back to the slow
+  import as needed.
```
dhermes: This should be much more specific.

```
Importing OpenSSL 0.14 typically takes up ~0.5s, which is unacceptable
at module import time. However, within a ``.par`` (packaed archive) file,
``imp.find_module`` does not work correctly, even when `OpenSSL` is in
the package archive. In this case, we first try to detect without loading,
and fall back to the slow import as needed.
```

tests/test_crypt.py
```diff
@@ -60,15 +60,22 @@ def test_succeeds(self):
   def test_without_openssl(self):
     import imp
     imp_find_module = imp.find_module
+    orig_sys_path = sys.path
     def find_module(module_name):
       raise ImportError('No module named %s' % module_name)
     try:
+      for m in list(sys.modules):
```
dhermes, line 67: You need to copy `sys.modules` before this and then restore it.

craigcitro, line 67: `sys.modules` doesn't like being replaced (or, at least, it would require a bit more fanciness). i'm leaving as-is, since it's "just a test". `;)`

oauth2client/crypt.py
craigcitro: done.

craigcitro:
thanks!

PR #191, opened 2015-05-23
Check for OpenSSL.crypto when detecting OpenSSL.

### LINKED ISSUE #190: OpenSSL detection fails if OpenSSL is available, but OpenSSL.crypto is missing.
in the case that someone has installed `OpenSSL` without including the `crypto` module, our `OpenSSL` detection now detects `OpenSSL`, but fails when trying to use it. this wouldn't happen in the case of using `pip`, but could be possible if someone's customized their install (say to remove C extensions in a case like Google App Engine).

@dhermes will surely get a chuckle out of this one. :smile:

craigcitro:
Previously, we assumed that anyone using `OpenSSL` would have installed it via
a mechanism like `pip`; this has come back to bite us.

We modify our code to look for `OpenSSL.crypto`, not just `OpenSSL`, and add
another (unpleasant) test.

Fixes #190.

PTAL @dhermes

### DIFF
--- a/oauth2client/crypt.py
+++ b/oauth2client/crypt.py
@@ -57,10 +57,14 @@ def _TryOpenSslImport():
 
   """
   try:
-    _ = imp.find_module('OpenSSL')
+    _, _package_dir, _ = imp.find_module('OpenSSL')
+    if not (os.path.isfile(os.path.join(_package_dir, 'crypto.py')) or
+            os.path.isfile(os.path.join(_package_dir, 'crypto.so')) or
+            os.path.isdir(os.path.join(_package_dir, 'crypto'))):
+      raise ImportError('No module named OpenSSL.crypto')
     return
   except ImportError:
-    import OpenSSL
+    import OpenSSL.crypto
 
 
 try:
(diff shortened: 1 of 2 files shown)


oauth2client/crypt.py
```diff
@@ -57,10 +57,13 @@ def _TryOpenSslImport():
 
   """
   try:
-    _ = imp.find_module('OpenSSL')
+    _, _package_dir, _ = imp.find_module('OpenSSL')
+    if not (os.path.isfile(os.path.join(_package_dir, 'crypto.py')) or
+            os.path.isfile(os.path.join(_package_dir, 'crypto.so'))):
+      raise ImportError('No module named OpenSSL.crypto')
```
dhermes, line 64: Didn't you say it was possible that there could be a folder called `crypto` as well?

craigcitro, line 64: i was letting those folks fall through to explicit import -- but you're right, it's easy to catch here.

craigcitro:
PTAL

dhermes:
LGTM.

PR #196, opened 2015-06-17
Remove @util.positional from the authorize() wrapper.

eamonnmcmanus:
Since the original method does not have a constraint on positional parameters,
its wrapper should not either. See
http://stackoverflow.com/questions/18126157/appengine-warning-during-python-app-update/30469210
for a case where this is producing annoying warnings.

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -532,7 +532,6 @@ class OAuth2Credentials(Credentials):
     request_orig = http.request
 
     # The closure that will replace 'httplib2.Http.request'.
-    @util.positional(1)
     def new_request(uri, method='GET', body=None, headers=None,
                     redirections=httplib2.DEFAULT_MAX_REDIRECTS,
                     connection_type=None):
--- a/tests/test_oauth2client.py
+++ b/tests/test_oauth2client.py
@@ -40,6 +40,7 @@ from .http_mock import HttpMockSequence
 from oauth2client import GOOGLE_REVOKE_URI
 from oauth2client import GOOGLE_TOKEN_URI
 from oauth2client import client
+from oauth2client import util as oauth2client_util
 from oauth2client.client import AccessTokenCredentials
 from oauth2client.client import AccessTokenCredentialsError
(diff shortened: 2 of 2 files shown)


dhermes:
Your change LGTM. The coverage decrease is just an unavoidable math fact caused by removing 1 line:

```
(a - 1) / (b - 1) < a / b <==> (a - 1) b < a (b - 1) <==> a < b
```

----

Regarding your comment

```
	
OK, I think github.com/google/oauth2client/pull/196 should fix this problem, assuming it is accepted and then propagated to the SDK.
```

I'm not sure how quickly a new version will be propagated to the SDK.

dhermes:
@nathanielmanistaatgoogle You are a go to merge, note that the method being monkey patched (ewwww I wish `httplib2` didn't monkey patch an instance method on a live instance) [does not have][1] the util positional restriction.

PS I now have merge privileges but just wanted to make sure someone else saw.

[1]: https://github.com/jcgregorio/httplib2/blob/f9f797223d075874b0031aea832152688ec23fef/python2/httplib2/__init__.py

nathanielmanistaatgoogle:
Thanks for the contribution!

(1) Please rewrite your commit message to conform to chris.beams.io/posts/git-commit/#seven-rules.

(2) You describe this change as a fix, so how about a regression test?

(3) How much digging did you do on this? In what commit was the original error made? What line of thinking or false assumption is the root cause of the mistake? Is there some larger lesson we should learn or action we should take beyond just deleting a single method decoration?

eamonnmcmanus:
(1) Updated the commit message.

(2) Added a regression test.

(3) To be honest, I think the cause of the mistake is using the `@util.positional` annotation at all. I don't see what advantage it confers, and it has certainly cost me and others a lot of time.

dhermes:
FYI that regression test has side-effects.

nathanielmanistaatgoogle:
That it does; I should have caught that.

eamonnmcmanus:
Yes sorry, I meant to add teardown logic. I'll send another PR with that.

PR #203, opened 2015-06-24
Casting message to bytes in PyCrypto verifier.

### LINKED ISSUE #201: Python 3 + PyCrypto: crypt.verify_signed_jwt_with_certs raises "Invalid token signature" due to string/bytes bug
I was getting a mysterious "Invalid token signature" error trying to validate Google Sign-In-generated tokens using this method on Python 3.4.0 with PyCrypto. After confirming with jwt.io and the `tokeninfo` endpoint that the token was, indeed, valid, I dove into the method source. [After realizing that all exceptions were swallowed in the PyCrypto wrapper's `verify` method](https://github.com/google/oauth2client/blob/c5eb9f14d46d89fb33c4409a0d32fad4b919e136/oauth2client/crypt.py#L231-L235), I found out that the call:

```python
SHA256.new(message)
```

was raising:

```python
TypeError: Unicode-objects must be encoded before hashing
```

This is because `message` is a string, and not bytes. This isn't surprising when you see where `message` is initially set: 
https://github.com/google/oauth2client/blob/c5eb9f14d46d89fb33c4409a0d32fad4b919e136/oauth2client/crypt.py#L413.

I applied a quick fix to that line:

```python
signed = ('%s.%s' % (segments[0], segments[1])).encode('utf-8')
```

And the method validated my token correctly.

In addition to this fix, I'd recommend removing that `try/except` block - I'm not sure why you'd want to swallow exceptions there, other than to make it harder for poor developers like me to debug issues ;)

dhermes:
Also
- Removing bare `except:` statements in both the PyCrypto and
  OpenSSL verifiers (in the `verify` method).
- Catching the only exception possible in the OpenSSL verifier
  (it is `OpenSSL.crypto.Error`).
- Converting the signature to bytes (if not already) in the
  OpenSSL verifier.
- Adding a test with both a unicode and bytes signature for
  each verifier.

Fixes #201.

/cc @nathanielmanistaatgoogle

### DIFF
--- a/oauth2client/crypt.py
+++ b/oauth2client/crypt.py
@@ -85,20 +85,24 @@ try:
       """Verifies a message against a signature.
 
       Args:
-        message: string, The message to verify.
-        signature: string, The signature on the message.
+        message: string or bytes, The message to verify. If string, will be
+                 encoded to bytes as utf-8.
+        signature: string or bytes, The signature on the message. If string,
+                   will be encoded to bytes as utf-8.
 
       Returns:
         True if message was signed by the private key associated with the public
         key that this object was constructed with.
       """
       from OpenSSL import crypto
+      if isinstance(message, six.text_type):
+        message = message.encode('utf-8')
+      if isinstance(signature, six.text_type):
+        signature = signature.encode('utf-8')
       try:
-        if isinstance(message, six.text_type):
-          message = message.encode('utf-8')
         crypto.verify(self._pubkey, signature, message, 'sha256')
         return True
-      except:
+      except crypto.Error:
         return False
 
     @staticmethod
@@ -221,18 +225,18 @@ try:
       """Verifies a message against a signature.
 
       Args:
-        message: string, The message to verify.
-        signature: string, The signature on the message.
+        message: string or bytes, The message to verify. If string, will be
+                 encoded to bytes as utf-8.
+        signature: string or bytes, The signature on the message.
 
       Returns:
         True if message was signed by the private key associated with the public
         key that this object was constructed with.
       """
-      try:
-        return PKCS1_v1_5.new(self._pubkey).verify(
-            SHA256.new(message), signature)
-      except:
-        return False
+      if isinstance(message, six.text_type):
+        message = message.encode('utf-8')
+      return PKCS1_v1_5.new(self._pubkey).verify(
+          SHA256.new(message), signature)
 
     @staticmethod
     def from_string(key_pem, is_x509_cert):
--- a/tests/test_jwt.py
+++ b/tests/test_jwt.py
@@ -59,7 +59,7 @@ class CryptTests(unittest.TestCase):
     self._check_sign_and_verify('privatekey.%s' % self.format)
 
   def test_sign_and_verify_from_converted_pkcs12(self):
-    """Tests that following instructions to convert from PKCS12 to PEM works."""
+    # Tests that following instructions to convert from PKCS12 to PEM works.
     if self.format == 'pem':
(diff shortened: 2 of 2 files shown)


dhermes:
@nathanielmanistaatgoogle @craigcitro As mentioned above, I removed the bare `except:` statements the verifiers. It's unclear if they were there intentionally or just done since the exception types were neglected.

tests/test_jwt.py
```diff
@@ -59,7 +59,7 @@ def test_sign_and_verify(self):
     self._check_sign_and_verify('privatekey.%s' % self.format)
 
   def test_sign_and_verify_from_converted_pkcs12(self):
-    """Tests that following instructions to convert from PKCS12 to PEM works."""
```
nathanielmanistaatgoogle, line 62: What's up with the change on this line?

tests/test_jwt.py
```diff
@@ -196,6 +198,7 @@ def setUp(self):
 
 
 class PEMCryptTestsOpenSSL(CryptTests):
+
```
nathanielmanistaatgoogle, line 201: Any comment on these blank line insertions?

oauth2client/crypt.py
```diff
@@ -93,12 +93,14 @@ def verify(self, message, signature):
         key that this object was constructed with.
       """
       from OpenSSL import crypto
+      if isinstance(message, six.text_type):
```
nathanielmanistaatgoogle, line 98: Any need to update the doc string for these arguments from "string"?

tests/test_jwt.py
dhermes, line 62: Docstrings on tests break the output of `nosetest` (and are generally frowned upon).

I can leave it out of this commit, I just noticed it because it showed up in a Py3.4 failure I fixed before adding

``` python
      if isinstance(signature, six.text_type):
        signature = signature.encode('utf-8')
```

to `OpenSSLVerifier.verify`.

nathanielmanistaatgoogle, line 62: Wait, I've without knowing been doing something wrong for years _again_? Can you link me to a source for this general frowning?

tests/test_jwt.py
dhermes, line 201: Just consistency with `CryptTests`. I thought the extra line was a requirement for the linter, but after some experimenting it doesn't seem to be.

oauth2client/crypt.py
dhermes, line 98: Yeah happy to do it.

tests/test_jwt.py
dhermes, line 62: Couldn't find anything in a few minutes of searching. I'm pretty sure I've seen something before.

Closest I got was the `unittest` docs for making the description ignore the docstring:
https://docs.python.org/2/library/unittest.html#unittest.TestCase.shortDescription

Maybe @tseaver has a quick reference

tseaver, line 62: Docstrings in tnittest-based testcase methods are a bad idea because (historically, anyway) testrunners tend to display the docstring, rather than the name of the testcase, in output, making it hard to diagnose failures.

craigcitro:
i don't know that there's a reason for the bare `except:` clauses -- but if there is, we'll probably get a bug report soon enough. :wink:

dhermes:
@nathanielmanistaatgoogle I have updated the docstrings on the updated `verify()` methods.

Sorry I let this thing fall away.

Are we good to go here?

nathanielmanistaatgoogle:
LGTM; I think we're good to go here.

dhermes:
Thanks!

PR #206, opened 2015-06-25
Supports client authentication using Authorization header 

bjmc:
Currently, this client library authenticates by passing `client_id` and `client_secret` in the POST body when exchanging an auth code for an access token.

According to [RFC 6749 section 2.3.1](https://tools.ietf.org/html/rfc6749#section-2.3.1):

>  The authorization server MUST support the HTTP Basic authentication scheme for authenticating clients that were issued a client password.

This changeset adds an optional parameter `authorization_header` to allow use with OAuth2 providers that do not support passing client credentials in the request body.

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -1828,6 +1828,8 @@ class OAuth2WebServerFlow(Flow):
       **kwargs: dict, The keyword arguments are all optional and required
                         parameters for the OAuth calls.
     """
+    # scope is a required argument, but to preserve backwards-compatibility
+    # we don't want to rearrange the positional arguments
     if scope is None:
       raise TypeError("The value of scope must not be None")
     self.client_id = client_id
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -1784,7 +1784,9 @@ class OAuth2WebServerFlow(Flow):
   """
 
   @util.positional(4)
-  def __init__(self, client_id, client_secret, scope,
+  def __init__(self, client_id,
+               client_secret=None,
+               scope=None,
                redirect_uri=None,
                user_agent=None,
                auth_uri=GOOGLE_AUTH_URI,
@@ -1792,6 +1794,7 @@ class OAuth2WebServerFlow(Flow):
                revoke_uri=GOOGLE_REVOKE_URI,
                login_hint=None,
                device_uri=GOOGLE_DEVICE_URI,
+               authorization_header=None,
                **kwargs):
     """Constructor for OAuth2WebServerFlow.
 
@@ -1819,9 +1822,14 @@ class OAuth2WebServerFlow(Flow):
         proper multi-login session, thereby simplifying the login flow.
       device_uri: string, URI for device authorization endpoint. For convenience
         defaults to Google's endpoints but any OAuth 2.0 provider can be used.
+      authorization_header: string, For use with OAuth 2.0 providers that
+        require a client to authenticate using a header value instead of passing
+        client_secret in the POST body.
       **kwargs: dict, The keyword arguments are all optional and required
                         parameters for the OAuth calls.
     """
+    if scope is None:
+      raise TypeError("The value of scope must not be None")
     self.client_id = client_id
     self.client_secret = client_secret
     self.scope = util.scopes_to_string(scope)
@@ -1832,6 +1840,7 @@ class OAuth2WebServerFlow(Flow):
     self.token_uri = token_uri
     self.revoke_uri = revoke_uri
     self.device_uri = device_uri
+    self.authorization_header = authorization_header
     self.params = {
         'access_type': 'offline',
         'response_type': 'code',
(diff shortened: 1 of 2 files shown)


nathanielmanistaatgoogle:
Please reform your commit to conform to http://chris.beams.io/posts/git-commit/#seven-rules.

@anthmgoogle, may I ask you to lead review of this?

oauth2client/client.py
```diff
@@ -1784,7 +1784,9 @@ class OAuth2WebServerFlow(Flow):
   """
 
   @util.positional(4)
-  def __init__(self, client_id, client_secret, scope,
+  def __init__(self, client_id,
+               client_secret=None,
+               scope=None,
```
dhermes, line 1789: You should change `util.positional(4)` if you are no longer having 4 positional arguments.

oauth2client/client.py
```diff
@@ -1819,9 +1822,14 @@ def __init__(self, client_id, client_secret, scope,
...
       device_uri: string, URI for device authorization endpoint. For convenience
         defaults to Google's endpoints but any OAuth 2.0 provider can be used.
+      authorization_header: string, For use with OAuth 2.0 providers that
+        require a client to authenticate using a header value instead of passing
+        client_secret in the POST body.
       **kwargs: dict, The keyword arguments are all optional and required
                         parameters for the OAuth calls.
     """
+    if scope is None:
+      raise TypeError("The value of scope must not be None")
```
dhermes, line 1834: Then make it a positional argument. Don't re-invent the wheel here.

Or is the issue that you can't change argument order for backwards compatibility?

oauth2client/client.py
bjmc, line 1789: Maybe I'm misunderstanding, but it seems like this decorator is supposed to declare the _maximum_ number of positional arguments. From the docstring:

> A decorator to declare that only the first N arguments my be positional.

I deliberately want to make sure that this changeset doesn't break any existing users of this library, so I think we still want to support passing up to 4 positional arguments.

oauth2client/client.py
bjmc, line 1834: > Or is the issue that you can't change argument order for backwards compatibility?

Yes, that's exactly the issue. I'm assuming this library is popular and widely used, and I do not want to introduce a breaking change to the method signature.

oauth2client/client.py
dhermes, line 1789: I think that sounds about right

oauth2client/client.py
dhermes, line 1834: Probably worthwhile to put in a comment about it then. Because a missing positional arg would throw a `TypeError`, so it needs to be explained why this is necessary.

bjmc:
@nathanielmanistaatgoogle @anthmgoogle Edited commit message.

oauth2client/client.py
bjmc, line 1834: I added a comment and a test to double-check this exception gets thrown.

tests/test_oauth2client.py
```diff
@@ -945,6 +945,9 @@ def test_override_flow_via_kwargs(self):
     self.assertEqual(OOB_CALLBACK_URN, q['redirect_uri'][0])
     self.assertEqual('online', q['access_type'][0])
 
+  def test_scope_is_required(self):
+    self.assertRaises(TypeError, OAuth2WebServerFlow, 'client_id+1')
```
dhermes, line 949: :+1:

tests/test_oauth2client.py
```diff
@@ -1027,6 +1030,29 @@ def __contains__(self, name):
...
+  def test_exchange_using_authorization_header(self):
+    auth_header = 'Basic Y2xpZW50X2lkKzE6c2VjcmV0KzE=',
+    flow = OAuth2WebServerFlow(
+      client_id='client_id+1',
+      authorization_header=auth_header,
+      scope='foo',
+      redirect_uri=OOB_CALLBACK_URN,
+      user_agent='unittest-sample/1.0',
+      revoke_uri='dummy_revoke_uri',
+        )
```
dhermes, line 1042: What is this lined up with?

dhermes:
It's unclear why coverage is decreasing. You seem to be hitting all new code-paths added.

bjmc:
Yeah, I didn't understand why the coverage decreased, either.

tests/test_oauth2client.py
bjmc, line 1042: I just copy-pasted from one of the other tests. Are you saying you want that close parenthesis moved?

dhermes, line 1042: Don't really care, just noticed it looked weird.

This library will someday have an automatic linter. Le sigh :fr:

dhermes:
@nathanielmanistaatgoogle I'd say this is ready to go if you like it. I don't know much about the actual feature being implemented though.

nathanielmanistaatgoogle:
@anthmgoogle are you going to be able to look at this?

anthmgoogle:
LGTM.

nathanielmanistaatgoogle:
Thanks everyone; merging.

bjmc:
Thanks!

PR #210, opened 2015-07-01
Make test_oauth2client not executable so it runs with nose.

pcostell:
Also replaces assertNotIn with py2.6 compatible call.


pcostell:
@dhermes I moved this over to a new pull request just to keep all the changes out of my feature request.

dhermes:
LGTM

PR #214, opened 2015-07-02
Automating release via Travis.

### LINKED ISSUE #213: Allow Travis to auto-release
See #208 for a discussion about release.

See https://github.com/GoogleCloudPlatform/gcloud-python/blob/bb25b9e924c12c648c3b5ac50d5bdc18409b2a5a/.travis.yml#L24-L33 for an example and http://docs.travis-ci.com/user/deployment/pypi/ for docs.

dhermes:
Fixes #213.

@craigcitro or @nathanielmanistaatgoogle we need a real PyPI account for pushing to this repo from Travis or we could just use one of yours and just encrypt it with the Travis command line tool:

```
travis encrypt PW_GOES_HERE
```

/cc @pcostell @eamonnmcmanus

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,3 +19,12 @@ after_success:
 - if [[ "${TOX_ENV}" == "py27" ]]; then tox -e coveralls; fi
 notifications:
   email: false
+
+deploy:
+  provider: pypi
+  user: gcloudpypi
+  password:
+    secure: "C9ImNa5kbdnrQNfX9ww4PUtQIr3tN+nfxl7eDkP1B8Qr0QNYjrjov7x+DLImkKvmoJd3dxYtYIpLE9esObUHu0gKHYxqymNHtuAAyoBOUfPtmp0vIEse9brNKMtaey5Ngk7ZWz9EHKBBqRHxqgN+Giby+K9Ta3K3urJIq6urYhE="
+  on:
+    tags: true
+    repo: google/oauth2client


nathanielmanistaatgoogle:
This needs to be reformed following the merge of #216, right?

dhermes:
Yup. Version doesn't need to be cut anymore (two people at Google needed a release).

Now it's just the discussion, do we want to use an existing PyPI account and encrypt to password (a la https://github.com/GoogleCloudPlatform/gcloud-python/blob/583d1cef41b7266994371a6fee546e9f079c382e/.travis.yml#L24-L33) or make a new account or disallow Travis from making releases.

dhermes:
Shall we discuss? Do we want to 

- use an existing PyPI account and encrypt the password
- make a new PyPI account to do the same
- disallow Travis from making releases.

dhermes:
@nathanielmanistaatgoogle Ping, can we discuss the three options listed above?

dhermes:

nathanielmanistaatgoogle:
Change content LGTM - how do we test this? Can we push to a branch like "1.4.11.test.release.do.not.use" and see that that appears on PyPI or something like that?

dhermes:
We can do an alpha release? Or just wait, it's not really that urgent.

dhermes:
WDYT about an alpha release?

nathanielmanistaatgoogle:
Make it so.

dhermes:
Here is a potential downside of doing this Picard: deleting a version from PyPI is not possible (or maybe just not easy).

Issue #212, opened 2015-07-02
Get to 100% code coverage


dhermes:
- [x] `oauth2client.client`
- [x] `oauth2client.contrib.appengine`
- [x] `oauth2client.contrib.django_orm`
- [x] `oauth2client.contrib.django_util.apps`
- [x] `oauth2client.contrib.locked_file`
- [x] `oauth2client.contrib.multistore_file`
- [x] `oauth2client.tools`
- [x] `oauth2client.util`
- [x] `tests.contrib.test_multistore_file` (#431, 3 branches: [1][2], [2][3], [3][4])
- [x] `tests.test_client` (#431, See [summary][1])
- [x] `tests.test_file` (#431, See [summary][1])

----

- [x] `oauth2client.__init__`
- [x] `oauth2client._helpers`
- [x] `oauth2client._openssl_crypt`
- [x] `oauth2client._pycrypto_crypt`
- [x] `oauth2client.clientsecrets` (PR: #278)
- [x] `oauth2client.contrib.__init__`
- [x] `oauth2client.contrib.django_util.__init__`
- [x] `oauth2client.contrib.django_util.decorators`
- [x] `oauth2client.contrib.django_util.signals`
- [x] `oauth2client.contrib.django_util.site`
- [x] `oauth2client.contrib.django_util.storage`
- [x] `oauth2client.contrib.django_util.views`
- [x] `oauth2client.contrib.flask_util` (Issue: #277, PR: #282)
- [x] `oauth2client.contrib.keyring_storage` (PR: #279)
- [x] `oauth2client.crypt` (PR: #290)
- [x] `oauth2client.devshell` (PR: #284)
- [x] `oauth2client.file`
- [x] `oauth2client.gce` (PR: #280)
- [x] ~~`oauth2client.old_run` (PR: #285)~~ yay it's gone
- [x] `oauth2client.service_account`
- [x] `oauth2client.xsrfutil` (PR: #281)
- [x] `tests.__init__`
- [x] `tests.contrib.__init__`
- [x] `tests.contrib.test_appengine` (See [summary][1])
- [x] `tests.contrib.test_django_orm`
- [x] `tests.contrib.test_django_settings`
- [x] `tests.contrib.test_django_util`
- [x] `tests.contrib.test_flask_util`
- [x] `tests.contrib.test_keyring_storage`
- [x] `tests.http_mock` (See [summary][1])
- [x] `tests.test__helpers`
- [x] `tests.test__pycrypto_crypt`
[...]

PR #215, opened 2015-07-07
Factoring out conditional code from `crypt.py`.

dhermes:
Until now, code that depended on PyCrypto or OpenSSL was
defined conditionally (e.g. indented) in `crypt.py`. Rather than
grouping all these together, we factor out the library specific
behavior into standalone modules (but make the modules
private / protected).

In addition, added a `_helpers.py` module with common behavior
that was previously defined in multiple places.

Finally, beefed up some test cases so that the three newly added
modules had 100% test coverage.

Towards #212.

### DIFF
--- a/oauth2client/_helpers.py
+++ b/oauth2client/_helpers.py
@@ -1,0 +1,53 @@
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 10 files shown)


nathanielmanistaatgoogle:
GitHub is saying that this is not auto-mergeable - does this need to be rebased on top of recent changes?

dhermes:
Yup #203 and this crossed paths. Rebasing now.

dhermes:
@nathanielmanistaatgoogle Rebased and pushed changes from #203 into this PR.

dhermes:
@nathanielmanistaatgoogle Bump

dhermes:
w00t! Can't wait to pull apart `client.py` and make this thing test-able (and suck less).

PR #216, opened 2015-07-08
Update for v1.4.12 release.

craigcitro:
PTAL @dhermes


dhermes:
LGTM

craigcitro:
Note that this is half of #214, specifically without the travis bits (until @nathanielmanistaatgoogle is back to decide on that part).

PR #217, opened 2015-07-08
Import webapp2 instead of google.appengine.ext.webapp

puccia:
This choice is apparently more robust in situations / runtimes where the import mechanism in `google/appengine/ext/webapp/__init__.py` does not appear to do the right thing.

### DIFF
--- a/oauth2client/appengine.py
+++ b/oauth2client/appengine.py
@@ -27,12 +27,12 @@ import pickle
 import threading
 
 import httplib2
+import webapp2 as webapp
 
 from google.appengine.api import app_identity
 from google.appengine.api import memcache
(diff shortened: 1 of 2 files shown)


puccia:
Covered by Google's own corporate agreement.

craigcitro:
yep, googlebot is confused because you're not (yet) in the google org. no worries there.

it looks like the docs build is failing; likely we just need to add `webapp2` to the list of packages we install for the docs build? are you familiar enough with `tox` that you could give it a shot? (you can just run `tox -e docs` locally to try it out, once you `pip install tox` as needed).

puccia:
Exactly what you wrote! Looks healthy now.

PR #220, opened 2015-07-10
Ignore PyCharm config files

frederikcreemers:

PR #221, opened 2015-07-10
Pinning mock at version 1.0.1.

dhermes:
Upgrading to both 1.1.0 and 1.1.1 causes unexplained failures that need to be investigated. This is a stop-gap measure to keep the tests passing and the changelog of `mock` will need to be inspected.

See https://travis-ci.org/google/oauth2client/builds/70413246 for an example failure (from #220).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,7 @@ envlist = py26,py27,py33,py34,pypy,cover
 
 [testenv]
 basedeps = keyring
-           mock
+           mock==1.0.1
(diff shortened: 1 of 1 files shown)


dhermes:
@nathanielmanistaatgoogle @anthmgoogle I am going to merge this once Travis passes so that other contributions won't fail for reasons unrelated to their code.

nathanielmanistaatgoogle:
LGTM; thanks for pursuing this.

PR #227, opened 2015-07-31
django_orm.py: python 3 compatibility

### LINKED ISSUE #168: "to_python" function not working in class CredentialsField with python3
I am a newbie with python.
Recently I was trying to build a small web app with Google API in Django and Python 3.4.

And I use the storage utilities from oauth2client.django_orm
```
from oauth2client.django_orm import Storage
...
storage = Storage(CredentialsModel, 'id', user, 'credential')
...
storage.put(credential)
...
credential = storage.get()
...
```

The storage.put() function work fine but when I use the storage.get() function, it just give me a string but not a credentials object.
After testing and reading the source code of oauth2client.django_orm.
I found the problem was the "to_python" function doesn't work and just return the Base64 data back.

Finally I found the cause from the internet: "In python 3 the module-global __ metaclass__ variable is no longer supported."
And it has to change like this to work in Python 3
```
class CredentialsField(models.Field, metaclass=models.SubfieldBase):
...
```

amezin:
Fixes #168 for me. Also, fix FlowField the same way.

### DIFF
--- a/oauth2client/django_orm.py
+++ b/oauth2client/django_orm.py
@@ -25,11 +25,10 @@ import base64
 import pickle
 
 from django.db import models
+from django.utils.six import with_metaclass
 from oauth2client.client import Storage as BaseStorage
 
-class CredentialsField(models.Field):
-
-  __metaclass__ = models.SubfieldBase
+class CredentialsField(with_metaclass(models.SubfieldBase, models.Field)):
 
   def __init__(self, *args, **kwargs):
     if 'null' not in kwargs:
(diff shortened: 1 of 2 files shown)


nathanielmanistaatgoogle:
You say "Fixes #168 for me" - how about test coverage so that this change can be said to fix #168 for everyone? :-)

amezin:
I've added two tests (one for each field class) that check field's metaclass.

But they won't improve coverage. Because problematic lines ('__metaclass__ = ') were already "covered".

Actually, I think, there should be tests that perform write to temporary (in-memory) database, and then try to read the data back. But I have no idea how to setup necessary Django environment in the test.

dhermes:
How would you set up the necessary environment if it wasn't in a test? Maybe we can help you get it working with `tox` / `unittest` / `nose`. FWIW there are probably existing plugins / test harnesses.

amezin:
If it wasn't in a test, I'd create a Django project, and run everything (including tests) using django-admin or ./manage.py. All I know is that django does a lot of setup before calling my code. I've found one tool that probably can run django tests as a part of nose test suite: https://pypi.python.org/pypi/NoseDjango. But it haven't been updated for a long time, works only with python 2 and years old versions of django.

PR #230, opened 2015-08-03
Credentials keep track of scopes.

### LINKED ISSUE #228: Credentials should keep track of which scopes are authorized
This will allow us to support [incremental auth](https://developers.google.com/+/web/api/rest/oauth#incremental-auth).

I propose:

1. We add a new property ``scopes`` to ``OAuth2Credentials``.
2. We add a new function ``has_scopes`` to ``OAuth2Credentials`` that determines if the provided scopes are a subset of the authorized scopes.
3. We modify ``OAuth2WebFlow`` to store the scopes when constructing the credentials.
4. We add a function ``refresh_scopes`` (or similar) to ``OAuth2Credentials`` that can get the canonical list of scopes from the [tokeninfo endpoint](https://www.googleapis.com/oauth2/v2/tokeninfo).

I'm willing to submit the PR for this, just want to validate my approach.

@anthmgoogle

theacodes:
- Added new attributes scopes and token_info_uri to OAuth2Credentials.
- Added method OAuth2Credentials.has_scopes to test which scopes the credentials have.
- OAuth2WebServerFlow now passes in the authorized scopes when constructing OAuth2Credentials.
- Added method Oauth2Credentials.refresh_scopes, which gets the canonical list of scopes from the OAuth2 tokeninfo service.
- Added new utility function string_to_scopes, the inverse of scopes_to_string.

This will fix #228 when merged.


oauth2client/client.py
```diff
@@ -614,6 +621,34 @@ def apply(self, headers):
...
+
+    Notes:
+      There are cases where the credentials are unaware of which scopes are
+      authorized. Notably, credentials obtained and stored before this code was
+      added will not have scopes, AccessTokenCredentials do not have scopes. In
+      both cases, you can use refresh_scopes() to obtain the canonical set of
+      scopes.
+    """
+    scopes = util.string_to_scopes(scopes)
+    return set(scopes).issubset(set(self.scopes))
```
dhermes: Maybe just store `set(self.scopes)` instead of `self.scopes`?

oauth2client/client.py
```diff
@@ -614,6 +621,34 @@ def apply(self, headers):
...
+
+  def refresh_scopes(self, http):
```
dhermes: Don't know if refresh is the right word here. Retrieve maybe?

oauth2client/client.py
```diff
@@ -873,6 +910,44 @@ def _do_revoke(self, http_request, token):
...
+    """
+    logger.info('Refreshing scopes')
+    query_params = {'access_token': token, 'fields': 'scope'}
+    token_info_uri = _update_query_params(self.token_info_uri, query_params)
+    resp, content  = http_request(token_info_uri)
+    if resp.status == 200:
+      d = json.loads(content)
+      self.scopes = util.string_to_scopes(d.get('scope', ''))
+    else:
+      error_msg = 'Invalid response %s.' % resp.status
```
dhermes: Easy habit to start: use `(resp.status,)` instead, so no unexpected tuple-unpacking occurs.

oauth2client/client.py
```diff
@@ -873,6 +910,44 @@ def _do_revoke(self, http_request, token):
...
+      try:
+        d = json.loads(content)
+        if 'error' in d:
+          error_msg = d['error_description']
```
dhermes, line 961: This doesn't match the check right before it, so could cause a `KeyError`.

oauth2client/client.py
theacodes: Is there a case where `resp.status` can be a tuple? This code is identical to the error handling code above in `_do_revoke`.

oauth2client/util.py
```diff
@@ -163,6 +163,26 @@ def scopes_to_string(scopes):
     return ' '.join(scopes)
 
 
+def string_to_scopes(scopes):
```
dhermes, line 166: I could've sworn this existed already.

**UPDATE**: `util.scopes_to_string` is the thing I was thinking of

oauth2client/client.py
theacodes, line 961: Good catch. Fixed.

oauth2client/client.py
dhermes: Not worried about this particular value, just a good habit to practice: pass a tuple with `%` formatting.

theacodes: Alrighty, updated it anyway.

oauth2client/client.py
theacodes: Summoning @anthmgoogle for his take.

oauth2client/client.py
theacodes: Hmm, not a bad idea, since they should be considered more or less immutable from the user perspective. Will do.

theacodes: Ugh, sets aren't json serializable. It may just be easier to store them as a list.

dhermes: Sad face. SGTM

anthmgoogle: Is it possible to store this as a set, but to convert to a space separated string when serializing or deserializing?

dhermes: Yes. List wrap it with a `list()` or a `sorted()` in `to_json`.

oauth2client/client.py
anthmgoogle: Perhaps something more explicit, e.g. update_scopes_from_tokens? That is long, but this is a bit of an advanced scenario so it may be appropriate.

oauth2client/client.py
```diff
@@ -873,6 +910,44 @@ def _do_revoke(self, http_request, token):
...
+      d = json.loads(content)
+      self.scopes = util.string_to_scopes(d.get('scope', ''))
+    else:
+      error_msg = 'Invalid response %s.' % (resp.status,)
+      try:
+        d = json.loads(content)
+        if 'error_description' in d:
+          error_msg = d['error_description']
+      except (TypeError, ValueError):
+        pass
```
anthmgoogle, line 963: Why are these skipped? Perhaps a comment for context?

anthmgoogle:
Finished review pass.

oauth2client/client.py
theacodes: let's be fully explicit - `update_scopes_from_access_token`, as it'll only work with an access token.

theacodes: Actually I don't like that either, it doesn't imply that a server request is involved like `refresh` and `revoke` do.

oauth2client/client.py
theacodes, line 963: Type and Value error indicate that the json was invalid or that error_description doesn't exist. In this case, the full content of the response will be used in the error message. This mimics the same error handling in `_do_revoke`.

oauth2client/client.py
theacodes: @anthmgoogle why a space separated string? 

I was able to modify `_to_json` to work with a set, but modifying it to do a space separated string might be more complicated than it's worth.

jonparrott:
@anthmgoogle pinging for another review pass.

I'm going to look into the python 3 issues.

jonparrott:
Python 3 is now passing on travis, and I fixed a python 3 bug in ``_do_revoke`` as well.

oauth2client/client.py
```diff
@@ -858,6 +897,10 @@ def _do_revoke(self, http_request, token):
     query_params = {'token': token}
     token_revoke_uri = _update_query_params(self.revoke_uri, query_params)
     resp, content = http_request(token_revoke_uri)
+
+    if six.PY3 and isinstance(content, bytes):
```
dhermes, line 906: I'm going to put this in another PR. FWIW this is also pointed out to me that we detect and convert to/from bytes in a very inconsistent way.

You can rebase after it's in?

theacodes, line 906: SGTM.

anthmgoogle:
LGTM.

oauth2client/client.py
dhermes, line 906: It's growing too quickly. Just merge this and I'll deal with it in parallel.

jonparrott:
Seems like the only outstanding item is the name of ``refresh_scopes``.

jonparrott:
@anthmgoogle @dhermes can I get your feedback on the naming of ``refresh_scopes``?

My thoughts are that It should be consistent with the other methods that cause http requests - ``refresh`` and ``revoke``, it shouldn't seem like it's not making a request, so I don't like ``get_scopes``.

dhermes:
1. `get` doesn't sound like making a request?
1. There is nothing to refresh. It's a retrieval (which potentially overwrites the local version)
1. Hilariously, past me said "Don't know if refresh is the right word here. Retrieve maybe?" and current me also said the exact same thing without remembering

jonparrott:
I'm okay with ``retrieve_scopes`` over ``get_scopes`` or ``refresh_scopes``. ``get`` always feels like an accessor (ie local data) to me. @anthmgoogle or @elibixby to confirm.

anthmgoogle:
I think refreshing the scopes from the token is a bit of a corner case, so it may be worth considering a longer and more explicit name like "refresh_scopes_from_token". But I am fine with "refresh_scopes". I agree that "get" is a problem. I think "retrieve" could be confusing because it is such a synonym of "get".

dhermes:
The synonym problem doesn't apply here. `get` is an issue because programmers associate that with getters and setters (i.e. local data), but `retrieve` has no such baggage. They are not being "refreshed". I don't care enough to argue though, so whichever you choose is fine.

Also, Python programmers don't expect to need getters and setters.

jonparrott:
:laughing: this is fantastic.

I think ``retrieve_scopes`` is best, as ``refresh_scopes_from_token`` is a bit of a misnomer (``refresh_scopes_using_token`` is more correct). So, I'm switching it to that and hopefully no one uses it enough to be bothered by the name. :stuck_out_tongue:

anthmgoogle:
LGTM.

dhermes:
@jonparrott LMK when to merge

jonparrott:
@dhermes should be good to go once travis is green.

jonparrott:
pypy tests failed because of cryptography - I think you mentioned this in another issue.

It's ready to merge.

PR #235, opened 2015-08-06
Adding favicon to match logo.

dhermes:
Also removing empty line at the end of build-docs.


scripts/build-docs
```diff
@@ -28,4 +28,3 @@ sphinx-apidoc -f -o docs/source oauth2client
 cd docs
 make html
 cd ..
-
```
nathanielmanistaatgoogle, line 31: Is this change pertinent to this commit? If so, how? If not, drop it.

nathanielmanistaatgoogle:
How sure are we that that's the file we wish to use? Are we following another project's example?

scripts/build-docs
dhermes, line 31: Nope. Will do.

dhermes:
I was just basing it off of the contents of `docs/_static/google_logo.png`

dhermes:
I removed the superfluous change and sent it as #236

PR #237, opened 2015-08-06
Making tweaks to allow project to build on readthedocs.org.

dhermes:
In particular
- Adding a requirements.txt file to aid in installing docs-only
  dependencies (this is a feature of readthedocs, see
  https://docs.readthedocs.org/en/latest/builds.html)
- Updating comment before `import django` in docs/conf.py. Without
  the setup that or tox `docs` env provides or the dependencies
  in docs/requirements.txt, sphinx-build will fail due to missing
  `django`
- Adding a section to docs/conf.py which downloads the
  App Engine SDK and adds it to the import path
- Adding base (non-docs) dir so sources get added when generating
  HTML (e.g. sys.path.append('..') within docs/ dir)
- Using package distribution for docs version instead of manually
  importing oauth2client in docs/conf.py.

@nathanielmanistaatgoogle See http://oauth2client.readthedocs.org/en/rtd-test/ for a final view of what this looks like (and compare it to http://google.github.io/oauth2client/)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -4,8 +4,14 @@
 # sphinx-quickstart on Wed Dec 17 23:13:19 2014.
 #
 
-import sys
 import os
+from pkg_resources import get_distribution
+import sys
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+sys.path.insert(0, os.path.abspath('..'))
 
 # -- General configuration ------------------------------------------------
 
@@ -24,14 +30,16 @@ project = u'oauth2client'
 copyright = u'2014, Google, Inc'
 
 # Version info
-import oauth2client
-version = oauth2client.__version__
-release = oauth2client.__version__
+distro = get_distribution('oauth2client')
+version = distro.version
+release = distro.version
 
 exclude_patterns = ['_build']
 
 # In order to load django before 1.7, we need to create a faux
-# settings module and load it.
+# settings module and load it. This assumes django has been installed
+# (but it must be for the docs to build), so if it has not already
+# been installed run `pip install -r docs/requirements.txt`.
 import django
 if django.VERSION[1] < 7:
   sys.path.insert(0, '.')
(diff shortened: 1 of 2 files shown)


dhermes:
@nathanielmanistaatgoogle Anything holding back a review here?

docs/requirements.txt
```diff
@@ -0,0 +1,9 @@
+django
+flask
+keyring
+pycrypto==2.6
```
nathanielmanistaatgoogle: How much do these have to be ==? Why can't this and the next line be >=?

dhermes: It was copy-pasta from `tox.ini`. Good call.

docs/conf.py
```diff
@@ -40,7 +48,24 @@
 # -- Options for HTML output ----------------------------------------------
 
 # We want to set the RTD theme, but not if we're on RTD.
-if os.environ.get('READTHEDOCS', '') != 'True':
+if os.environ.get('READTHEDOCS', '') == 'True':
```
nathanielmanistaatgoogle: My case-sensitivity-bug spidey-sense is tingling. How important is it that values such as "TRUE", "true", "tru", and "tRuE" cause this expression to evaluate to False?

nathanielmanistaatgoogle:
Thanks for the poke.

docs/conf.py
dhermes: Mine too.

It is what the RTD folks [recommend](https://docs.readthedocs.org/en/latest/faq.html#how-do-i-change-behavior-for-read-the-docs). Honestly I'd rather just do `os.environ.has_key('READTHEDOCS')`.

nathanielmanistaatgoogle: That example uses None as the second argument in the call to os.environ.get, and that's what I prefer as well. Any reason you went with the empty string instead?

dhermes: Craig did it, I just changed the existing stuff in `docs/conf.py`.

I'll do whatever you like, just let me know.

nathanielmanistaatgoogle: None, please. Any other value calls attention to itself and invites the question "now why isn't that None?".

dhermes: Shall I just use `os.getenv`? Also, what value should I compare it to (case-sensitive or not)?

nathanielmanistaatgoogle: Let's follow their example.

Just using os.getenv sounds like a bad idea in case the environment is ever populated with a "READTHEDOCS"/"False" key-value pair.

dhermes: I meant `os.getenv('foo')` is equivalent to `os.environ.get('foo', None)` is equivalent to `os.environ.get('foo')`. Sending now.

dhermes:
@nathanielmanistaatgoogle Change (from `''` to `None`) sent

nathanielmanistaatgoogle:
LGTM and waiting on test results (which I believe will be forthcoming...).

dhermes:
Looks like the `docs` env failed to fetch the GAE SDK, I restarted it and amazingly didn't have to restart the others (good job Travis?).

dhermes:
Succeeded on rebuild.

dhermes:
Victory: http://oauth2client.readthedocs.org

PR #238, opened 2015-08-06
Adding a high level title to contributing.rst.

dhermes:
Without it, all the 5 headers in the file show up in the main TOC.

e.g. without this change we have

screen_shot_001

and with it we have

screen_shot_003

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -1,3 +1,6 @@
+Contributing
+============
+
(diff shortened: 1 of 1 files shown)


dhermes:
@nathanielmanistaatgoogle Do you guys have a contact at Travis? WTF is going on? `gcloud-python` isn't having build issues, so I think it's specific to this project

PR #155, opened 2015-03-30
Adding CONTRIBUTING.md for future maintainers.

dhermes:
@nathanielmanistaatgoogle See
https://github.com/dhermes/oauth2client/blob/add-contrib/CONTRIBUTING.md
for the pretty rendered version of the file

Spinning off from previous PR based from https://github.com/google/oauth2client/pull/153#issuecomment-87793610

PR #244, opened 2015-08-07
Moving CONTRIBUTING out of docs to repository top-level.

dhermes:
Also unifying the old `docs/contributing.rst` with the current contents of `CONTRIBUTING.md`.


nathanielmanistaatgoogle:
"... and into repository" sounds like content is coming into source control from outside, which isn't the case. How about "Move CONTRIBUTING to repository top-level"?

dhermes:
@nathanielmanistaatgoogle I also rewrote the original commit to say the same.

While doing it I realized that there is an existing `CONTRIBUTING.md`, which I added in #155!

I suggest we unify these two documents in a follow-up PR. WDYT?

dhermes:
I should also mention that you can see the rendered RST on my branch: https://github.com/dhermes/oauth2client/blob/move-contrib/CONTRIBUTING.rst

nathanielmanistaatgoogle:
Yes; it would be rude of us to maintain CONTRIBUTING.rst and CONTRIBUTING.md as separate documents.

dhermes:
@nathanielmanistaatgoogle I went ahead and merged the two together. Can be seen at: https://github.com/dhermes/oauth2client/blob/move-contrib/CONTRIBUTING.md

nathanielmanistaatgoogle:
LGTM!

PR #246, opened 2015-08-07
Removing unnecessary top-level TOC for all modules.

dhermes:
If we had more than one base package, it would be relevant, but we only have oauth2client, so it is irrelevant and just visually distracting.


dhermes:
Current TOC:

screen_shot_012

TOC after this change:

screen_shot_013

PR #248, opened 2015-08-07
Autodoc the optional modules (private).

dhermes:
This is relevant because users may actually interact with the classes in _openssl_crypt and _pycrypto_crypt depending on which crypto library they have installed.


nathanielmanistaatgoogle:
How is it that users might actually interact with elements of those modules? Do we have an abstraction violation somewhere with something that we think should be private actually being exposed?

dhermes:
Anything that begins with `_` won't be exposed, so it's really based on the things we name in those modules.

Running `tox -e docs` (you can checkout the branch and run it if you like), the `oauth2client._helpers` has nothing documented, `oauth2client._openssl_crypt` has `OpenSSLSigner`, `OpenSSLVerifier`, `pkcs12_key_as_pem` and `oauth2client._pycrypto_crypt` has `PyCryptoSigner` and `PyCryptoVerifier`.

nathanielmanistaatgoogle:
What interaction can users expect to have with those classes? Will users construct them or possibly encounter instances of them by some other means?

dhermes:
I guess not really. They are used in `client.py` to detect if pyOpenSSL is installed (we should do a better job on that) and then within public functions in `crypt.py` and the protected `SignedJwtAssertionCredentials._generate_assertion`.

nathanielmanistaatgoogle:
I'm still confused - private code elements may be used within a public function but don't count as being improperly exposed unless they cross the surface of the function either as parameters or return values. What do users of the library need to know about these modules in order to use the library? Where do the code elements being documented in this pull request cross the public surface area of the library?

dhermes:
That was the "I guess not really" part. They don't. We can happily close this PR if you like.

nathanielmanistaatgoogle:
Glad to see it happen happily. :-)

PR #249, opened 2015-08-07
Removing outdated docs from gh-pages

### LINKED ISSUE #240: Remove gh-pages (defer to readthedocs.org)
Current gh-pages branch is serving at, but it doesn't need to be: http://google.github.io/oauth2client/

We could put a redirect there to go to readthedocs.org

Also, I was under the impression that `gh-pages` was automatically updated by Travis, but it turns out that it was a manual process, so the docs are out of date.

From [`gh-pages` log][1]  it looks like @craigcitro was doing it by hand, with July 8 the [most recent][2].

PS See #237 for context.

[1]: https://github.com/google/oauth2client/commits/gh-pages
[2]: https://github.com/google/oauth2client/commit/8efa3f76893f8bc381fd96ad283f885a97d2176d

dhermes:
Fixes #240

---

You can check that the redirect is working on:
https://dhermes.github.io/oauth2client/

I don't think custom 404's work, it seems to be using the one I have setup for my [blog](https://github.com/dhermes/dhermes.github.io), which gets [served](https://github.com/dhermes/dhermes.github.io/blob/master/CNAME) on https://blog.bossylobster.com/.

So for example when you visit
https://dhermes.github.io/oauth2client/source/modules.html
you get the 404 for my blog.

Seems to be a good explanation: http://webmasters.stackexchange.com/a/56986

---

An alternative to a meta redirect would be just a small page saying, hey we moved and letting users click a link if they want.


nathanielmanistaatgoogle:
Yeah, I think I have a slight preference for a "hey we've moved!" page.

dhermes:
OK. I'll try to give it a stab.

BTW I should have mentioned that the PR is easier tor review by commits.

nathanielmanistaatgoogle:
I always review every pull request commit-by-commit.

Why bother maintaining the branch history and adding a commit that removes all the content at the branch? Why not just create one new commit with the "we've moved our docs" page and point the gh-pages branch at it?

dhermes:
So you want me to nuke the whole history? Seems unnecessary / drastic.

nathanielmanistaatgoogle:
Not nuke, just simply decline to carry forward. Declining to list an old commit as a parent of a new commit isn't an inherently destructive act, is it? Who says that all commits ever committed to a repository must be ancestors reachable from the current branches of the repository?

Let's look at it a different way by asking a different question: why is it important to project history going forward that the generated documentation was in the past hosted in a non-master branch in the repository?

Ah, source control philosophy! I could do this all day. :-P

dhermes:
So I think this will require a `git push -f` and then GitHub will GC all the commits that aren't in a branch anymore. (Hence the "nuke" description.)

nathanielmanistaatgoogle:
I'm in favor of nuking. It was generated content all along, right? The fact that we stored it in git was a reflection of "this is how we serve our documentation" and not "these are original artifacts that belong in source control", right?

dhermes:
But then we lose the "history" of "this is how we serve our documentation". Since GitHub is hosting it, and  no one will need to actually check this branch out, who is hurt by leaving the commits? (I'm really close to just giving up and doing what you say :smile:)

nathanielmanistaatgoogle:
I find your last point persuasive, actually. :-) So now this just needs a "we've moved" page?

dhermes:
Yeah. I'll hunt for some already made HTML/CSS so I don't have to use my brain.

dhermes:
I'm going to use the GitHub generated one (stolen from http://jcgregorio.github.io/httplib2/)

dhermes:
@nathanielmanistaatgoogle 

Checkout https://dhermes.github.io/oauth2client/ (redirects to my blog https://blog.bossylobster.com/oauth2client/)

dhermes:
w00t: http://google.github.io/oauth2client/

Unfortunately (as predicated) the 404 doesn't do anything: http://google.github.io/oauth2client/foo

PR #250, opened 2015-08-10
Unifying all conversions to bytes.

dhermes:
Also making sure the _urlsafe_b64encode returns bytes and updating dependent code with the change in return type.

**NOTE**: This doesn't address the opposite direction, i.e. `a.decode('utf-8')` where is a `bytes` (or `str`) object.

### DIFF
--- a/oauth2client/_helpers.py
+++ b/oauth2client/_helpers.py
@@ -39,15 +39,42 @@ def _json_encode(data):
   return json.dumps(data, separators=(',', ':'))
 
 
+def _to_bytes(value, encoding='ascii'):
+  """Converts a string value to bytes, if necessary.
+
+  Unfortunately, ``six.b`` is insufficient for this task since in
+  Python2 it does not modify ``unicode`` objects.
+
+  Args:
+      value: The string/bytes value to be converted.
+      encoding: The encoding to use to convert unicode to bytes. Defaults
+                to "ascii", which will not allow any characters from ordinals
+                larger than 127. Other useful values are "latin-1", which
+                which will only allows byte ordinals (up to 255) and "utf-8",
+                which will encode any unicode that needs to be.
+
+  Returns:
+      The original value converted to bytes (if unicode) or as passed in
+      if it started out as bytes.
+
+  Raises:
+      ValueError if the value could not be converted to bytes.
+  """
+  result = (value.encode(encoding)
+            if isinstance(value, six.text_type) else value)
+  if isinstance(result, six.binary_type):
+    return result
+  else:
+    raise ValueError('%r could not be converted to bytes' % (value,))
+
+
 def _urlsafe_b64encode(raw_bytes):
-  if isinstance(raw_bytes, six.text_type):
-    raw_bytes = raw_bytes.encode('utf-8')
-  return base64.urlsafe_b64encode(raw_bytes).decode('ascii').rstrip('=')
+  raw_bytes = _to_bytes(raw_bytes, encoding='utf-8')
+  return base64.urlsafe_b64encode(raw_bytes).rstrip(b'=')
 
 
 def _urlsafe_b64decode(b64string):
   # Guard against unicode strings, which base64 can't handle.
-  if isinstance(b64string, six.text_type):
-    b64string = b64string.encode('ascii')
+  b64string = _to_bytes(b64string)
   padded = b64string + b'=' * (4 - len(b64string) % 4)
   return base64.urlsafe_b64decode(padded)
--- a/tests/test__helpers.py
+++ b/tests/test__helpers.py
@@ -17,6 +17,7 @@ import unittest
 
 from oauth2client._helpers import _json_encode
 from oauth2client._helpers import _parse_pem_key
+from oauth2client._helpers import _to_bytes
 from oauth2client._helpers import _urlsafe_b64decode
 from oauth2client._helpers import _urlsafe_b64encode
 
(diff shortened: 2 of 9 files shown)


oauth2client/_helpers.py
```diff
@@ -39,15 +39,42 @@ def _json_encode(data):
...
+                which will encode any unicode that needs to be.
+
+  Returns:
+      The original value converted to bytes (if unicode) or as passed in
+      if it started out as bytes.
+
+  Raises:
+      ValueError if the value could not be converted to bytes.
+  """
+  result = value
```
nathanielmanistaatgoogle: Try to avoid

``` python
my_field = my_first_expression
if my_condition:
  my_field = my_second_expression
```

Would you mind restating this as?

``` python
result = value.encode(encoding) if isinstance(value, six.text_type) else value
if isinstance(result, six.binary_type):
  return result
else:
  raise ValueError('%r could not be converted to bytes' % (value,))
```

oauth2client/_helpers.py
```diff
@@ -39,15 +39,42 @@ def _json_encode(data):
...
+  if isinstance(value, six.text_type):
+    result = value.encode(encoding)
+  if not isinstance(result, six.binary_type):
+    raise ValueError('%r could not be converted to bytes' % (value,))
```
nathanielmanistaatgoogle, line 68: The tupleization of value is weird here - is it something you can avoid by using string.format rather than %?

oauth2client/_helpers.py
```diff
@@ -39,15 +39,42 @@ def _json_encode(data):
   return json.dumps(data, separators=(',', ':'))
 
 
+def _to_bytes(value, encoding='ascii'):
```
nathanielmanistaatgoogle, line 42: I'm inherently skeptical of optional arguments in private functions, and I think the call sites of this function are more clear if the encoding is present and explicit at each one.

nathanielmanistaatgoogle, line 42: Since (elsewhere in this commit) _to_bytes is used outside this module, make it module-public by naming it "to_bytes". I think this may break a pattern otherwised used in this codebase but I think it's the right thing and it's worth it. The function won't be public-to-users-of-the-oauth2client-library because _helpers is itself a library-private module.

tests/test__helpers.py
```diff
@@ -49,17 +50,35 @@ def test_list_input(self):
...
+    self.assertEqual(_to_bytes(value), encoded_value)
+
+  def test_with_nonstring_type(self):
+    value = object()
+    self.assertRaises(ValueError, _to_bytes, value)
+
+
 class Test__urlsafe_b64encode(unittest.TestCase):
 
+  DEADBEEF_ENCODED = b'ZGVhZGJlZWY'
```
nathanielmanistaatgoogle, line 71: Module-scope constant rather than class-scope constant? Any reason for this to be at class scope?

nathanielmanistaatgoogle:
As they are used throughout this commit what's the difference between "isinstance(value, bytes)" and "isinstance(value, six.binary_type)"? I imagine the distinction is subtle and tricky?

oauth2client/_helpers.py
dhermes, line 68: `str.format` is not properly supported in Python 2.6 (AFAIK). The tupleization is very much intentional, as it avoids spurious errors (e.g. if `value` was already a `tuple`). The fact that `str.__mod__` (e.g. `str %`) accepts a single val instead of a tuple / iterable is kind of a "mistake".

oauth2client/_helpers.py
dhermes: Don't mind it, but find it harder to read.

tests/test__helpers.py
dhermes, line 71: It's only used in this class. I'm for scope reduction.

oauth2client/_helpers.py
dhermes, line 42: OK. I never know where to come down on that naming with `_foo.py` modules.

oauth2client/_helpers.py
nathanielmanistaatgoogle, line 68: "String Formatting / New in version 2.6": https://docs.python.org/2.6/library/string.html#string-formatting

I agree with you about the mistake in %.

dhermes:
AFAIK the only difference is that `bytes` is an alias for `str` in Py2.6 and Py2.7 but it doesn't exist in Py2.5, so `six.py` just [says][1]:

```python
if PY3:
    binary_type = bytes
else:
    binary_type = str
```

[1]: https://bitbucket.org/gutworth/six/src/8a545f4e906f6f479a6eb8837f31d03731597687/six.py?at=default#six.py-40

oauth2client/_helpers.py
dhermes, line 68: So do you want me to use `.format`?

nathanielmanistaatgoogle, line 68: No, I now see that it's fine as it is. Sorry for the wild goose chase.

dhermes:
@nathanielmanistaatgoogle I rebased after some of the changes made it un-mergeable

Things that remain:

- Use of ternary expression
- Using `.format` instead of `% (varname,)`.
- Renaming `_to_bytes` to `to_bytes` (this will make it a public symbol in modules it gets used in)

----

Seems this will mostly cover #157

nathanielmanistaatgoogle:
* Yes to [conditional expression](https://docs.python.org/2/reference/expressions.html#conditional-expressions)
* No, not unless I persuaded you, and I don't think I did because I think you persuaded me
* Oh crap, that's because this codebase has a feel-free-to-import-elements-of-modules policy instead of import-only-modules-not-elements-of-modules policy. Oof, leave the underscore for now in that case.

dhermes:
@nathanielmanistaatgoogle I folded the ternary change into the original PR.

It's more common in Python to import the objects you need rather than the namespaces you need. That's just a Google rule. When unit testing, it's easier (maybe only marginally so, maybe a lot?) to monkey patch the objects in the module under test rather than tracking down the namespace which has that object and then monkey patching that other module.

dhermes:
I learned a little `git`-fu for anyone who wants to learn:

For #136:

```bash
$ # Visit: https://api.github.com/repos/google/oauth2client/pulls/136/commits
$ # parent in MASTER: 3c02a812165004aeb18cc89e10639694bf2d3b0f
$ git branch pr-136 3c02a812165004aeb18cc89e10639694bf2d3b0f
$ git checkout pr-136
$ wget https://patch-diff.githubusercontent.com/raw/google/oauth2client/pull/136.patch
$ git am 136.patch
```

For #157 :

```bash
$ # Visit: https://api.github.com/repos/google/oauth2client/pulls/157/commits
$ # parent in MASTER: 0a6241c792f0c833f2f176292b0c7ea5623eabfd
$ git branch pr-157 0a6241c792f0c833f2f176292b0c7ea5623eabfd
$ git checkout pr-157
$ wget https://patch-diff.githubusercontent.com/raw/google/oauth2client/pull/157.patch
$ git am 157.patch
```

Now I will try to rebase them onto `master` and see how it goes.

Issue #257, opened 2015-08-12
Fail builds if `tox -e docs` generates content that isn't checked in yet.


dhermes:
See #234 for an example of newly generated content

dhermes:
I previously mentioned this: https://github.com/google/oauth2client/issues/240#issuecomment-128783932

PR #259, opened 2015-08-12
Using a separate autodoc page for each module.

dhermes:
I was working on #257 and reading http://sphinx-doc.org/man/sphinx-apidoc.html and noticed that `--separate` is a useful flag (to keep the docs pages relatively small and digestible).

### DIFF
--- a/docs/source/oauth2client.appengine.rst
+++ b/docs/source/oauth2client.appengine.rst
@@ -1,0 +1,7 @@
+oauth2client.appengine module
+=============================
+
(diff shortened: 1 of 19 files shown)


nathanielmanistaatgoogle:
What do we know about the failing Travis build?

dhermes:
It looks like it's a consistent failure.

Probably related to the fact that `cryptography==1.0` was [released][1] yesterday.

[1]: https://pypi.python.org/pypi/cryptography

PR #260, opened 2015-08-12
Making docs tox target fail on untracked changes.

### LINKED ISSUE #257: Fail builds if `tox -e docs` generates content that isn't checked in yet.

dhermes:
Checking if sphinx-apidoc generates new content and failing the script if the content is not checked in to version control.

Fixes #257.

@nathanielmanistaatgoogle I debated adding a `tox` target that would just generate the new content, but `tox -e docs` will still do this before failing, after which a user could then

``` bash
$ git add docs/
$ tox -e docs  # This time it won't fail
```

### DIFF
--- a/scripts/build-docs
+++ b/scripts/build-docs
@@ -27,6 +27,13 @@ rm -rf docs/_build/* docs/source/*
 sphinx-apidoc -f -o docs/source oauth2client
 # We only have one package, so modules.rst is overkill.
 rm -f docs/source/modules.rst
+
+# If anything has changed
+if [[ -n "$(git diff -- docs/)" ]]; then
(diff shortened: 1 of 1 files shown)


dhermes:
@nathanielmanistaatgoogle If you want to see it in action you can check out https://github.com/dhermes/oauth2client/tree/fake-change and run `tox -e docs`.

If you already have the code checked out locally (from here)

```bash
$ git remote add dhermes https://github.com/dhermes/oauth2client/
$ git fetch dhermes
$ git branch fake-change dhermes/fake-change
$ git checkout fake-change
$ tox -e docs
```

nathanielmanistaatgoogle:
(Merging; PyPy platform test failures are unrelated.)

PR #263, opened 2015-08-13
Making sure libgmp is ignored in pypy tox env.

dhermes:
This is because when libgmp is detected, installing PyCrypto tries to build _fastmath.c. But in PyPy this
fails because _fastmath.c tries to access parts of the CPython implementation.

See for `with_gmp=no` suggestion:
https://github.com/dlitz/pycrypto/pull/59#issuecomment-86569001

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,9 @@ basedeps = keyring
            flask
 deps = {[testenv]basedeps}
        django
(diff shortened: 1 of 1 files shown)


nathanielmanistaatgoogle:
Needs conflict resolution.

dhermes:

PR #265, opened 2015-08-13
Updating docs and CONTRIBUTING file with pypy tips.

dhermes:
As it turns out, both of our optional crypto dependencies have issues in pypy.

See it on https://github.com/dhermes/oauth2client/blob/add-pypy-note-to-contributing/CONTRIBUTING.md

@nathanielmanistaatgoogle I have two potential uneasy things here
1. Maybe this should go in `README`
2. Maybe it should be in its own section in `CONTRIBUTING` (rather than in "Running Tests")

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -121,6 +121,10 @@ Running Tests
     $ tox
     ```
 
+-   In order to run the `pypy` environment (in `tox`) you'll need at
+    least version 2.6 of `pypy` installed. See the [docs][13] for
+    more information.
+
 Running System Tests
 --------------------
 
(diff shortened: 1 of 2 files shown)


nathanielmanistaatgoogle:
My impression is that it should go in `README` - the content seems to be about running the software, and what's there applies whether the context of running the software is running the software's tests or doing something else with it. And all the information there is valid whether or not the reader intends to contribute patches or go live under a rock for the next century, so CONTRIBUTING definitely feels like the wrong place?

dhermes:
@nathanielmanistaatgoogle I put most of it in the docs and left a tiny reference in `CONTRIBUTING` since the `pypy` target will fail on too early a version of `pypy`. (Relevant to testing.)

PR #271, opened 2015-08-14
Rebased version of #136

dhermes:
This will be closed, I just wanted to show the output of Travis after re-basing.


dhermes:
As can be seen, this change is no longer merge-able (I think do to other changes underneath it):
https://travis-ci.org/google/oauth2client/builds/75641281

PR #272, opened 2015-08-14
Adding _from_bytes helpers as a foil for _to_bytes.

### LINKED ISSUE #239: [Python 3] oauth exception (the JSON object must be str, not 'bytes') when using gcloud.storage
In <https://github.com/google/oauth2client/blob/0aad0a09f50312e0fcc3a1234f0e7fb525010bfd/oauth2client/gce.py#L86>, the code uses `d = json.loads(content)` to load `content` as JSON. The code is broken in Python 3 because `content` is binary, while `json.loads` expects string.

dhermes:
@nathanielmanistaatgoogle This is the child of #157.

I labeled it don't merge because I essentially just looked around for `.decode` and tried to preserve a lot of the stuff from the other PR, but haven't written any new tests.

Let's discuss (and I'll iterate) before merging.

**UPDATE**: I should note that I rebased the original PR if anyone wants to have a look: https://github.com/dhermes/oauth2client/tree/pr-157-rebased

Fixes #239 (~~but no tests yet~~ tested by `AssertionCredentialsTests.test_refresh_success_bytes`)

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -40,6 +40,7 @@ from oauth2client import GOOGLE_DEVICE_URI
 from oauth2client import GOOGLE_REVOKE_URI
 from oauth2client import GOOGLE_TOKEN_URI
 from oauth2client import GOOGLE_TOKEN_INFO_URI
+from oauth2client._helpers import _from_bytes
 from oauth2client._helpers import _to_bytes
 from oauth2client._helpers import _urlsafe_b64decode
 from oauth2client import clientsecrets
@@ -279,8 +280,7 @@ class Credentials(object):
       An instance of the subclass of Credentials that was serialized with
       to_json().
     """
-    if isinstance(s, bytes):
-      s = s.decode('utf-8')
+    s = _from_bytes(s)
     data = json.loads(s)
     # Find and call the right classmethod from_json() to restore the object.
     module = data['_module']
@@ -673,8 +673,7 @@ class OAuth2Credentials(Credentials):
     Returns:
       An instance of a Credentials subclass.
     """
-    if isinstance(s, bytes):
-      s = s.decode('utf-8')
+    s = _from_bytes(s)
     data = json.loads(s)
     if (data.get('token_expiry') and
         not isinstance(data['token_expiry'], datetime.datetime)):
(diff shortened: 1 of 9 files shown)


dhermes:
I should also FYI to @saaros that this is happening

oauth2client/client.py
```diff
@@ -279,8 +280,7 @@ def new_from_json(cls, s):
       An instance of the subclass of Credentials that was serialized with
       to_json().
     """
-    if isinstance(s, bytes):
-      s = s.decode('utf-8')
+    s = _from_bytes(s)
```
nathanielmanistaatgoogle: Might it be worth using a different name for the result of _from_bytes, since it is potentially of a different type?

oauth2client/client.py
```diff
@@ -279,8 +280,7 @@ def new_from_json(cls, s):
       An instance of the subclass of Credentials that was serialized with
       to_json().
```
nathanielmanistaatgoogle, line 282: Should the "Args" section of this doc string be changed to better describe the input type as a byte string?

nathanielmanistaatgoogle:
Looks mostly good; what else do you want to do here?

oauth2client/client.py
dhermes: The old code was using `s = s.decode('utf-8')`, so I was just being mindless. Will change to `s_bytes` (and use a better varname than `s`).

New commit or fold into original?

oauth2client/client.py
dhermes, line 282: byte string or unicode string. `_from_bytes` will take something of `six.text_type` and just leave it as-is.

oauth2client/client.py
nathanielmanistaatgoogle: New commit is fine; thanks for asking.

oauth2client/client.py
nathanielmanistaatgoogle, line 282: In that case, let's be explicit about supporting both? Documentation that just says "string" suggests to the Python 3 reader that we're not describing the value completely.

dhermes, line 282: Yup. I updated the whole docstring (first line was too long).

dhermes:
@nathanielmanistaatgoogle Sent the new commit.

I was looking for places that might need testing but kind of realized that the whole reason we test `_from_bytes` with all inputs is so that we don't have to send each of those types through every single code path which hits `_from_bytes`.

Also note I massively updated the `Credentials.new_from_json` method but left `s` as the only argument. This is because it is part of the public interface, so people can execute

```
Credentials.new_from_json(s=file_obj.read())
```

nathanielmanistaatgoogle:
Looks good to me - anything remaining to be done here before merge? Anyone else who should review this?

dhermes:
No and no. Merging.

Issue #252, opened 2015-08-10
JwtAccess Credentials Feature


anthmgoogle:
Tracking completion of implementation of JwtAccess Credentials for service accounts for Python. This is a service account that passes a JWT directly as an a bearer token instead of exchanging for an access token. 

For reference, the Java implementation:
https://github.com/google/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java

dhermes:
Cool. See #211 as well. May be able to do them all with one class.

dhermes:
@nathanielmanistaatgoogle This "should be" sufficient:

```diff
diff --git a/oauth2client/service_account.py b/oauth2client/service_account.py
index f18f192..5c8ec9d 100644
--- a/oauth2client/service_account.py
+++ b/oauth2client/service_account.py
@@ -71,6 +71,8 @@ class ServiceAccountCredentials(AssertionCredentials):
                         header of a JWT token assertion.
         client_id: string, (Optional) Client ID for the project that owns the
                    service account.
+        audience: string, (Optional) An audience to use for scopeless
+                  credentials.
         user_agent: string, (Optional) User agent to use when sending
                     request.
         kwargs: dict, Extra key-value pairs (both strings) to send in the
@@ -97,6 +99,7 @@ class ServiceAccountCredentials(AssertionCredentials):
                  scopes='',
                  private_key_id=None,
                  client_id=None,
+                 audience=None,
                  user_agent=None,
                  **kwargs):
 
@@ -109,6 +112,7 @@ class ServiceAccountCredentials(AssertionCredentials):
         self._private_key_id = private_key_id
         self.client_id = client_id
         self._user_agent = user_agent
+        self._audience = audience
         self._kwargs = kwargs
 
     def _to_json(self, strip, to_serialize=None):
@@ -137,7 +141,7 @@ class ServiceAccountCredentials(AssertionCredentials):
             strip, to_serialize=to_serialize)
 
     @classmethod
-    def _from_parsed_json_keyfile(cls, keyfile_dict, scopes):
+    def _from_parsed_json_keyfile(cls, keyfile_dict, scopes, audience):
         """Helper for factory constructors from JSON keyfile.
 
         Args:
@@ -145,6 +149,8 @@ class ServiceAccountCredentials(AssertionCredentials):
                           containing the contents of the JSON keyfile.
             scopes: List or string, Scopes to use when acquiring an
                     access token.
[...]

Issue #211, opened 2015-07-01
Clean up Service Account support


dhermes:
See the discussion in #207. Though not directly related, it revealed a few things.

1. There should just be a single service account class with factory constructors to handle PKCS12/P12 vs. JSON keys
1. It seems that usable auth created a shadowed `service_account._ServiceAccountCredentials` class that does essentially the same thing that `client.SignedJwtAssertionCredentials` does
1. We should better document the difference between a PKCS12/P12 key and a JSON key (and hopefully there is somewhere on the Google official OAuth docs explaining the difference as well)

sr-murthy:
Hi,

would you like to me to do the change and make a pull request for this?

Sandeep

dhermes:
No I or some other maintainer can handle it. I just wanted to give a source for the issue. Thanks though!

dhermes:
I [played around][1] a bit and it seems like P12->JSON is possible but JSON->P12 is not.

The reason is that a P12 key is made up of a PEM certificate and a PEM key, which can be seen via:

```python
from OpenSSL import crypto
with open('keyfile.p12', 'rb') as fh:
    p12 = crypto.load_pkcs12(fh.read(), 'notasecret')
print crypto.dump_privatekey(crypto.FILETYPE_PEM, p12.get_privatekey())
print crypto.dump_certificate(crypto.FILETYPE_PEM, p12.get_certificate())
```

On the other hand, the JSON key file has a `private_key` field, but no corresponding certificate field. It may be that the `private_key_id` field corresponds to a certificate stored internally within Google (or something similar).

This manifests in the first section of the JWT sent (the header). For the P12 

```json
{
  "alg": "RS256", 
  "typ": "JWT"
}
```

vs. for a JSON key (with a `private_key_id`):

```json
{
  "alg": "RS256", 
  "kid": "DEADBEEF", 
  "typ": "JWT"
}
```

----

This may not matter if we never use the certificate from the P12 key.

[1]: https://gist.github.com/dhermes/5907b8e19b2321fce35b

dhermes:
As it turns out, JSON->P12 can be done as well (at least for getting an access token). It's as easy as 

1. Creating a PKCS12 object
1. Adding the JSON private key to the object
1. Exporting the object to a file with the given passphrase


```python
p12_obj = crypto.PKCS12()  # 1
p12_obj.set_privatekey(crypto.load_privatekey(  # 2
    crypto.FILETYPE_PEM, private_key))
[...]

PR #274, opened 2015-08-14
Add JWT access token support for service accounts

soltanmm:
This is similar to #161, but takes into account the intended use case of simply 'promoting' an assertion to using a self-issued JWT access token instead of [something completely different](https://www.youtube.com/watch?v=K2P86C-1x3o).

But, I'm not quite sure of the appropriate way to test this. There's some test code that covers it (apparently the test code generated the assertion even when the old `create_scoped_required` would have returned `True`), but nothing particularly functional.

It might negatively interact with the spirit of #211.

cc @nathanielmanistaatgoogle @anthmgoogle #252 #253

### DIFF
--- a/oauth2client/service_account.py
+++ b/oauth2client/service_account.py
@@ -61,31 +61,49 @@ class _ServiceAccountCredentials(AssertionCredentials):
   def _generate_assertion(self):
     """Generate the assertion that will be used in the request."""
 
+    now = int(time.time())
     header = {
         'alg': 'RS256',
         'typ': 'JWT',
         'kid': self._private_key_id
     }
-
-    now = int(time.time())
     payload = {
         'aud': self._token_uri,
-        'scope': self._scopes,
         'iat': now,
         'exp': now + _ServiceAccountCredentials.MAX_TOKEN_LIFETIME_SECS,
         'iss': self._service_account_email
     }
-    payload.update(self._kwargs)
+    if self._scopes:
+      # Use usual process to acquire token
+      payload.update({
+          'scope': self._scopes,
+      })
+      payload.update(self._kwargs)
 
-    first_segment = _urlsafe_b64encode(_json_encode(header))
-    second_segment = _urlsafe_b64encode(_json_encode(payload))
-    assertion_input = first_segment + b'.' + second_segment
+      first_segment = _urlsafe_b64encode(_json_encode(header))
+      second_segment = _urlsafe_b64encode(_json_encode(payload))
+      assertion_input = first_segment + b'.' + second_segment
 
-    # Sign the assertion.
-    rsa_bytes = rsa.pkcs1.sign(assertion_input, self._private_key, 'SHA-256')
-    signature = base64.urlsafe_b64encode(rsa_bytes).rstrip(b'=')
+      # Sign the assertion.
+      rsa_bytes = rsa.pkcs1.sign(assertion_input, self._private_key, 'SHA-256')
+      signature = base64.urlsafe_b64encode(rsa_bytes).rstrip(b'=')
 
-    return assertion_input + b'.' + signature
+      return assertion_input + b'.' + signature
+
+    else:
+      # Use self-signed JWT
+      payload.update({
+          'sub': self._service_account_email
+      })
+      header['Authentication'] = payload
+      assertion_input = (_urlsafe_b64encode(_json_encode(header)) + b'.' +
+                         _urlsafe_b64encode(_json_encode(self._kwargs)))
+
+      # Sign the assertion.
+      rsa_bytes = rsa.pkcs1.sign(assertion_input, self._private_key, 'SHA-256')
+      signature = base64.urlsafe_b64encode(rsa_bytes).rstrip(b'=')
+
+      return assertion_input + b'.' + signature
 
   def sign_blob(self, blob):
     # Ensure that it is bytes
@@ -108,7 +126,7 @@ class _ServiceAccountCredentials(AssertionCredentials):
     }
 
   def create_scoped_required(self):
-    return not self._scopes
+    return False
 
   def create_scoped(self, scopes):
     return _ServiceAccountCredentials(self._service_account_id,
--- a/tests/test_service_account.py
+++ b/tests/test_service_account.py
@@ -76,9 +76,6 @@ class ServiceAccountCredentialsTests(unittest.TestCase):
     self.assertEqual(self.service_account_email,
                      self.credentials.service_account_email)
 
-  def test_create_scoped_required_without_scopes(self):
-    self.assertTrue(self.credentials.create_scoped_required())
-
   def test_create_scoped_required_with_scopes(self):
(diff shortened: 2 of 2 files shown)


oauth2client/service_account.py
```diff
@@ -61,31 +61,50 @@ def __init__(self, service_account_id, service_account_email, private_key_id,
...
+      payload = {
+          'aud': self._token_uri,
+          'scope': self._scopes,
+          'iat': now,
+          'exp': now + _ServiceAccountCredentials.MAX_TOKEN_LIFETIME_SECS,
+          'iss': self._service_account_email
+      }
+      payload.update(self._kwargs)
+
+      first_segment = _urlsafe_b64encode(_json_encode(header))
```
anthmgoogle, line 83: Can more of this code be shared with the copy below.

oauth2client/service_account.py
```diff
@@ -107,9 +126,6 @@ def serialization_data(self):
         'private_key': self._private_key_pkcs8_text
     }
 
-  def create_scoped_required(self):
```
anthmgoogle, line 110: This should be left in for compatibility, although it can be changed to return false.

oauth2client/service_account.py
```diff
@@ -61,31 +61,50 @@ def __init__(self, service_account_id, service_account_email, private_key_id,
   def _generate_assertion(self):
```
anthmgoogle, line 61: This shares the code for the the Jwt creation. However, they are also used differently, as JWT Access goes in directly into the authorization headers as a bearer token, while scoped credentials fetch an access token and use that for authorization token. I was expecting to see some logic reflecting that.

anthmgoogle:
Finished review pass.

oauth2client/service_account.py
soltanmm, line 61: I was hoping someone could elucidate for me where that happens in the codebase as part of guidance for writing the tests...

oauth2client/service_account.py
```diff
@@ -61,31 +61,49 @@ def __init__(self, service_account_id, service_account_email, private_key_id,
...
-    first_segment = _urlsafe_b64encode(_json_encode(header))
-    second_segment = _urlsafe_b64encode(_json_encode(payload))
-    assertion_input = first_segment + b'.' + second_segment
-
-    # Sign the assertion.
-    rsa_bytes = rsa.pkcs1.sign(assertion_input, self._private_key, 'SHA-256')
-    signature = base64.urlsafe_b64encode(rsa_bytes).rstrip(b'=')
-
-    return assertion_input + b'.' + signature
+    if self._scopes:
```
dhermes, line 76: What follows is plenty of code added to a method that was already non-trivial. Use helper methods to make testing and comprehension easier.

oauth2client/service_account.py
```diff
@@ -61,31 +61,49 @@ def __init__(self, service_account_id, service_account_email, private_key_id,
...
+      second_segment = _urlsafe_b64encode(_json_encode(payload))
+      assertion_input = first_segment + b'.' + second_segment
+
+      # Sign the assertion.
+      rsa_bytes = rsa.pkcs1.sign(assertion_input, self._private_key, 'SHA-256')
+      signature = base64.urlsafe_b64encode(rsa_bytes).rstrip(b'=')
+
+      return assertion_input + b'.' + signature
+
+    else:
```
dhermes, line 93: So the idea is to enable the new behavior with the existing class. Any reason you don't want to define a sub-class to make the distinction clear?

tests/test_service_account.py
```diff
@@ -76,9 +76,6 @@ def test_service_account_email(self):
     self.assertEqual(self.service_account_email,
                      self.credentials.service_account_email)
 
-  def test_create_scoped_required_without_scopes(self):
-    self.assertTrue(self.credentials.create_scoped_required())
```
dhermes, line 80: Why is this deleted?

oauth2client/service_account.py
```diff
@@ -108,7 +126,7 @@ def serialization_data(self):
     }
 
   def create_scoped_required(self):
-    return not self._scopes
+    return False
```
dhermes, line 129: I'm unclear on the reason for this change.

dhermes:
New code paths with no supporting tests. Write the tests please.

Also the design questions is any important one. Why did you elect to do this with the existing class rather than a subclass?

oauth2client/service_account.py
soltanmm, line 93: Entirely arbitrary. Earlier I made an attempt to get a clear spec on what it was that needed to be done with respect to this effort, and got unclear responses from multiple sources. That previous attempt used a separate class, but when I was finally informed that the credentials ought to be transparently 'promoted', I scrapped the approach.

tests/test_service_account.py
soltanmm, line 80: Because `create_scoped_required` arguably became somewhat pointless, and was deleted, and when added back in I forgot to put the test back in (i.e. this in particular is an oversight). I'm still unclear why we'd want to keep that method if it'll always be false and is not meant to be overridden anywhere in the hierarchy (see @anthmgoogle's response in an outdated commit), however, given that I've been told we want an automatic and transparent upgrade of service account credentials in the absence of scopes.

soltanmm:
@dhermes You seem to assume that there was a design to begin with; I'm still trying to figure out what it is that is desired here (and frankly I don't have enough familiarity with the oauth2client library structure or its use to confidently figure out what it is that needs to be here for this feature). At this point I've sort of relegated it to a constraint problem that'll need to be figured out through the pains of circumlocution.

oauth2client/service_account.py
soltanmm, line 129: See @anthmgoogle's response in an old commit. I believe I'll have to avoid commit-overwriting for the duration of this process...

oauth2client/service_account.py
soltanmm, line 76: I'd understand if this suggestion were about separating the functionality such that said separate functionality may be tested (and I'd love to do that, given, y'know, an actual understanding of what it is that's supposed to be here), but as it stands I don't believe the inherent complexity of this method warrants yet more helper methods s.t. reading it requires jumping around moreso than already necessitated by the current structure.

soltanmm:
@dhermes Actually, can we talk offline? This warrants an offline discussion.

dhermes:
Sorry I forgot to peek at the outdated diffs, that was a fail by me.

I won't have time for a VC until probably Friday, but am happy to chat over email.

oauth2client/service_account.py
dhermes, line 93: OK

oauth2client/service_account.py
dhermes, line 76: The whole point of a helper is to encapsulate an action into something concrete, e.g. `_make_assertion_for_jwt` and `_use_jwt_as_assertion`. Then each of those concrete actions can be analyzed in isolation.

dhermes:
Got all caught up on the outdated diffs/comments. LMK when / how you want to chat off of GitHub.

PR #275, opened 2015-08-17
Adding travis and tox configuration to run the app engine tests

### LINKED ISSUE #270: test_appengine.py should get run on Travis
[Current][1] approach is:

```
commands = nosetests --ignore-files=test_appengine\.py {posargs}
```

----

We could just make a custom `tox` environ for that test only and run

```
nosetests test_appengine\.py {posargs}
```

and then on Travis use the caching feature to make sure we don't grab the GAE SDK too often. Though I'd image enough other people on Travis grab the SDK that they have it cached locally anyway. (We're already pulling it down to build `docs`.)

[1]: https://github.com/google/oauth2client/blob/891c8649d8a1f4f435e8007287df6313c036ecad/tox.ini#L18

theacodes:
Fixes #270.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,12 @@ env:
   - TOX_ENV=docs
   - TOX_ENV=system-tests
   - TOX_ENV=system-tests3
+  - TOX_ENV=gae
+  global:
+  - GAE_PYTHONPATH=${HOME}/.cache/google_appengine
+cache:
+  directories:
+  - ${HOME}/.cache
 install:
 - ./scripts/install.sh
 script:
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -25,3 +25,7 @@ if [[ "${TOX_ENV}" == "pypy" ]]; then
     pyenv install pypy-2.6.0
     pyenv global pypy-2.6.0
 fi
+
+if [[ "${TOX_ENV}" == "gae" && ! -d ${GAE_PYTHONPATH} ]]; then
+    python scripts/fetch_gae_sdk.py `dirname ${GAE_PYTHONPATH}`
+fi
--- a/tests/data/app.yaml
+++ b/tests/data/app.yaml
@@ -1,0 +1,10 @@
+# Dummy app.yaml to placate nosegae.
+application: oauth2client
+version: 1
+runtime: python27
+api_version: 1
+threadsafe: yes
+
+handlers:
+- url: /.*
+  script: null.app
--- a/tests/test_appengine.py
+++ b/tests/test_appengine.py
@@ -26,6 +26,7 @@ import datetime
 import httplib2
 import json
 import os
+import tempfile
 import time
 import unittest
 import urllib
@@ -58,6 +59,7 @@ from oauth2client.appengine import FlowProperty
 from oauth2client.appengine import OAuth2Decorator
 from oauth2client.appengine import OAuth2DecoratorFromClientSecrets
 from oauth2client.appengine import StorageByKeyName
+from oauth2client.client import _CLOUDSDK_CONFIG_ENV_VAR
 from oauth2client.client import AccessTokenRefreshError
 from oauth2client.client import Credentials
 from oauth2client.client import FlowExchangeError
@@ -240,8 +242,10 @@ class TestAppAssertionCredentials(unittest.TestCase):
     self.assertEqual(None, token.expires_in)
 
   def test_save_to_well_known_file(self):
+    os.environ[_CLOUDSDK_CONFIG_ENV_VAR] = tempfile.mkdtemp()
     credentials = AppAssertionCredentials([])
     self.assertRaises(NotImplementedError, save_to_well_known_file, credentials)
+    del os.environ[_CLOUDSDK_CONFIG_ENV_VAR]
 
 
 class TestFlowModel(db.Model):
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py26,py27,py33,py34,pypy,cover
+envlist = py26,py27,py33,py34,pypy,gae,cover
 
 [testenv]
 basedeps = keyring
@@ -50,6 +50,15 @@ commands = {toxinidir}/scripts/build-docs
 basepython = python2.6
 deps = {[testenv]basedeps}
        django>=1.5,<1.6
+
+[testenv:gae]
+basepython = python2.7
+deps = {[testenv]basedeps}
+       nosegae
+commands =
+    nosetests --with-gae --gae-application=tests/data --logging-level=INFO {posargs} tests/test_appengine.py
+setenv:
+    PYTHONPATH={env:GAE_PYTHONPATH:}
 
 [testenv:system-tests]
 basepython =


jonparrott:
(Going to have to modify this a few times until travis works, will ping when it's green)

dhermes:
We need to turn on caching so these installs don't take 50 seconds every time.

dhermes:
I've seen this before on my machine when trying to fool it.

```python
OSError: Config directory does not exist: /home/travis/.config/gcloud
```

jonparrott:
Was waiting to enable cache until I got it passing. The latest push should light up all green and should be ready to merge.

jonparrott:
Sigh. Not yet. Hold on.

dhermes:
w00t green!

Regarding #212, we'll never light up as "100%" coverage unless the GAE tests run side-by-side with the rest. This PR is a fine stop-gap for now, but probably worth just doing them all at once once the cache is turned on (and we can just pull GAE stuff from the cache).

I know right know that the GAE imports get angry when `django>1.5` is installed, but we could probably deal with that in the `cover` env.

jonparrott:
You can also trick coverage into compiling stats from multiple runs. We did this with the docs samples as well.

.travis.yml
```diff
@@ -3,14 +3,20 @@ python: 2.7
...
-  - TOX_ENV=system-tests
-  - TOX_ENV=system-tests3
+  # - TOX_ENV=py26
+  # - TOX_ENV=py27
+  # - TOX_ENV=py33
+  # - TOX_ENV=py34
+  # - TOX_ENV=pypy
+  # - TOX_ENV=docs
+  # - TOX_ENV=system-tests
+  # - TOX_ENV=system-tests3
```
dhermes: You're gonna want to uncomment then before the merge (I'm sure you're aware, just don't want to forget).

jonparrott:
Can you reset travis' cache? It seems I cached a bad install of the SDK. Once that's reset, we should have green tests.

.travis.yml
theacodes: ofc.

.travis.yml
```diff
@@ -3,14 +3,20 @@ python: 2.7
...
+  - TOX_ENV=gae
+  global:
+  - GAE_PYTHONPATH=$HOME/google_appengine
```
dhermes: I think it's considered good form to use `${HOME}` instead? New vs. old bash syntax or something. (I can find a link if you really want, I need to remember better than "I think it's considered..." anyhow.)

.travis.yml
```diff
@@ -3,14 +3,20 @@ python: 2.7
...
+cache:
+  directories:
+  - $HOME/google_appengine
```
dhermes: I think we should use `${HOME}/.cache/google_appengine`. It's what I've seen be used most commonly as the root directory (and gives a clear indication of purpose).

scripts/install.sh
```diff
@@ -25,3 +25,11 @@ if [[ "${TOX_ENV}" == "pypy" ]]; then
     pyenv install pypy-2.6.0
     pyenv global pypy-2.6.0
 fi
+
+if [[ "${TOX_ENV}" == "gae" && ! -d $HOME/google_appengine ]]; then
```
dhermes: Why aren't we using `${GAE_PYTHONPATH}`?

Aside: I am doing a bash code review. Weird!

.travis.yml
theacodes: I think either works, but if you find a source that says "use {} instead" I'll switch it. This is how I saw it in travis' docs, anyway.

scripts/install.sh
theacodes: Fair enough, let's use that instead. They were separate before because we use using gcloud.

scripts/install.sh
```diff
@@ -25,3 +25,11 @@ if [[ "${TOX_ENV}" == "pypy" ]]; then
...
+    cd $HOME
+    wget https://storage.googleapis.com/appengine-sdks/featured/google_appengine_1.9.25.zip
```
dhermes: We don't want the version to be hard-coded. Luckily @craigcitro wrote us a nice script: `scripts/fetch_gae_sdk.py` which takes an optional argument to specify where to download it to.

scripts/install.sh
```diff
@@ -25,3 +25,11 @@ if [[ "${TOX_ENV}" == "pypy" ]]; then
...
+    unzip -q google_appengine_1.9.25.zip
+    # Directory is required for the write to well known file test to pass.
+    mkdir -p $HOME/.config/gcloud
```
dhermes: I disagree. We should just monkey patch the `os.getenv` / `os.path.exists` that fails in the `gae` unit tests.

tests/data/app.yaml
```diff
@@ -0,0 +1,10 @@
+# Dummy app.yaml to placate nosegae.
```
dhermes, line 1: Dumb `nosegae`.

Any link to explanation?

.travis.yml
theacodes: Will do.

tests/data/app.yaml
theacodes, line 1: It parses the app.yaml to make sure the right versions of any declared libraries are used, I think. (at least, that's what ferrisnose did)

jonparrott:
Still trying to get travis' cache to play along.

tox.ini
```diff
@@ -51,6 +51,15 @@ basepython = python2.6
 deps = {[testenv]basedeps}
        django>=1.5,<1.6
 
+[testenv:gae]
```
dhermes, line 54: What happens when we run this locally?

theacodes, line 54: It works, so long as you have GAE_PYTHONPATH set.

.travis.yml
```diff
@@ -3,14 +3,20 @@ python: 2.7
...
+  # - TOX_ENV=py27
+  # - TOX_ENV=py33
+  # - TOX_ENV=py34
+  # - TOX_ENV=pypy
+  # - TOX_ENV=docs
+  # - TOX_ENV=system-tests
+  # - TOX_ENV=system-tests3
+  - TOX_ENV=gae
+  global:
+  - GAE_PYTHONPATH=$HOME/.cache/google_appengine
```
dhermes: RE: source for curlies, see [this answer](http://stackoverflow.com/a/8748880/1068170). In particular

> Doing this everywhere, instead of just in potentially ambiguous cases, can be considered good programming practice. This is both for consistency and to avoid surprises like `$foo_$bar.jpg`, where it's not visually obvious that the underscore becomes part of the variable name.

tests/data/app.yaml
dhermes, line 1: OK. SGTM.

tox.ini
dhermes, line 54: And if not?

**UPDATE**: I should note that the `docs` build rule also relies on the SDK and so you could re-use/share the SDK between the two.

.travis.yml
theacodes: will do.

dhermes:
RE: 

> You can also trick coverage into compiling stats from multiple runs. We did this with the docs samples as well.

So cool! Does this work when using matrix env in Travis?

jonparrott:
> So cool! Does this work when using matrix env in Travis?
I don't think so. Maybe?

GAE causes all kinds of testing problems.

dhermes:
I can clear it for you in the settings if need be (you might not have privileges, I might not either though): http://docs.travis-ci.com/user/caching/#Clearing-Caches

jonparrott:
I don't have rights to clear the cache.

dhermes:
Do you want me to try to clear it?

jonparrott:
Okay - confirmed the settings are correct with a new cache directory (``cache2``). If someone can kill the broken cache, then this should light up all green. :)

jonparrott:
> Do you want me to try to clear it?
Yes, please.

dhermes:
@jonparrott Just deleted the cache

jonparrott:
*Presses big red button*

scripts/install.sh
```diff
@@ -25,3 +25,12 @@ if [[ "${TOX_ENV}" == "pypy" ]]; then
     pyenv install pypy-2.6.0
     pyenv global pypy-2.6.0
 fi
+
+if [[ "${TOX_ENV}" == "gae" && ! -d ${GAE_PYTHONPATH} ]]; then
+    mkdir -p `dirname ${GAE_PYTHONPATH}`
+    cd `dirname ${GAE_PYTHONPATH}`
+    wget https://storage.googleapis.com/appengine-sdks/featured/google_appengine_1.9.25.zip
+    unzip -q google_appengine_1.9.25.zip
```
dhermes: Still need to use the `scripts/fetch_gae_sdk.py` script here

scripts/install.sh
```diff
@@ -25,3 +25,12 @@ if [[ "${TOX_ENV}" == "pypy" ]]; then
...
+    # Directory is required for the write to well known file test to pass.
+    mkdir -p $HOME/.config/gcloud
```
dhermes: ~~`mkdir -p ${GAE_PYTHONPATH}`?~~ NVM that

However, I think we should just rewrite the test to not rely on the local environ.

jonparrott:
``system-tests3`` failed for some weird unrelated reason. This should be good to go, but I can trigger another build to be sure.

scripts/install.sh
theacodes: Holy crap. I didn't even notice that. Is it used anywhere else?

dhermes:
No need to re-trigger, but there are at two comments un-addressed (about hard-coding the version and making the `~/.config/gcloud` dir)

scripts/install.sh
dhermes: `docs`

scripts/install.sh
theacodes: Can we leave that for another CL? We can mock os.exists or whatever in that test.

scripts/install.sh
theacodes: Travis' caching will make it where most tests still run with stale versions. :(

scripts/install.sh
```diff
@@ -25,3 +25,9 @@ if [[ "${TOX_ENV}" == "pypy" ]]; then
     pyenv install pypy-2.6.0
     pyenv global pypy-2.6.0
 fi
+
+if [[ "${TOX_ENV}" == "gae" && ! -d ${GAE_PYTHONPATH} ]]; then
+    python scripts/fetch_gae_sdk.py `dirname ${GAE_PYTHONPATH}`
```
dhermes, line 30: RE: Stale version, I agree. This is the same locally. I'll need to update `fetch_gae_sdk` to do a better job, but we can do that in another PR.

scripts/install.sh
```diff
@@ -25,3 +25,9 @@ if [[ "${TOX_ENV}" == "pypy" ]]; then
...
+    # Directory is required for the write to well known file test to pass.
+    mkdir -p ${HOME}/.config/gcloud
```
dhermes: RE: another PR/CL. Sure thing. File the issue?

jonparrott:
travis has forced my hand on the one test that depends on ~/.config/gcloud, I just pushed a change that *should* fix that.

tox.ini
theacodes, line 54: If not, it'll die with "dev_appserver.py" not found. I'm not sure if there's another way to surface this.

jonparrott:
Finally, green.

tests/test_appengine.py
```diff
@@ -240,8 +242,10 @@ def test_get_access_token(self):
     self.assertEqual(None, token.expires_in)
 
   def test_save_to_well_known_file(self):
+    os.environ[_CLOUDSDK_CONFIG_ENV_VAR] = tempfile.mkdtemp()
```
dhermes, line 245: Yay for this!

tox.ini
dhermes, line 54: I'll just handle it when updating `fetch_gae_sdk`. This may make people who run `tox` with no eviron angry, but that might be a total of 1 (me) right now and I'm not angry.

dhermes:
LGTM

PR #276, opened 2015-08-20
Makes all code PEP8 compliant.

### LINKED ISSUE #137: PEP8-conformity
Is there any reason why oauth2client is using two spaces for indentation instead of four, as recommended by [PEP 8](https://www.python.org/dev/peps/pep-0008/#indentation)?

Would people be opposed to switching to PEP8-friendly code?

dhermes:
Fixes #137.

Note that running

``` bash
pep8 $(git ls-files '*py')
```

still has some failures. These are caused by
- `test_appengine.py` since `dev_appserver.fix_sys_path()` happens before some imports
- `test_django_orm.py` since modifications to `django.conf.global_setting` happen before some imports

and the rest are

```
docs/conf.py:43:1: E402 module level import not at top of file
samples/googleappengine/call_compute_service_from_gae.py:26:47: E231 missing whitespace after ','
scripts/run_system_tests.py:25:5: E129 visually indented line with same indent as next logical line
setup.py:31:1: E402 module level import not at top of file
setup.py:47:1: E402 module level import not at top of file
```


nathanielmanistaatgoogle:
Thanks for the multiple well-formed commits. What do you make of Travis being displeased with this pull request?

dhermes:
`tox -e docs` failed, I never ran these tests with a fresh environment (e.g. `tox -e docs --recreate`) and more importantly never ran it at all since `tox` with no environment flag just [runs][1] `py26,py27,py33,py34,pypy,gae,cover`.

The error was:

```
Warning, treated as error:
/home/travis/build/google/oauth2client/oauth2client/client.py:docstring of oauth2client.client.OAuth2Credentials.set_store:5: ERROR: Unexpected indentation.
```

I just `git push --force`d with the needed change folded into `Docstring pass after pep8ify in oauth2client/`. I also removed (from history) an accidental `*.bak` file that was added in one commit and removed in another. (Easy-ish to do while `git cherry-pick`ing the commits back together.)

----

As for the coverage drop, I kind of just brush that off. It's probably because some long lines made the total line count go up, e.g.

```python
  # 2 spaces
  foo = datetime.datetime.utcnow() + datetime.timedelta(seconds=3600)
```

becoming

```python
    # 4 spaces
    foo = (datetime.datetime.utcnow() + 
           datetime.timedelta(seconds=3600))
```

[1]: https://github.com/google/oauth2client/blob/043e066e54d00bd3f8c5a90d2cb83292f30f8ede/tox.ini#L2

dhermes:
@nathanielmanistaatgoogle Travis passes now and there is still that tiny `-0.03%` drop in coverage because math (#212 would make this annoyance stop).

nathanielmanistaatgoogle:
Yeah, I agree about brushing off the coverage drop. Re-reviewing now...

nathanielmanistaatgoogle:
LGTM; anything else to be done here?

dhermes:
Nope. But I'd be wary of making a new release until we have 100% test coverage.

PR #278, opened 2015-08-23
Getting to 100% coverage for clientsecrets module.

dhermes:


oauth2client/clientsecrets.py
```diff
@@ -60,32 +60,39 @@
...
         raise InvalidClientSecretsError(
             _INVALID_FILE_FORMAT_MSG + ' '
             'Expected a JSON object with a single property for a "web" or '
             '"installed" application')
-    client_type = tuple(obj)[0]
+    client_type, = clientsecrets_dict.keys()
     if client_type not in VALID_CLIENT:
         raise InvalidClientSecretsError(
             'Unknown client type: %s.' % (client_type,))
-    client_info = obj[client_type]
```
nathanielmanistaatgoogle, line 88: Why not

client_info, = clientsecrets_dict.values()

to match what's done four lines above?

nathanielmanistaatgoogle:
Looks good to me; consider my one comment optional and merge at your leisure.

oauth2client/clientsecrets.py
dhermes, line 88: Was going for test coverage here rather than changing the existing (not so stellar) code. Happy to change if desired.

We actually check both and could just watch for a `ValueError`:

``` python
>>> (a, b), = {1: 2, 3: 4}.items()
ValueError: too many values to unpack
>>> (a, b), = {1: 2}.items()
>>> (a, b), = {}.items()
ValueError: need more than 0 values to unpack
```

dhermes:
I did as you asked and also dealt with the possibility that the input was not a dictionary.

nathanielmanistaatgoogle:
I like it; nice use of unpacking.

PR #279, opened 2015-08-24
100 percent coverage keyring storage

dhermes:
I feel like a wizard from the future when I correctly use https://docs.python.org/3/library/unittest.mock.html


tests/test_keyring_storage.py
```diff
@@ -0,0 +1,173 @@
...
+# limitations under the License.
+
+"""Tests for oauth2client.keyring_storage tests.
+
+Unit tests for oauth2client.keyring_storage.
+"""
+
+import datetime
+import keyring
+import thread
```
nathanielmanistaatgoogle: From https://docs.python.org/2/library/thread.html: "Note: The thread module has been renamed to _thread in Python 3. The 2to3 tool will automatically adapt imports when converting your sources to Python 3; however, you should consider using the high-level threading module instead.".

Travis seems to agree that this test does not work in Python 3.

dhermes: Aw man! I was sleepy and figured it would work fine on Travis! Boo.

It doesn't look like `thread` is in `six.moves` so maybe I'll just leave it out. The only use is in `test_constructor`:

``` python
self.assertTrue(isinstance(store._lock, thread.LockType))
```

nathanielmanistaatgoogle: Yeah, I noticed it was just in that one place. Let me know what you decide.

dhermes: Swapped it out for `import threading` and then just used

``` python
lock_type = type(threading.Lock())
```

PR #281, opened 2015-08-24
Getting to 100% coverage for xsrfutil module.

dhermes:


tests/test_xsrfutil.py
```diff
@@ -16,24 +16,208 @@
...
+        token_time = 123456789
+        # Make sure it isn't too old.
+        curr_time = token_time + xsrfutil.DEFAULT_TIMEOUT_SECS - 1
+
+        key = object()
+        user_id = object()
+        action_id = object()
+        token = base64.b64encode(_to_bytes(str(token_time)))
+        # It is encoded as b'MTIzNDU2Nzg5', which has length 12.
+        # b'MMMMMMMMMMMM'
```
nathanielmanistaatgoogle: This could be a comment, but it looks more like commented-out first-draft code?

dhermes: It could be interpreted as something that became `b'M' * 12` or as a comment expressing that the code is tasty.

Will squash it out ASAP.

nathanielmanistaatgoogle:
As the former tech lead of the Melange project I'm always amused when I look at the xsrfutil header block. :-P

dhermes:
I had never seen this module until just now. It seems scary.

PR #282, opened 2015-08-24
Increasing coverage for flask_util

theacodes:
Coverage is at 97%. Apparently I'm missing a few branches somehow, but I'm unsure how to find them.

### DIFF
--- a/oauth2client/flask_util.py
+++ b/oauth2client/flask_util.py
@@ -177,7 +177,7 @@ try:
     from flask import request
     from flask import session
     from flask import url_for
-except ImportError:
+except ImportError:  # pragma: NO COVER
     raise ImportError('The flask utilities require flask 0.9 or newer.')
 
 from oauth2client.client import FlowExchangeError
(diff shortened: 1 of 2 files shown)


dhermes:
To find the branches run:

```bash
$ tox -e cover
$ coverage html
$ open htmlcov/oauth2client_flask_util.html
```

and then view the branch misses as HTML in your browser.

Ping me after they are addressed?

jonparrott:
Thanks for the tip. 100% coverage.

oauth2client/flask_util.py
```diff
@@ -536,9 +536,7 @@ def locked_get(self):
             return None
 
         credentials = OAuth2Credentials.from_json(serialized)
-
-        if credentials:
-            credentials.set_store(self)
+        credentials.set_store(self)
```
dhermes, line 539: Oof. How did we miss that.

dhermes:
LGTM

@nathanielmanistaatgoogle I'll let you weigh in

PR #285, opened 2015-08-25
Removing old_run module.

dhermes:
Has been deprecated since 700f13f3db1a9d1e3617eafa204459ba3f97ebd0 which occurred on July 2, 2013.

@craigcitro Do you see any reason not to remove this module?

(I debated getting 100% coverage for it in #212 and this seemed like an easier path forward)


craigcitro:
I suspect a number of people still use it.

dhermes:
OK @nathanielmanistaatgoogle shall we

- Remove it anyway and instruct those people to upgrade to using `run_flow` (we could even provide a snippet that would show them how to use `run_flow` with `gflags` without having to provide the `old_run` module)
- Un-deprecate the `old_run` module
- Do nothing but make sure it gets 100% test coverage for #212

dhermes:
@craigcitro I assume you mean internal to Google yes?

craigcitro:
i'm sure there are some in google -- but actually, i suspect there are plenty outside. i wager some folks just silenced or ignore the deprecation warning and have been happy their code is working with no work on their part.

i don't see "we need 100% test coverage" as a good reason to potentially break them?

dhermes:
@craigcitro That isn't the reason for breaking them. It just came up because I was looking at the module and realized it's been deprecated for 2 years 1 month (this library is only 5 years 4 months old: ed1325284e6507e66f104ce462ea73f0184c2e12)

nathanielmanistaatgoogle:
I approve this commit and am fine with "we no longer wish to be responsible for this code and have given more than enough advanced warning" as a reason for its removal. @craigcitro would that work for you? If I were to offer "a developer who depends on more-than-two-years-deprecated code is already broken and merely doesn't yet know it" as a philosophical proposition, how would you receive it?

dhermes:
@nathanielmanistaatgoogle I'll let you pull the trigger on the merge if/when this discussion sorts itself out.

craigcitro:
i don't have strong feelings either way -- i was just answering danny's question about "safe to delete".

i'd be hard-pressed to believe any philosophical propositions about an auth library.

PR #290, opened 2015-08-26
Getting to 100% coverage for crypt module.

dhermes:
I also re-factored the module a bit since the code was opaque and there was too much going on in one method.

### DIFF
--- a/oauth2client/crypt.py
+++ b/oauth2client/crypt.py
@@ -137,16 +137,17 @@ def verify_signed_jwt_with_certs(jwt, certs, audience):
         AppIdentityError if any checks are failed.
     """
     jwt = _to_bytes(jwt)
-    segments = jwt.split(b'.')
 
-    if len(segments) != 3:
-        raise AppIdentityError('Wrong number of segments in token: %s' % jwt)
-    message_to_sign = segments[0] + b'.' + segments[1]
+    if jwt.count(b'.') != 2:
+        raise AppIdentityError(
+            'Wrong number of segments in token: %s' % (jwt,))
 
-    signature = _urlsafe_b64decode(segments[2])
+    header, payload, signature = jwt.split(b'.')
+    message_to_sign = header + b'.' + payload
+    signature = _urlsafe_b64decode(signature)
 
     # Parse token.
-    json_body = _urlsafe_b64decode(segments[1])
+    json_body = _urlsafe_b64decode(payload)
     try:
         parsed = json.loads(_from_bytes(json_body))
     except:
--- a/oauth2client/crypt.py
+++ b/oauth2client/crypt.py
@@ -147,44 +147,46 @@ def verify_signed_jwt_with_certs(jwt, certs, audience):
     signature = _urlsafe_b64decode(signature)
 
     # Parse token.
-    json_body = _urlsafe_b64decode(payload)
+    payload_bytes = _urlsafe_b64decode(payload)
     try:
-        parsed = json.loads(_from_bytes(json_body))
+        payload_dict = json.loads(_from_bytes(payload_bytes))
     except:
-        raise AppIdentityError('Can\'t parse token: %s' % json_body)
+        raise AppIdentityError('Can\'t parse token: %s' % (payload_bytes,))
 
     # Check signature.
     _verify_signature(message_to_sign, signature, certs)
 
     # Check creation timestamp.
-    iat = parsed.get('iat')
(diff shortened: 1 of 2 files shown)


dhermes:
@nathanielmanistaatgoogle FYI

nathanielmanistaatgoogle:
I intend to get to this but won't until the weekend.

dhermes:
No worries just wanted to make sure it was still on your radar.

nathanielmanistaatgoogle:
Just the one comment on the second of the seven commits.

oauth2client/crypt.py
```diff
@@ -95,7 +97,108 @@ def make_signed_jwt(signer, payload):
     return b'.'.join(segments)
 
 
-def verify_signed_jwt_with_certs(jwt, certs, audience):
+def _verify_signature(message, signature, certs):
```
dhermes, line 100: There is no really good reason to use a dictionary here, but that is the "native" format that the certs are given by Google: https://www.googleapis.com/oauth2/v1/certs

This URI is [specified](https://github.com/google/oauth2client/blob/5dda03224fbb408fbb3ac450733b997b525496c6/oauth2client/client.py#L67) in `client.py`.

---

**UPDATE**: I'm happy to change to whatever you like. We could even log which cert / key ID / kid matches the token. (This way we'd use the keys and values of the dict.)

nathanielmanistaatgoogle, line 100: I don't think from a function design point of view that we should be looking for reasons to justify taking more than is needed rather than just paring down the parameter set to only what the function needs? Especially for a private function.

Also AppIdentityError does seem like a rather slim class that includes only a fixed message and none of the actual values that led to its generation - would that benefit from being fleshed out a little?

dhermes, line 100: I'm OK with that. I'll change it so that the only caller (`verify_signed_jwt_with_certs`) just sends `certs.values()` instead of `certs`.

Unfortunately, since `verify_signed_jwt_with_certs` is public, we can't change the signature without breaking users (should happen at a `1.x -> 2.x` change, maybe we should pile up things we don't like for that day?).

As for `AppIdentityError` I'm happy to change whatever you'd like, but suggest we punt to another PR. I really dislike the way exceptions are used in this library so am happy to re-work.

oauth2client/crypt.py
```diff
@@ -95,7 +97,107 @@ def make_signed_jwt(signer, payload):
...
 
 
-def verify_signed_jwt_with_certs(jwt, certs, audience):
+def _verify_signature(message, signature, cert_list):
+    """Verifies signed content using a list of certificates.
+
+    Args:
+        message: string or bytes, The message to verify.
+        signature: string or bytes, The signature on the message.
+        cert_list: list, certificates in PEM format.
```
dhermes: @nathanielmanistaatgoogle FYI this is now `cert_list`.

As for the `AppIdentityError`, I suggested in the previous comment (now hidden as "outdated" by GitHub) that we punt on it and follow up after.

nathanielmanistaatgoogle: Why is this parameter called "cert_list" and not just "certs"? Can't it be any iterable of certificates in PEM format? Isn't it passed an iterable that is not a list in Python3, as the code is currently drafted?

nathanielmanistaatgoogle:
I would never request a change to a public function signature in a refactoring pull request. :-) At least not accidentally. Private functions are totally fair game.

oauth2client/crypt.py
dhermes: Will rename as `certs` and fold back into the "latest" commit in the PR.

dhermes:
@nathanielmanistaatgoogle pushed the rename.

It seems you don't really care about the `AppIdentityError` since you keep ignoring my replies about it :grinning:

nathanielmanistaatgoogle:
It's fine to punt on AppIdentityError to a later time when you address what you "dislike [about] the way exceptions are used in this library". :-)

The commit log message of 3f9b592ea307d0b397d could use correction.

dhermes:
Correction from list to iterable or something else?

nathanielmanistaatgoogle:
Yes, just "iterable" in the commit log message.

dhermes:
OK cool. Updated the code and the commit message.

I can wait on Travis to merge if it LGTY.

nathanielmanistaatgoogle:
Looks good; merge at will and thanks for the discussion.

PR #306, opened 2015-09-16
Adding no-cover pragmas to unittest.main().

dhermes:
This is so we get 100% line coverage in our tests. Also
adding unittest.main() lines to the files missing it:

```
test__helpers.py
test__pycrypto_crypt.py
test_crypt.py
test_devshell.py
test_gce.py
test_keyring_storage.py
test_service_account.py
test_util.py
```

### DIFF
--- a/tests/test__helpers.py
+++ b/tests/test__helpers.py
@@ -115,3 +115,7 @@ class Test__urlsafe_b64decode(unittest.TestCase):
         bad_string = b'+'
         self.assertRaises((TypeError, binascii.Error),
                           _urlsafe_b64decode, bad_string)
+
+
+if __name__ == '__main__':  # pragma: NO COVER
+    unittest.main()
--- a/tests/test__pycrypto_crypt.py
+++ b/tests/test__pycrypto_crypt.py
@@ -61,3 +61,7 @@ class TestPyCryptoVerifier(unittest.TestCase):
         public_key = public_key.decode('utf-8')
         verifier = PyCryptoVerifier.from_string(public_key, is_x509_cert=True)
         self.assertTrue(isinstance(verifier, PyCryptoVerifier))
+
+
+if __name__ == '__main__':  # pragma: NO COVER
+    unittest.main()
--- a/tests/test_appengine.py
+++ b/tests/test_appengine.py
@@ -880,5 +880,5 @@ class DecoratorXsrfProtectionTests(unittest.TestCase):
                           appengine._parse_state_value, state[1:], UserMock())
 
 
-if __name__ == '__main__':
+if __name__ == '__main__':  # pragma: NO COVER
     unittest.main()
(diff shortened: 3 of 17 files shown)


dhermes:
@nathanielmanistaatgoogle I'm unclear if `unittest.main()` is still needed inside Google or not. The fact that 8 of our test files don't have them indicates to me that it isn't a big deal.

If in fact they are required inside Google, we can just add a `pragma: NO COVER` to them.

nathanielmanistaatgoogle:
It's got nothing to do with whether or not they're used inside Google and it's got nothing to do with whether or not the tests pass without them: including them is the *supported*, *warranted* way to write a test ([2.7](https://docs.python.org/2/library/unittest.html), [3.5](https://docs.python.org/3/library/unittest.html)). What's the motivation for taking them out at all?

dhermes:
1. Motivation: these lines show up as uncovered in our coverage report since they never get run by `nose` (can just as easily fix by making them `pragma: NO COVER` lines, though you may want me to also add the `main` statement to the 8 test modules without them)
1. `nosetests` doesn't use them (and neither does `py.test`), it imports the modules and grabs cases of the subclasses of `unittest.TestCase`.

nathanielmanistaatgoogle:
How important is our coverage report for test modules? How important is it that every line of test code be executed?

dhermes:
- 100% line coverage should be a minimum bar (it would immediately alert us of the issue we've had happen twice where the executable bit gets set on `test_client.py`)
- If we don't look out for non-100% coverage of unit tests, we may have code that doesn't run, even though the committer intended it to (this is happening right now in the newly added `flask` tests)
- It also finds dead code (I obsoleted a helper class and forgot to remove it)
- `pragma: NO COVER` is already enabled by `.coveragerc` and already in use in this repo for Py2/Py3 special cases

nathanielmanistaatgoogle:
Keep the `if __name__ == '__main__': unittest.main()` lines and `pragma: NO COVER` them.

The `unittest` module defines a unit test to be something that includes them, so I'd much rather have them than have modules that are not unit tests but that happen to look like unit tests when looked at by `nose` and `py.test`. As @soltanmm has been discovering recently, all Python unit testing frameworks are terrible. We shouldn't alter our code away from what conforms to the standard library's specifications just because two frameworks seem to allow us to do so.

dhermes:
@nathanielmanistaatgoogle I updated with the no-pragma, it moves `test_devshell`, `test_tools` and `test_xsrfutil` into 100% covered.

----

As for your remark, I don't think I'd call these frameworks "terrible". Also, AFAIK a large majority (like 95+%) of projects are covered by `nose` and `py.test`. Maybe @tseaver can comment on this.

I seem to recall you saying something about doing what the community deems relevant (e.g. PEP8), and ISTM `nose` and `py.test` are what the community has deemed the relevant test runners (as well as `tox`).

Issue #309, opened 2015-09-21
OAuth2Credentials access token refresh failures swallow HTTP status


thobrla:
Refresh token exceptions raised by [OAuth2Credentials._do_refresh_request](https://github.com/google/oauth2client/blob/621c8e3904da0964901440f048717e19c656bdd2/oauth2client/client.py#L888) include only the error message and not the HTTP response code that cause the error.

This is fine if the [http_request](https://github.com/google/oauth2client/blob/621c8e3904da0964901440f048717e19c656bdd2/oauth2client/client.py#L867) supports automatic retrying of HTTP 5xx codes, but there are cases where retries cannot be handled at this layer.  For example, during session operations like resumable downloads and uploads, repeating the original request is not the correct retry behavior.

For errors that have HTTP status, I'd like to include the response code so that the caller can make the correct decision.  To that end, I propose creating a subclass of AccessTokenRefreshError that includes the HTTP status when one is available.

dhermes:
:+1: 

@thobrla, I'm hoping at some point to use a `__call__`-able `AuthorizedHttp` class instead of monkey patching `httplib2.Http` instances. This way library devs (like you and I) can extend this class and handle special cases just like this issue refers to.

PR #310, opened 2015-09-21
Add HttpAccessTokenRefreshError

thobrla:
This adds an exception class including the HTTP status for access
token refresh requests made over HTTP.

Fixes https://github.com/google/oauth2client/issues/309

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -134,6 +134,13 @@ class AccessTokenRefreshError(Error):
     """Error trying to refresh an expired access token."""
 
 
+class HttpAccessTokenRefreshError(AccessTokenRefreshError):
+    """Error (with HTTP status) trying to refresh an expired access token."""
+    def __init__(self, *args, **kwargs):
+        super(HttpAccessTokenRefreshError, self).__init__(*args)
+        self.status = kwargs.get('status')
+
+
 class TokenRevokeError(Error):
     """Error trying to revoke a token."""
 
@@ -830,7 +837,7 @@ class OAuth2Credentials(Credentials):
                           refresh request.
 
         Raises:
-            AccessTokenRefreshError: When the refresh fails.
+            HttpAccessTokenRefreshError: When the refresh fails.
         """
         if not self.store:
             self._do_refresh_request(http_request)
@@ -858,7 +865,7 @@ class OAuth2Credentials(Credentials):
                           refresh request.
 
         Raises:
-            AccessTokenRefreshError: When the refresh fails.
+            HttpAccessTokenRefreshError: When the refresh fails.
         """
         body = self._generate_refresh_request_body()
         headers = self._generate_refresh_request_headers()
@@ -898,7 +905,7 @@ class OAuth2Credentials(Credentials):
                         self.store.locked_put(self)
             except (TypeError, ValueError):
                 pass
-            raise AccessTokenRefreshError(error_msg)
+            raise HttpAccessTokenRefreshError(error_msg, status=resp.status)
 
     def _revoke(self, http_request):
         """Revokes this credential and deletes the stored copy (if it exists).
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -31,6 +31,7 @@ import unittest
 
 import mock
 import six
+from six.moves import http_client
 from six.moves import urllib
 
 from .http_mock import CacheMock
@@ -43,7 +44,7 @@ from oauth2client import client
 from oauth2client import util as oauth2client_util
 from oauth2client.client import AccessTokenCredentials
 from oauth2client.client import AccessTokenCredentialsError
-from oauth2client.client import AccessTokenRefreshError
+from oauth2client.client import HttpAccessTokenRefreshError
 from oauth2client.client import ADC_HELP_MSG
 from oauth2client.client import AssertionCredentials
 from oauth2client.client import AUTHORIZED_USER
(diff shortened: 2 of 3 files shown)


tests/test_client.py
```diff
@@ -695,9 +695,9 @@ def test_token_refresh_failure(self):
             http = self.credentials.authorize(http)
             try:
                 http.request('http://example.com')
-                self.fail('should raise AccessTokenRefreshError exception')
-            except AccessTokenRefreshError:
-                pass
+                self.fail('should raise HttpAccessTokenRefreshError exception')
+            except HttpAccessTokenRefreshError, e:
+                self.assertEqual(400, e.status)
```
nathanielmanistaatgoogle: http_client.BAD_REQUEST rather than the literal value 400, please.

oauth2client/client.py
```diff
@@ -134,6 +134,13 @@ class AccessTokenRefreshError(Error):
     """Error trying to refresh an expired access token."""
 
 
+class HttpAccessTokenRefreshError(AccessTokenRefreshError):
+    """Error (with HTTP status) trying to refresh an expired access token."""
+    def __init__(self, *args, **kwargs):
+        super(HttpAccessTokenRefreshError, self).__init__(*args)
```
nathanielmanistaatgoogle, line 140: What happens to keyword args that do not have "status" as their key? It looks like they get dropped on the floor?

thobrla, line 140: I'm not aware that the base exception class does anything with keyword arguments to begin with, so I don't think forwarding them to the base class is a problem.  If you disagree, I'm find with including them.

tests/test_client.py
thobrla: Done.

oauth2client/client.py
dhermes, line 140: It [inherits](https://github.com/google/oauth2client/blob/29347a91356a958f500b2c59ae3ed85387c282bb/oauth2client/client.py#L125-L134) via `AccessTokenRefreshError -> Error -> Exception` and neither subclass actually modifies the parent (beyond the name). I can confirm that `Exception` rejects `kwargs`:

``` python
>>> Exception(a=10)
TypeError: exceptions.Exception does not take keyword arguments
```

---

Unfortunately, if you want to pass `*args`, you use named `kwargs` so this is the only way to go.

dhermes:
Also `except HttpAccessTokenRefreshError, e:` is not acceptable syntax in Python 3. See Travis [failure][1]

[1]: https://travis-ci.org/google/oauth2client/jobs/81469436

tests/test_client.py
```diff
@@ -690,14 +691,15 @@ def test_token_refresh_failure(self):
...
+                 b'{"error":"access_denied"}'),
             ])
             http = self.credentials.authorize(http)
             try:
                 http.request('http://example.com')
-                self.fail('should raise AccessTokenRefreshError exception')
-            except AccessTokenRefreshError:
-                pass
+                self.fail('should raise HttpAccessTokenRefreshError exception')
+            except HttpAccessTokenRefreshError, e:
```
dhermes: Using `except HttpAccessTokenRefreshError as e:`

tests/test_client.py
```diff
@@ -690,14 +691,15 @@ def test_token_refresh_failure(self):
...
+                ({'status': http_client.BAD_REQUEST},
+                 b'{"error":"access_denied"}'),
             ])
             http = self.credentials.authorize(http)
             try:
                 http.request('http://example.com')
-                self.fail('should raise AccessTokenRefreshError exception')
-            except AccessTokenRefreshError:
-                pass
+                self.fail('should raise HttpAccessTokenRefreshError exception')
```
dhermes, line 700: While you're mucking with this code, mind switching from `self.fail` and catching the error to just using the modern approach: `self.assertRaises`?

thobrla, line 700: I don't see a clean way to use assertRaises and also check the status property of the exception.  Do you have a suggestion here?

dhermes, line 700: You're right, I forgot about that part.

dhermes:
All LGTM, but before merging I am going to parrot for @nathanielmanistaatgoogle 

He [usually asks][1] people

> Please squash commits with each round of code review back-and-forth; the successive stages of progress here shouldn't necessarily become a part of canonical project history.

He [also asks][2] that you

> Please rewrite your commit message to conform to http://chris.beams.io/posts/git-commit/#seven-rules

[1]: https://github.com/google/oauth2client/pull/185#issuecomment-103598795
[2]: https://github.com/google/oauth2client/pull/196#issuecomment-112982411

thobrla:
Yes, was waiting for LGTM to squash.  In theory, the initial commit message should follow those rules (the referenced issue explains the "why").

dhermes:
@thobrla How desperate are you for a release? We did `1.5.0` on 2015-09-03 and `1.5.1` on 2015-09-16 (previously the release schedule is much less frequest).

thobrla:
Not desperate.  I'll need it for gsutil 4.16 but that release does not yet have a date planned.  I can pin the git submodule to this commit until then.

dhermes:
OK. Sounds like we can be independent then?

Side note: Aren't you guys moving to Go for the CLI?

thobrla:
Not completely independent - I will need an oauth2client release prior to gsutil 4.16, but if one hasn't happened by the time that is ready, I will ask for it.

dhermes:
OK SGTM.

As for Golang?

PR #312, opened 2015-09-22
Isolate the Django ORM test from App Engine.

dhermes:
Since most of our tests run with `--ignore-files=test_appengine\.py`, the App Engine SDK is "never" available when this test runs in our set-up.

### DIFF
--- a/tests/test_django_orm.py
+++ b/tests/test_django_orm.py
@@ -26,13 +26,6 @@ import pickle
 import sys
 import unittest
 
-# Ensure that if app engine is available, we use the correct django from it
-try:
-    from google.appengine.dist import use_library
-    use_library('django', '1.5')
-except ImportError:
-    pass
-
 from oauth2client.client import Credentials
 from oauth2client.client import Flow
 


nathanielmanistaatgoogle:
I don't understand this - tell me more about it?

dhermes:
The `tox -e cover` environ essentially does as follows

```
nosetests ...(LOTS OF FLAGS)... --ignore-files=test_appengine\.py
nosetests ...(LOTS OF FLAGS)... --cover-package=oauth2client.appengine --gae-application=tests/data ...
coverage report --show-missing --fail-under 80
```

We intentionally isolate the App Engine tests because importing the SDK **really** messes with imports. (We also have `tox -e py27` and `tox -e gae` separate).

So in the coverage report, the only time `test_django_orm.py` is getting executed is in the first run through `nosetests`. As a result, the import of the GAE SDK never occurs.

----

Doing a `git blame` on this code, the `try` block was added in 02b5f3f369d9dceaf197eac540ce5797d1a07250 (September 2012).

Going back to that original review (https://codereview.appspot.com/6490081) it doesn't seem that the App Engine tests needed to be isolated from the others at that time.

nathanielmanistaatgoogle:
Why leave a note behind rather than just remove the try/import/except entirely and cleanly?

dhermes:
That's fine with me, I was just trying to leave some evidence of what may occur.

Shall I proceed and just remove the note?

nathanielmanistaatgoogle:
Yes I think so.

dhermes:
Note removed. PTAL.

PR #297, opened 2015-09-03
Adding tests to coverage report.

dhermes:
This reveals there are dead lines in many of our tests. Some are innocuous, like

https://github.com/google/oauth2client/blob/f15e80ee5a4ff07803113204f6b8eda9be65f758/tests/test_xsrfutil.py#L296
https://github.com/google/oauth2client/blob/f15e80ee5a4ff07803113204f6b8eda9be65f758/tests/test_tools.py#L31

while others are caused by using `self.fail()` statements which never occur:

https://github.com/google/oauth2client/blob/f15e80ee5a4ff07803113204f6b8eda9be65f758/tests/test_service_account.py#L66
https://github.com/google/oauth2client/blob/f15e80ee5a4ff07803113204f6b8eda9be65f758/tests/test_service_account.py#L72

---

At `HEAD` right now, `tox -e cover` gets line coverage on

```
oauth2client/_helpers
oauth2client/_openssl_crypt
oauth2client/_pycrypto_crypt
oauth2client/appengine
oauth2client/client
oauth2client/clientsecrets
oauth2client/crypt
oauth2client/devshell
oauth2client/django_orm
oauth2client/file
oauth2client/flask_util
oauth2client/gce
oauth2client/keyring_storage
oauth2client/locked_file
oauth2client/multistore_file
oauth2client/service_account
oauth2client/tools
oauth2client/util
oauth2client/xsrfutil
```

With this PR, the following files also get line coverage (this is the point of `--cover-tests` BTW):

```
tests/__init__
tests/http_mock
tests/test__helpers
tests/test__pycrypto_crypt
tests/test_appengine
tests/test_client
tests/test_clientsecrets
tests/test_crypt
tests/test_devshell
tests/test_django_orm
tests/test_file
tests/test_flask_util
tests/test_gce
tests/test_jwt
tests/test_keyring_storage
tests/test_service_account
tests/test_tools
tests/test_util
tests/test_xsrfutil
```

PR #313, opened 2015-09-22
Calling two() method in flask util test.

dhermes:
In the original code, the line never executes since it is in a method that is decorated with `oauth2.required` and the method's scopes don't match the credentials used (forcing a redirect). Thus, we call it again with credentials that actually match the scopes.

### DIFF
--- a/tests/test_flask_util.py
+++ b/tests/test_flask_util.py
@@ -407,6 +407,15 @@ class FlaskOAuth2Tests(unittest.TestCase):
             self.assertTrue('include_granted_scopes' in q)
             self.assertEqual(q['scope'][0], 'email one two three')
 
+        # Actually call two() without a redirect.
+        credentials2 = self._generate_credentials(scopes=['two', 'three'])
+        with self.app.test_client() as c:
+            with c.session_transaction() as session:
+                session['google_oauth2_credentials'] = credentials2.to_json()
+
+            rv = c.get('/two')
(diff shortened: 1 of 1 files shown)


dhermes:
@nathanielmanistaatgoogle After this, the only line coverage [missing in tests][1] are `self.fail()` lines.

Towards #212 inch-by-inch!

[1]: https://github.com/google/oauth2client/pull/297#issuecomment-142150899

dhermes:
@jonparrott FYI I figured this was the easiest fix.

In general I do my best to avoid `NO COVER` pragmas. If you have a better idea for how to get this line actually executed. LMK.

jonparrott:
You could make another request to `/two` with all the valid scopes, but it seems like overkill.

dhermes:
I'll let @nathanielmanistaatgoogle decide.

Now that you point that out, it seems it'd be as easy as:

```python
        credentials2 = self._generate_credentials(scopes=['two'])

        with self.app.test_client() as c:
            with c.session_transaction() as session:
                session['google_oauth2_credentials'] = credentials2.to_json()

            rv = c.get('/two')
            self.assertEqual(rv.status_code, httplib.OK)
```

nathanielmanistaatgoogle:
Let's go with the latter approach that is a handful of lines of more code and avoids the coverage suppression pragma.

dhermes:
@nathanielmanistaatgoogle Updated the commit (and the messaging on the PR).

Issue #142, opened 2015-03-07
Deserialization error (Invalid Padding) after django_orm.CredentialsField.to_python


qcaron:
Hi guys !

I am using google-api-python-client to deal with Google Analytics API. google-api-python-client uses oauth2client to deal with Google's OAuth 2 authentication tokens (which are saved in my Django PostgreSQL DB).

I have created tests which used to pass with Django==1.6.5 and google-api-python-client==1.2 but now fail. I run my test with a fixture containing a model using a oauth2client.django_orm.CredentialsField field. Here is my full stack trace generated when the test is run:

```python
Traceback (most recent call last):
  File "/Users/quentin/Python/virtualenvs/mtanalytics/lib/python2.7/site-packages/django/test/testcases.py", line 182, in __call__
    self._pre_setup()
  File "/Users/quentin/Python/virtualenvs/mtanalytics/lib/python2.7/site-packages/django/test/testcases.py", line 754, in _pre_setup
    self._fixture_setup()
  File "/Users/quentin/Python/virtualenvs/mtanalytics/lib/python2.7/site-packages/django/test/testcases.py", line 907, in _fixture_setup
    'skip_checks': True,
  File "/Users/quentin/Python/virtualenvs/mtanalytics/lib/python2.7/site-packages/django/core/management/__init__.py", line 115, in call_command
    return klass.execute(*args, **defaults)
  File "/Users/quentin/Python/virtualenvs/mtanalytics/lib/python2.7/site-packages/django/core/management/base.py", line 338, in execute
    output = self.handle(*args, **options)
  File "/Users/quentin/Python/virtualenvs/mtanalytics/lib/python2.7/site-packages/modeltranslation/management/commands/loaddata.py", line 50, in handle
    return super(Command, self).handle(*fixture_labels, **options)
  File "/Users/quentin/Python/virtualenvs/mtanalytics/lib/python2.7/site-packages/django/core/management/commands/loaddata.py", line 61, in handle
    self.loaddata(fixture_labels)
  File "/Users/quentin/Python/virtualenvs/mtanalytics/lib/python2.7/site-packages/django/core/management/commands/loaddata.py", line 91, in loaddata
    self.load_label(fixture_label)
  File "/Users/quentin/Python/virtualenvs/mtanalytics/lib/python2.7/site-packages/django/core/management/commands/loaddata.py", line 142, in load_label
    for obj in objects:
[...]

PR #316, opened 2015-10-01
Issue #142: CredentialField Python 3 fix

apragacz:
This PR should fix issue #142.

Changes
- in `oauth2client.django_orm.CredentialsField`:
  - Python 2/3 metaclass support 
  - using `get_prep_value` instead of `get_db_prep_value`
  - added `value_to_string` method 
  - explicit conversions to text/bytes types in `get_prep_value`, `to_python` (required by Python 3)
- in `test_django_orm` tests:
  - added explicit conversion to bytes so the tests can work on Python 3

### DIFF
--- a/oauth2client/django_orm.py
+++ b/oauth2client/django_orm.py
@@ -21,17 +21,17 @@ the Django datastore.
 import oauth2client
 import base64
 import pickle
+import six
 
 from django.db import models
+from django.utils.encoding import smart_bytes, smart_text
 from oauth2client.client import Storage as BaseStorage
 
 
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 
 
-class CredentialsField(models.Field):
-
-    __metaclass__ = models.SubfieldBase
+class CredentialsField(six.with_metaclass(models.SubfieldBase, models.Field)):
 
     def __init__(self, *args, **kwargs):
         if 'null' not in kwargs:
@@ -46,12 +46,24 @@ class CredentialsField(models.Field):
             return None
         if isinstance(value, oauth2client.client.Credentials):
             return value
-        return pickle.loads(base64.b64decode(value))
+        return pickle.loads(base64.b64decode(smart_bytes(value)))
 
-    def get_db_prep_value(self, value, connection, prepared=False):
+    def get_prep_value(self, value):
         if value is None:
             return None
-        return base64.b64encode(pickle.dumps(value))
+        return smart_text(base64.b64encode(pickle.dumps(value)))
+
+    def value_to_string(self, obj):
+        """Convert the field value from the provided model to a string.
+        Used during model serialization.
+
+        Args:
+            obj: db.Model, model object
+        Returns:
+            string, the serialized field value
+        """
+        value = self._get_val_from_obj(obj)
+        return self.get_prep_value(value)
 
 
 class FlowField(models.Field):
--- a/tests/test_django_orm.py
+++ b/tests/test_django_orm.py
@@ -36,10 +36,11 @@ os.environ['DJANGO_SETTINGS_MODULE'] = 'django_settings'
 sys.modules['django_settings'] = django_settings = imp.new_module(
     'django_settings')
 django_settings.SECRET_KEY = 'xyzzy'
+from django.db import models
 
 from oauth2client.django_orm import CredentialsField
 from oauth2client.django_orm import FlowField
-
+from oauth2client._helpers import _from_bytes, _to_bytes
 
 __author__ = 'conleyo@google.com (Conley Owens)'
 
@@ -58,10 +59,39 @@ class TestCredentialsField(unittest.TestCase):
         self.assertTrue(isinstance(self.field.to_python(self.pickle),
                                    Credentials))
 
+    def test_field_unpickled_none(self):
+        self.assertEqual(self.field.to_python(None), None)
+
     def test_field_pickled(self):
         prep_value = self.field.get_db_prep_value(self.credentials,
                                                   connection=None)
-        self.assertEqual(prep_value, self.pickle)
+        self.assertEqual(_to_bytes(prep_value), self.pickle)
+
+
+class TestCredentialsFieldViaModel(unittest.TestCase):
+
+    class TestModel(models.Model):
+        credentials = CredentialsField()
+
+    def setUp(self):
+        self.model = self.TestModel()
+        # using the meta api:
+        # https://docs.djangoproject.com/en/1.8/ref/models/meta/#field-access-api
+        self.field = self.model._meta.get_field('credentials')
+        self.credentials = Credentials()
+        self.pickle_str = _from_bytes(base64.b64encode(pickle.dumps(
+            self.credentials
+        )))
+
+    def test_field_value_to_string(self):
+        self.model.credentials = self.credentials
+        value_str = self.field.value_to_string(self.model)
+        self.assertEqual(value_str, self.pickle_str)
+
+    def test_field_value_to_string_none(self):
+        self.model.credentials = None
+        value_str = self.field.value_to_string(self.model)
+        self.assertEqual(value_str, None)
 
 
 class TestFlowField(unittest.TestCase):


szopu:
I signed it!

dhermes:
@nathanielmanistaatgoogle Here are the docs for `smart_bytes` and `smart_text`:

https://docs.djangoproject.com/en/1.8/ref/utils/#django.utils.encoding.smart_bytes (behaves like our `_helpers._to_bytes`)
https://docs.djangoproject.com/en/1.8/ref/utils/#django.utils.encoding.smart_text (behaves like our `_helpers._from_bytes`)

dhermes:
Also worth noting this fixes the Python3 issues with metaclasses, previously attempted in #227 and it looks like this fixes #168 (in addition to #142, where the conversation with @szopu started).

@nathanielmanistaatgoogle Should we also have him fix `class FlowField(models.Field)` to work in Python3.

oauth2client/django_orm.py
```diff
@@ -46,12 +46,16 @@ def to_python(self, value):
             return None
         if isinstance(value, oauth2client.client.Credentials):
             return value
-        return pickle.loads(base64.b64decode(value))
+        return pickle.loads(base64.b64decode(smart_bytes(value)))
 
-    def get_db_prep_value(self, value, connection, prepared=False):
```
dhermes, line 51: @szopu I'm not a Django user, can you explain why you got rid of `get_db_prep_value` and swapped out for `get_prep_value` and `value_to_string`?

tests/test_django_orm.py
```diff
@@ -61,7 +61,7 @@ def test_field_unpickled(self):
     def test_field_pickled(self):
         prep_value = self.field.get_db_prep_value(self.credentials,
                                                   connection=None)
-        self.assertEqual(prep_value, self.pickle)
+        self.assertEqual(_to_bytes(prep_value), self.pickle)
```
dhermes, line 68: It seems `get_db_prep_value` is still defined. Is this because it relies on the methods we have implemented?

oauth2client/django_orm.py
```diff
@@ -46,12 +46,16 @@ def to_python(self, value):
...
+    def get_prep_value(self, value):
         if value is None:
             return None
-        return base64.b64encode(pickle.dumps(value))
+        return smart_text(base64.b64encode(pickle.dumps(value)))
+
+    def value_to_string(self, obj):
```
dhermes, line 56: Sorry to be nit-picky, can you add a docstring? (I realize the original author of this module wasn't as nice. I'm happy to help write it if you're having trouble.)

dhermes, line 56: Also, it seems this code is not tested. Can you add a test? (Again, I can help if you're having issues.)

oauth2client/django_orm.py
apragacz, line 51: @dhermes I think that @qcaron mentioned that properly implemented `value_to_string` method was required to perform proper serialization in newer Django versions (the default one will dump something like default object representation, i.e. "<oauth2client.client.OAuth2Credentials..."). Also `get_prep_value` is more general, so:
- it can be used in `value_to_string`
- `get_db_prep_value` calls it by default.

tests/test_django_orm.py
apragacz, line 68: @dhermes Yes, exactly.

oauth2client/django_orm.py
apragacz, line 56: @dhermes I can add a docstring, no problem. But testing `value_to_string` can be more tricky because I may need to create additional model class (and after that, an object) with this kind of field.
I could add additional test case which would test how the simple model object serializes to JSON for example (which should use `value_to_string`). What do you think about this idea?

BTW: Is it better to add additional commits to the PR or amend current commit with the changes?

oauth2client/django_orm.py
dhermes, line 51: Can you point to any docs?

apragacz, line 51: @dhermes sure:
- the relationship between `get_db_prep_value` and `get_prep_value`:
  https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.Field.get_db_prep_value
- the relationship between `value_to_string` and serialization: https://docs.djangoproject.com/en/1.8/howto/custom-model-fields/#converting-field-data-for-serialization

dhermes, line 51: Thanks!

szopu:
@dhermes Done. Added the docstring & additonal tests. I decided that low-level testing of `value_to_string` should be enough.

dhermes:
I wanted to make sure we don't introduce newly untested code paths. Before this PR, the untested code paths are


https://github.com/google/oauth2client/blob/af6d3eb8067d5c95fb59af6fb62a192e6c7fa16b/oauth2client/django_orm.py#L46
https://github.com/google/oauth2client/blob/af6d3eb8067d5c95fb59af6fb62a192e6c7fa16b/oauth2client/django_orm.py#L48
https://github.com/google/oauth2client/blob/af6d3eb8067d5c95fb59af6fb62a192e6c7fa16b/oauth2client/django_orm.py#L53
https://github.com/google/oauth2client/blob/af6d3eb8067d5c95fb59af6fb62a192e6c7fa16b/oauth2client/django_orm.py#L71
https://github.com/google/oauth2client/blob/af6d3eb8067d5c95fb59af6fb62a192e6c7fa16b/oauth2client/django_orm.py#L73
https://github.com/google/oauth2client/blob/af6d3eb8067d5c95fb59af6fb62a192e6c7fa16b/oauth2client/django_orm.py#L78
https://github.com/google/oauth2client/blob/af6d3eb8067d5c95fb59af6fb62a192e6c7fa16b/oauth2client/django_orm.py#L101-L104
https://github.com/google/oauth2client/blob/af6d3eb8067d5c95fb59af6fb62a192e6c7fa16b/oauth2client/django_orm.py#L112-L120
https://github.com/google/oauth2client/blob/af6d3eb8067d5c95fb59af6fb62a192e6c7fa16b/oauth2client/django_orm.py#L131-L140
https://github.com/google/oauth2client/blob/af6d3eb8067d5c95fb59af6fb62a192e6c7fa16b/oauth2client/django_orm.py#L145-L146

Now they are

https://github.com/szopu/oauth2client/blob/b6bb23f0559c48bfb229de09142f01ee612ae3d4/oauth2client/django_orm.py#L83
https://github.com/szopu/oauth2client/blob/b6bb23f0559c48bfb229de09142f01ee612ae3d4/oauth2client/django_orm.py#L85
https://github.com/szopu/oauth2client/blob/b6bb23f0559c48bfb229de09142f01ee612ae3d4/oauth2client/django_orm.py#L90
https://github.com/szopu/oauth2client/blob/b6bb23f0559c48bfb229de09142f01ee612ae3d4/oauth2client/django_orm.py#L113-L116
https://github.com/szopu/oauth2client/blob/b6bb23f0559c48bfb229de09142f01ee612ae3d4/oauth2client/django_orm.py#L124-L132
https://github.com/szopu/oauth2client/blob/b6bb23f0559c48bfb229de09142f01ee612ae3d4/oauth2client/django_orm.py#L143-L152
https://github.com/szopu/oauth2client/blob/b6bb23f0559c48bfb229de09142f01ee612ae3d4/oauth2client/django_orm.py#L157-L158

oauth2client/django_orm.py
dhermes, line 56: RE: "BTW: Is it better to add additional commits to the PR or amend current commit with the changes?"

@nathanielmanistaatgoogle Usually tells people to make new commits during review and then squash into a single commit before doing the merge

tests/test_django_orm.py
```diff
@@ -58,10 +59,39 @@ def test_field_unpickled(self):
...
+        self.assertEqual(self.field.to_python(None), None)
+
     def test_field_pickled(self):
         prep_value = self.field.get_db_prep_value(self.credentials,
                                                   connection=None)
-        self.assertEqual(prep_value, self.pickle)
+        self.assertEqual(_to_bytes(prep_value), self.pickle)
+
+
+class TestCredentialsFieldViaModel(unittest.TestCase):
```
dhermes, line 71: :+1:

tests/test_django_orm.py
```diff
@@ -58,10 +59,39 @@ def test_field_unpickled(self):
...
+
+    class TestModel(models.Model):
+        credentials = CredentialsField()
+
+    def setUp(self):
+        self.model = self.TestModel()
+        # using the meta api:
+        # https://docs.djangoproject.com/en/1.8/ref/models/meta/#field-access-api
```
dhermes, line 79: What are the implications of using this? Is it required to be in version `>=1.8`? Is there some minimum version we should keep around for Django compatibility?

apragacz, line 79: the `get_field` method (and the `_meta` API) has existed for a long time, but wasn't documented until now. Not sure when it was added, but it definitely can be found in Django `1.3.*`, which is pretty old version (more than 2 years old).

dhermes, line 79: Great thanks!

dhermes:
This LGTM

@nathanielmanistaatgoogle are we a go to merge this?

----

After merging, someone (I can do it) needs to update `django_orm.FlowField` to also be Python3 compatible (modeled after this PR) and then add tests for it (that PR might as well also add tests for `django_orm.Storage`).

nathanielmanistaatgoogle:
Looks mergeable to me; thank you for leading review!

dhermes:
Thanks @szopu !

Issue #319, opened 2015-10-09
[FR] Useful storage for web applications.


jonparrott:
Currently, we have the following storage classes:

* ``appengine.StorageByKeyName``, which is of course particular to GAE and Datastore.
* ``django_orm.Storage``, which appears to be a port of ``StorageByKeyName`` to Django's ORM.
* ``flask_util.FlaskSessionStorage``.
* ``file.Storage``, ``keyring_storage.Storage``, and ``multistore_file.Storage`` which are not useful in a web application context.

We really don't have a good set of general-purpose storage classes for web applications. I propose a set of new storage classes to make usage of the ``flask_util`` and incoming``django_util`` and ``bottle_util`` easier. Myself and @waprin are willing to implement these, we just want to make sure we're on the right track and that these will be welcome additions.

/cc @dhermes @anthmgoogle

### `storage.DictionaryStorage`

Similar to `StorageByKeyName`, but will write to any object with a dictionary-like interface. This will be used to implement storage for Flask, Django, and other web frameworks' ``session`` object easily.

```python
storage = DictionaryStorage(flask.session, key='google_oauth2_credentials') 
```

### `storage.KeyValueStorage`

Again, similar to `StorageByKeyName`, but will write to any object with ``get(key)``, ``set(key, value)`` and, ``delete(key)`` methods. This will be used to provide storage for redis, memcache, and other kv-stores.

```python
r = redis.StrictRedis()
key = "google_oauth2_credentials_{}".format(flask.session['user_id'])
storage = KeyValueStorage(r, key=key)
```

### App Engine changes

We'll create two new storage classes ``NdbStorage`` and ``DbStorage`` that will default the model to the existing ``CredentialsNDBModel`` and ``CredentialsModel``

Finally, `OAuth2Decorator` will be modified to accept a `storage` argument and only use the existing `_storage_class`, `_credentials_class `, and `_credentials_property_name` for backwards-compatibility.

### `storage.CloudDatastoreStorage`

Uses ``gcloud-python`` to store the credentials, similar usage to ``KeyValueStorage``.

```python
ds = gcloud.datastore.Client()
key = flask.session['user_id']
storage = CloudDatastoreStorage(ds, key=key)
```

### `storage.SqlAlchemyStorage`

Uses sqlalchemy to store the credentials. I haven't given this much thought, but it'll likely be similar to the Django ORM storage.

## Implementation / usage suggestions

[...]

PR #322, opened 2015-10-12
Fixing incremental auth in flask_util.

### LINKED ISSUE #320: flask_util: Custom Scopes Causes Redirect Loops
I am experiencing redirect loops when passing additional scopes to the decorator that is not part of the `DEFAULT_SCOPES`. You can see a sample that reproduces it below.

I think this is because the flow is recreated in the callback but with just the default scopes, which causes the credentials saved to not have the additional scopes, which means the decorator's predicate is never satisfied.

@jonparrott 

```python
from oauth2client.flask_util import UserOAuth2

app.config['SECRET_KEY'] = 'your-secret-key'
app.config['GOOGLE_OAUTH2_CLIENT_ID'] = 'client-id'
app.config['GOOGLE_OAUTH2_CLIENT_SECRET'] = 'client-secret'
oauth2 = UserOAuth2(app)


@app.route('/hello/')
@oauth2.required
def hello():
    return 'Hello World'

@app.route('/calendar')
@oauth2.required(scopes=['https://www.googleapis.com/auth/calendar'])
def calendar():
    return 'got calendar'
```

theacodes:
- Flow is now stored in the session, ensuring that the scopes survive the round trip to the auth server.
- The base credentials are now checked in `required`, solving an issue where it's possible to pass the check with the incremental scopes but without the base scopes.

Fixes #320

### DIFF
--- a/oauth2client/flask_util.py
+++ b/oauth2client/flask_util.py
@@ -23,7 +23,7 @@ available.
 Configuration
 =============
 
-To configure, you'll need a set of OAuth2 client ID from the
+To configure, you'll need a set of OAuth2 web application credentials from the
 `Google Developer's Console <https://console.developers.google.com/project/_/\
 apiui/credential>`__.
 
@@ -164,6 +164,7 @@ available outside of a request context, you will need to implement your own
 import hashlib
 import json
 import os
+import pickle
 from functools import wraps
 
 import six.moves.http_client as httplib
@@ -185,12 +186,26 @@ from oauth2client.client import OAuth2Credentials
 from oauth2client.client import OAuth2WebServerFlow
 from oauth2client.client import Storage
 from oauth2client import clientsecrets
-from oauth2client import util
 
 
 __author__ = 'jonwayne@google.com (Jon Wayne Parrott)'
 
-DEFAULT_SCOPES = ('email',)
+_DEFAULT_SCOPES = ('email',)
+_CREDENTIALS_KEY = 'google_oauth2_credentials'
+_FLOW_KEY = 'google_oauth2_flow_{0}'
+_CSRF_KEY = 'google_oauth2_csrf_token'
+
+
+def _get_flow_for_token(csrf_token):
+    """Retrieves the flow instance associated with a given CSRF token from
+    the Flask session."""
+    flow_pickle = session.get(
+        _FLOW_KEY.format(csrf_token), None)
+
+    if flow_pickle is None:
+        return None
+    else:
+        return pickle.loads(flow_pickle)
 
 
 class UserOAuth2(object):
@@ -250,7 +265,7 @@ class UserOAuth2(object):
         self.storage = storage
 
         if scopes is None:
-            scopes = app.config.get('GOOGLE_OAUTH2_SCOPES', DEFAULT_SCOPES)
+            scopes = app.config.get('GOOGLE_OAUTH2_SCOPES', _DEFAULT_SCOPES)
         self.scopes = scopes
 
         self._load_config(client_secrets_file, client_id, client_secret)
@@ -300,7 +315,8 @@ class UserOAuth2(object):
         client_type, client_info = clientsecrets.loadfile(filename)
         if client_type != clientsecrets.TYPE_WEB:
             raise ValueError(
-                'The flow specified in %s is not supported.' % client_type)
+                'The flow specified in {0} is not supported.'.format(
+                    client_type))
 
         self.client_id = client_info['client_id']
         self.client_secret = client_info['client_secret']
@@ -310,7 +326,7 @@ class UserOAuth2(object):
         # Generate a CSRF token to prevent malicious requests.
         csrf_token = hashlib.sha256(os.urandom(1024)).hexdigest()
 
-        session['google_oauth2_csrf_token'] = csrf_token
+        session[_CSRF_KEY] = csrf_token
 
         state = json.dumps({
             'csrf_token': csrf_token,
@@ -320,16 +336,21 @@ class UserOAuth2(object):
         kw = self.flow_kwargs.copy()
         kw.update(kwargs)
 
-        extra_scopes = util.scopes_to_string(kw.pop('scopes', ''))
-        scopes = ' '.join([util.scopes_to_string(self.scopes), extra_scopes])
+        extra_scopes = kw.pop('scopes', [])
+        scopes = set(self.scopes).union(set(extra_scopes))
 
-        return OAuth2WebServerFlow(
+        flow = OAuth2WebServerFlow(
             client_id=self.client_id,
             client_secret=self.client_secret,
             scope=scopes,
             state=state,
             redirect_uri=url_for('oauth2.callback', _external=True),
             **kw)
+
+        flow_key = _FLOW_KEY.format(csrf_token)
+        session[flow_key] = pickle.dumps(flow)
+
+        return flow
 
     def _create_blueprint(self):
         bp = Blueprint('oauth2', __name__)
@@ -367,11 +388,12 @@ class UserOAuth2(object):
         if 'error' in request.args:
             reason = request.args.get(
                 'error_description', request.args.get('error', ''))
-            return 'Authorization failed: %s' % reason, httplib.BAD_REQUEST
+            return ('Authorization failed: {0}'.format(reason),
+                    httplib.BAD_REQUEST)
 
         try:
             encoded_state = request.args['state']
-            server_csrf = session['google_oauth2_csrf_token']
+            server_csrf = session[_CSRF_KEY]
             code = request.args['code']
         except KeyError:
             return 'Invalid request', httplib.BAD_REQUEST
@@ -386,14 +408,17 @@ class UserOAuth2(object):
         if client_csrf != server_csrf:
             return 'Invalid request state', httplib.BAD_REQUEST
 
-        flow = self._make_flow()
+        flow = _get_flow_for_token(server_csrf)
+
+        if flow is None:
+            return 'Invalid request state', httplib.BAD_REQUEST
 
         # Exchange the auth code for credentials.
         try:
             credentials = flow.step2_exchange(code)
         except FlowExchangeError as exchange_error:
             current_app.logger.exception(exchange_error)
-            content = 'An error occurred: %s' % (exchange_error,)
+            content = 'An error occurred: {0}'.format(exchange_error)
             return content, httplib.BAD_REQUEST
 
         # Save the credentials to the storage.
@@ -409,7 +434,7 @@ class UserOAuth2(object):
         """The credentials for the current user or None if unavailable."""
         ctx = _app_ctx_stack.top
 
-        if not hasattr(ctx, 'google_oauth2_credentials'):
+        if not hasattr(ctx, _CREDENTIALS_KEY):
             ctx.google_oauth2_credentials = self.storage.get()
 
         return ctx.google_oauth2_credentials
@@ -432,7 +457,7 @@ class UserOAuth2(object):
             return self.credentials.id_token['email']
         except KeyError:
             current_app.logger.error(
-                'Invalid id_token %s', self.credentials.id_token)
+                'Invalid id_token {0}'.format(self.credentials.id_token))
 
     @property
     def user_id(self):
@@ -448,7 +473,7 @@ class UserOAuth2(object):
             return self.credentials.id_token['sub']
         except KeyError:
             current_app.logger.error(
-                'Invalid id_token %s', self.credentials.id_token)
+                'Invalid id_token {0}'.format(self.credentials.id_token))
 
     def authorize_url(self, return_url, **kwargs):
         """Creates a URL that can be used to start the authorization flow.
@@ -473,27 +498,29 @@ class UserOAuth2(object):
         def curry_wrapper(wrapped_function):
             @wraps(wrapped_function)
             def required_wrapper(*args, **kwargs):
-
                 return_url = decorator_kwargs.pop('return_url', request.url)
 
-                # No credentials, redirect for new authorization.
-                if not self.has_credentials():
+                requested_scopes = set(self.scopes)
+                if scopes is not None:
+                    requested_scopes |= set(scopes)
+                if self.has_credentials():
+                    requested_scopes |= self.credentials.scopes
+
+                requested_scopes = list(requested_scopes)
+
+                # Does the user have credentials and does the credentials have
+                # all of the needed scopes?
+                if (self.has_credentials() and
+                        self.credentials.has_scopes(requested_scopes)):
+                    return wrapped_function(*args, **kwargs)
+                # Otherwise, redirect to authorization
+                else:
                     auth_url = self.authorize_url(
                         return_url,
-                        scopes=scopes,
+                        scopes=requested_scopes,
                         **decorator_kwargs)
-                    return redirect(auth_url)
 
-                # Existing credentials but mismatching scopes, redirect for
-                # incremental authorization.
-                if scopes and not self.credentials.has_scopes(scopes):
-                    auth_url = self.authorize_url(
-                        return_url,
-                        scopes=list(self.credentials.scopes) + scopes,
-                        **decorator_kwargs)
                     return redirect(auth_url)
-
-                return wrapped_function(*args, **kwargs)
 
             return required_wrapper
 
@@ -530,7 +557,7 @@ class FlaskSessionStorage(Storage):
     """
 
     def locked_get(self):
-        serialized = session.get('google_oauth2_credentials')
+        serialized = session.get(_CREDENTIALS_KEY)
 
         if serialized is None:
             return None
@@ -541,8 +568,8 @@ class FlaskSessionStorage(Storage):
         return credentials
 
     def locked_put(self, credentials):
-        session['google_oauth2_credentials'] = credentials.to_json()
+        session[_CREDENTIALS_KEY] = credentials.to_json()
 
     def locked_delete(self):
-        if 'google_oauth2_credentials' in session:
-            del session['google_oauth2_credentials']
+        if _CREDENTIALS_KEY in session:
+            del session[_CREDENTIALS_KEY]
--- a/tests/test_flask_util.py
+++ b/tests/test_flask_util.py
@@ -26,6 +26,7 @@ import six.moves.urllib.parse as urlparse
 from oauth2client import GOOGLE_AUTH_URI
 from oauth2client import GOOGLE_TOKEN_URI
 from oauth2client import clientsecrets
+from oauth2client.flask_util import _get_flow_for_token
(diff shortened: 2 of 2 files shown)


jonparrott:
*Sigh*, python 2.6, you are a thorn in my side.

waprin:
LGTM. Tested it on my sample cases and it worked fine.

oauth2client/flask_util.py
```diff
@@ -331,6 +333,20 @@ def _make_flow(self, return_url=None, **kwargs):
             redirect_uri=url_for('oauth2.callback', _external=True),
             **kw)
 
+        flow_key = 'google_oauth2_flow_{0}'.format(csrf_token)
+        session[flow_key] = pickle.dumps(flow)
```
dhermes, line 351: Sorry for not mentioning before. Why do we use `pickle`?

theacodes, line 351: pickle's only used for the flow because it doesn't implement `to_json` and `from_json`. I think it should, but it might be out of the scope of this bug fix. Maybe something to follow up on?

dhermes, line 351: Filed #323

jonparrott:
@dhermes anything holding this back from being merged?

dhermes:
I didn't really review this and don't have throughput to do it. I'll leave it to @nathanielmanistaatgoogle

nathanielmanistaatgoogle:
I'll try to get to this sometime this week. Does it have time-sensitivity to it?

waprin:
I don't think it's especially urgent (though not sure what Jon thinks), but it does fix some critical bugs with the extension and it's a somewhat small patch.

jonparrott:
Not overly urgent as I doubt many people are using incremental auth at this point.

I would like to have this #326, #325, #323, and DictionaryStorage from #319 released together.

oauth2client/flask_util.py
```diff
@@ -331,6 +333,20 @@ def _make_flow(self, return_url=None, **kwargs):
...
+
+        return flow
+
+    def _get_flow_for_token(self, csrf_token):
```
nathanielmanistaatgoogle: No use of "self" parameter - why is this a private method at instance scope rather than a private function at module scope?

oauth2client/flask_util.py
```diff
@@ -331,6 +333,20 @@ def _make_flow(self, return_url=None, **kwargs):
...
+        flow_pickle = session.get(
+            'google_oauth2_flow_{0}'.format(csrf_token), None)
```
nathanielmanistaatgoogle: This string literal appears twice in the file (at least that I've seen so far). Define a constant? Or a helper function? No repeated string literals, please. :-)

oauth2client/flask_util.py
```diff
@@ -473,24 +492,25 @@ def required(self, decorated_function=None, scopes=None,
...
                 return_url = decorator_kwargs.pop('return_url', request.url)
 
-                # No credentials, redirect for new authorization.
-                if not self.has_credentials():
-                    auth_url = self.authorize_url(
-                        return_url,
-                        scopes=scopes,
-                        **decorator_kwargs)
-                    return redirect(auth_url)
+                requested_scopes = list(self.scopes) + (scopes or [])
```
nathanielmanistaatgoogle: Use of "and" and "or" for non-boolean values is now considered low-class. Although it's longer, please go with `[] if scopes is None else scopes` instead of `scopes or []`.

oauth2client/flask_util.py
```diff
@@ -473,24 +492,25 @@ def required(self, decorated_function=None, scopes=None,
...
+
+                if self.has_credentials():
+                    requested_scopes += list(self.credentials.scopes)
+
+                requested_scopes = list(set(requested_scopes))
```
nathanielmanistaatgoogle: I was suspecting that deduplication would be warranted! Why not just use a set for the accumulation all along?

oauth2client/flask_util.py
```diff
@@ -473,24 +492,25 @@ def required(self, decorated_function=None, scopes=None,
...
-                # No credentials, redirect for new authorization.
-                if not self.has_credentials():
-                    auth_url = self.authorize_url(
-                        return_url,
-                        scopes=scopes,
-                        **decorator_kwargs)
-                    return redirect(auth_url)
+                requested_scopes = list(self.scopes) + (scopes or [])
+
+                if self.has_credentials():
```
nathanielmanistaatgoogle, line 506: This smells inconsistent. How about

```
requested_scopes = set(self.scopes)
if scopes is not None:
  requested_scopes |= scopes
if self.has_credentials():
  requested_scopes |= self.credentials.scopes
```

? Double-check me; you may have to use `update` rather than `|=`.

oauth2client/flask_util.py
```diff
@@ -473,24 +492,25 @@ def required(self, decorated_function=None, scopes=None,
...
+                    requested_scopes += list(self.credentials.scopes)
+
+                requested_scopes = list(set(requested_scopes))
+
+                # If no credentials or if existing credentials but mismatching
+                # scopes, redirect for incremental authorization.
+                if (not self.has_credentials() or
```
nathanielmanistaatgoogle: For readability please consider flipping

```
if not <A> or not <B>:
  <statement>
  return <expression>
return <other expression>
```

into

```
if <A> and <B>:
  return <other expression>
else:
  <statement>
  return <expression>
```

.

tests/test_flask_util.py
```diff
@@ -282,15 +285,18 @@ def test_callback_view_errors(self):
...
-                state = json.dumps({
-                    'csrf_token': 'tokenz',
-                    'return_url': '/return_url'
-                })
+                    session.update(flask.session)
+                    csrf_token = session['google_oauth2_csrf_token']
+                    flow = self.oauth2._get_flow_for_token(csrf_token)
+                    state = flow.params['state']
 
+            with Http2Mock(status=500):
```
nathanielmanistaatgoogle: `six.moves.http_client.INTERNAL_SERVER_ERROR` rather than the literal numeric value 500, please.

Or `httplib.INTERNAL_SERVER_ERROR` as it's known locally in this file, I guess.

tests/test_flask_util.py
```diff
@@ -282,15 +285,18 @@ def test_callback_view_errors(self):
 
         # Code exchange error
         with self.app.test_client() as c:
-            with Http2Mock(status=500):
+            with self.app.test_request_context():
```
nathanielmanistaatgoogle: These ten lines look token-for-token identical to the ten lines fifty lines earlier. Pull out a common helper function?

tests/test_flask_util.py
```diff
@@ -303,6 +309,19 @@ def test_callback_view_errors(self):
             rv = c.get('/oauth2callback?state=%s&code=codez' % state)
             self.assertEqual(rv.status_code, httplib.BAD_REQUEST)
 
+        # Missing flow.
+        with self.app.test_client() as c:
```
nathanielmanistaatgoogle: Single-letter identifiers are disallowed by our style guide. Surely there must be some more descriptive way by which we might know the value assigned here?

tests/test_flask_util.py
```diff
@@ -303,6 +309,19 @@ def test_callback_view_errors(self):
...
+            with c.session_transaction() as session:
+                session['google_oauth2_csrf_token'] = 'tokenz'
+
+            state = json.dumps({
+                'csrf_token': 'tokenz',
+                'return_url': '/return_url'
+            })
+
+            rv = c.get('/oauth2callback?state=%s&code=codez' % state)
```
nathanielmanistaatgoogle: Initialisms are as bad as single-variable identifiers.

tests/test_flask_util.py
```diff
@@ -303,6 +309,19 @@ def test_callback_view_errors(self):
...
+            self.assertEqual(rv.status_code, httplib.BAD_REQUEST)
```
nathanielmanistaatgoogle: Feeling a little cheeky I would also be satisfied with `assertIs` here rather than `assertEqual`. Your call.

tests/test_flask_util.py
```diff
@@ -416,6 +442,32 @@ def two():
             rv = c.get('/two')
             self.assertEqual(rv.status_code, httplib.OK)
 
+    def test_incremental_auth_exchange(self):
+        self._create_incremental_auth_app()
+
+        with Http2Mock():
+            with self.app.test_client() as c:
+                with self.app.test_request_context():
```
nathanielmanistaatgoogle: These ten lines of code also look similar but are in fact slightly different. Worth capturing in the same helper function?

tests/test_flask_util.py
```diff
@@ -416,6 +442,32 @@ def two():
             rv = c.get('/two')
             self.assertEqual(rv.status_code, httplib.OK)
 
+    def test_incremental_auth_exchange(self):
+        self._create_incremental_auth_app()
+
+        with Http2Mock():
+            with self.app.test_client() as c:
```
nathanielmanistaatgoogle: Another single-letter identifier.

tests/test_flask_util.py
```diff
@@ -416,6 +442,32 @@ def two():
...
+                        return_url='/return_url',
+                        # Incremental auth scopes.
+                        scopes=['one', 'two'])
+                    with c.session_transaction() as session:
+                        session.update(flask.session)
+                        csrf_token = session['google_oauth2_csrf_token']
+                        flow = self.oauth2._get_flow_for_token(csrf_token)
+                        state = flow.params['state']
+
+                rv = c.get('/oauth2callback?state=%s&code=codez' % state)
```
nathanielmanistaatgoogle: Another initialism.

oauth2client/flask_util.py
theacodes: It now uses self. But, in general flask extensions tend to be completely contained in the extension class. _shrug_.

oauth2client/flask_util.py
theacodes: Done, pulled this and a few other string literals into private instance variables.

oauth2client/flask_util.py
theacodes: Done, via using your set accumulation method.

oauth2client/flask_util.py
theacodes: Done.

oauth2client/flask_util.py
theacodes, line 506: Brilliant. Done.

oauth2client/flask_util.py
theacodes: Done.

tests/test_flask_util.py
theacodes: Done.

tests/test_flask_util.py
theacodes: Done.

tests/test_flask_util.py
theacodes: Done, renamed to `client`. (I was following typical flask usage here)

tests/test_flask_util.py
theacodes: Done, renamed to `response`. (again I was following typical flask usage)

tests/test_flask_util.py
theacodes: I prefer `assertEqual` unless I'm comparing to None, True, or False.

tests/test_flask_util.py
theacodes: Done.

tests/test_flask_util.py
theacodes: Fixed.

tests/test_flask_util.py
theacodes: Fixed.

jonparrott:
Thank you for that thorough review @nathanielmanistaatgoogle. I think I got everything.

oauth2client/flask_util.py
nathanielmanistaatgoogle: Actually I'm seeing them at class scope rather than instance scope. Please place them at module scope. It's extraordinarily rare for private constants to ever belong at any other scope, and this isn't one of those cases.

oauth2client/flask_util.py
nathanielmanistaatgoogle: Huh; I didn't know that. But that's an implementation pattern that we needn't and shouldn't follow, no matter who else does it.

oauth2client/flask_util.py
```diff
@@ -331,6 +337,20 @@ def _make_flow(self, return_url=None, **kwargs):
...
+        flow_key = self._flow_key.format(csrf_token)
+        session[flow_key] = pickle.dumps(flow)
+
+        return flow
+
+    def _get_flow_for_token(self, csrf_token):
+        flow_pickle = session.get(
+            self._flow_key.format(csrf_token), None)
+
+        if not flow_pickle:
```
nathanielmanistaatgoogle: I think `flow_pickle is None` would more precisely convey your authorial intent here than `not flow_pickle`. If `flow_pickle` is the empty string at this point in the code, would you rather return or would you rather crash?

oauth2client/flask_util.py
```diff
@@ -331,6 +337,20 @@ def _make_flow(self, return_url=None, **kwargs):
...
+            return
```
nathanielmanistaatgoogle: Never mix return-a-value and implicitly-return-None in the same behavior body. If there is any `return <expression>` anywhere in a behavior body, then every `return` statement should be of the form `return <expression>` and there should be no way to reach the end of the behavior's statements without hitting a `return <expression`.

oauth2client/flask_util.py
theacodes: Agreed.

oauth2client/flask_util.py
theacodes: Fixed, explicit check for none. If something is stored there, it should be a valid pickle except in truly exceptional circumstances.

oauth2client/flask_util.py
theacodes: Fixed, also fixed the conditional where this is called.

jonparrott:
Okay, moved `_get_flow_for_token` into a private module-level function, and ensured tests all pass. @nathanielmanistaatgoogle it's ready for another look.

oauth2client/flask_util.py
```diff
@@ -191,6 +192,21 @@ def requires_calendar():
 __author__ = 'jonwayne@google.com (Jon Wayne Parrott)'
 
 DEFAULT_SCOPES = ('email',)
+CREDENTIALS_KEY = 'google_oauth2_credentials'
```
nathanielmanistaatgoogle: How about leading underscores? Or do you intend for these values to be used outside this module?

oauth2client/flask_util.py
```diff
@@ -191,6 +192,21 @@ def requires_calendar():
...
+CSRF_KEY = 'google_oauth2_csrf_token'
+
+
+def _get_flow_for_token(csrf_token):
+    """Retrieves the flow instance associated with a given CSRF token from
+    the Flask session."""
+    flow_pickle = session.get(
+        FLOW_KEY.format(csrf_token), None)
+
+    if flow_pickle is None:
```
nathanielmanistaatgoogle, line 205: Consider shortening this to `return None if flow_pickle is None else pickle.loads(flow_pickle)`?

oauth2client/flask_util.py
theacodes: Users could modify them if needed to avoid conflicts, but I imagine that's a pretty slim situation because they're prefixed with `google_`. What do you think?

oauth2client/flask_util.py
theacodes, line 205: Eh, that toes the line of being more terse than readable. I'd prefer not to, but if you feel strongly I'll change it.

oauth2client/flask_util.py
```diff
@@ -320,10 +336,11 @@ def _make_flow(self, return_url=None, **kwargs):
         kw = self.flow_kwargs.copy()
         kw.update(kwargs)
 
-        extra_scopes = util.scopes_to_string(kw.pop('scopes', ''))
-        scopes = ' '.join([util.scopes_to_string(self.scopes), extra_scopes])
+        extra_scopes = kw.pop('scopes', [])
+        scopes = util.scopes_to_string(
```
nathanielmanistaatgoogle: Is this `util.scopes_to_string` call still required? The `scope` parameter of `OAuth2WebServerFlow` is documented to be "string or iterable of strings, scope(s) of the credentials being requested.", which of course is the funniest thing I've seen today since a string is already an iterable of strings. I don't think that that will ever stop being funny.

theacodes: Hahaha. I don't think the call is needed. I'll remove it and see if anything blows up.

theacodes: Wasn't needed. Fixed.

tests/test_flask_util.py
```diff
@@ -272,36 +285,44 @@ def test_callback_view_errors(self):
...
-        with self.app.test_client() as c:
-            with c.session_transaction() as session:
+        with self.app.test_client() as client:
+            with client.session_transaction() as session:
                 session['google_oauth2_csrf_token'] = 'tokenz'
 
             state = '[{'
-            rv = c.get('/oauth2callback?state=%s&code=codez' % state)
-            self.assertEqual(rv.status_code, httplib.BAD_REQUEST)
+            response = client.get('/oauth2callback?state=%s&code=codez' % state)
```
nathanielmanistaatgoogle: Any reason for using `.format` in some places in this comment and `%` in other places?

oauth2client/flask_util.py
nathanielmanistaatgoogle: Let's be conservative and not support use of these constants outside of this module. If someone wants to violate our API and use them of course the interpreter will let them and if someone is so happy with how that turns out that they want formal support they can file an issue and explain the situation and add test coverage with their patch and we can go through the whole process.

oauth2client/flask_util.py
nathanielmanistaatgoogle, line 205: Let's go with either the conditional expression or putting the lower return statement in an `else` branch. Having it hang out at behavior top-level too easily misreads the reader.

nathanielmanistaatgoogle:
I think I'm done with comments now. Fingers crossed.

oauth2client/flask_util.py
theacodes, line 205: Done.

tests/test_flask_util.py
theacodes: Everything should be using `format` now.

jonparrott:
Fingers crossed here too, thanks for the review. :)

oauth2client/flask_util.py
nathanielmanistaatgoogle: Did this one sneak by in the crowd? I'm not yet seeing this done.

theacodes: Arg, yes, fixed (for real this time).

jonparrott:
Thanks!

PR #328, opened 2015-10-18
Syntax fixes for fetch_gae_sdk

jayvdb:

PR #330, opened 2015-10-19
Remove unused imports

jayvdb:


oauth2client/client.py
```diff
@@ -1106,7 +1105,10 @@ def _in_gae_environment():
         return SETTINGS.env_name in ('GAE_PRODUCTION', 'GAE_LOCAL')
 
     try:
-        import google.appengine
+        import google.appengine  # noqa: unused import
+    except ImportError:
+        pass
+    else:
```
theacodes, line 1111: Well, TIL about `try`...`else`.

nathanielmanistaatgoogle:
Thank you for the clean-up! Is this being done in the service of some larger objective or just because you noticed it and saw the chance to make it right?

jayvdb:
Many of these are bog standard import removals.  However some of these imports may have side effects that I am not aware of.

I am especially unsure about whether these `import six` are needed, as I dont know it well enough to say it has no side effects that are needed by these modules.  From my reading of the six docs, its sys.modules and sys.meta_path hacks dont have relevant side effects.

I also dont know mock very well, but 4d020992b3 appears to have intentionally removed the use of mock when test code was moved, but the import was left behind.

And the tests all pass.

jayvdb:
My objective is to fix the code quality of https://github.com/shazow/urllib3/blob/master/_travis/fetch_gae_sdk.py , so I am fixing the issues upstream; i.e. here.   #328 was part of that.  Happy to fix other issues while I am here.

nathanielmanistaatgoogle:
If any of our imports have side effects of which we're not aware and removing those imports breaks us we are Bad People Who Deserve What We Get™. Any imports that are needed only for their side effects and our textually unused in the file in which they appear should be clearly commented as such, though of course it's better to have no such imports at all.

Thank you for running the tests. I intend to merge this after our Travis run turns green.

Glad to have you contributing improvements no matter the reason. :+1:

dhermes:
I can "confirm" that none of the imports have implicit behavior that `oauth2client` depends on.

dhermes:
Don't worry about the drop in coverage, it's just math:

```
(A - n) / (B - n) < A / B
<==>
(A - n) B < A (B - n)
<==>
- Bn < -An
<==>
A < B
```

PR #292, opened 2015-08-26
Releasing 1.5.0.

dhermes:
/cc @nathanielmanistaatgoogle @jonparrott @craigcitro 

(Working out changelog still, will update PR once done)

PR #336, opened 2015-11-10
Fall back to credential refresh on EDEADLK

### LINKED ISSUE #335: multistore_file raises errno.EDEADLK when trying to open credential store
multistore_file claims to be process and thread safe, but if multiple threads try to refresh credentials simultaneously, it can result in an exception like the stack trace at the bottom of this issue.

multistore_file catches errno.ENOSYS and errno.ENOLCK already.  The simple fix would be to add EDEADLK to this list, since multiple threads opening the file can definitely result in this error.  Additionally, it might be useful to add an actual Python lock to protect this access, which would help a lot when threads (or, optionally, child processes) contend for this resource on startup.

Encountered exception while copying:
```python
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/gslib/command.py", line 1707, in PerformTask
    results = task.func(cls, task.args, thread_state=self.thread_gsutil_api)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/gslib/copy_helper.py", line 1796, in _PerformSlicedDownloadObjectToFile
    end_byte=args.end_byte))
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/gslib/copy_helper.py", line 2185, in _DownloadObjectToFileResumable
    digesters=digesters, progress_callback=progress_callback)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/gslib/cloud_api_delegator.py", line 225, in GetObjectMedia
    serialization_data=serialization_data, digesters=digesters)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/gslib/gcs_json_api.py", line 694, in GetObjectMedia
    serialization_data=serialization_data)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/gslib/gcs_json_api.py", line 716, in _PerformResumableDownload
    end_byte=end_byte, serialization_data=serialization_data)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/gslib/gcs_json_api.py", line 772, in _PerformDownload
    use_chunks=False)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/apitools/apitools/base/py/transfer.py", line 459, in GetRange
    additional_headers=additional_headers)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/apitools/apitools/base/py/transfer.py", line 396, in __GetChunk
    retries=self.num_retries)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/apitools/apitools/base/py/http_wrapper.py", line 325, in MakeRequest
    http, http_request, e, retry, max_retry_wait))
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/apitools/apitools/base/py/http_wrapper.py", line 316, in MakeRequest
    check_response_func=check_response_func)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/apitools/apitools/base/py/http_wrapper.py", line 365, in _MakeRequestNoRetry
    redirections=redirections, connection_type=connection_type)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/gslib/gcs_json_media.py", line 377, in NewRequest
    connection_type=connection_type)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/oauth2client/oauth2client/util.py", line 137, in positional_wrapper
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/oauth2client/oauth2client/client.py", line 541, in new_request
    self._refresh(request_orig)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/apitools/apitools/base/py/credentials_lib.py", line 308, in _refresh
    oauth2client.client.OAuth2Credentials._refresh(self, do_request)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/oauth2client/oauth2client/client.py", line 771, in _refresh
    self.store.acquire_lock()
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/oauth2client/oauth2client/multistore_file.py", line 224, in acquire_lock
    self._multistore._lock()
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/oauth2client/oauth2client/multistore_file.py", line 284, in _lock
    self._file.open_and_lock()
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/oauth2client/oauth2client/locked_file.py", line 374, in open_and_lock
    self._opener.open_and_lock(timeout, delay)
  File "/usr/local/share/google/google-cloud-sdk/platform/gsutil/third_party/oauth2client/oauth2client/locked_file.py", line 209, in open_and_lock
    fcntl.lockf(self._fh.fileno(), fcntl.LOCK_EX)
IOError: [Errno 35] Resource deadlock avoided
```

thobrla:
Fixes https://github.com/google/oauth2client/issues/335

This change fixes a bug where multiple threads and/or processes
using multistore_file to access the same backing store could
raise IOError errno.EDEADLK to the calling application. Since
EDEADLK is a possibility with concurrent access, this change
instead causes a fallback to read only mode and refresh
credentials if necessary.

### DIFF
--- a/tests/test_multistore_file.py
+++ b/tests/test_multistore_file.py
@@ -1,0 +1,43 @@
+"""Unit tests for oauth2client.multistore_file."""
+
+import errno
+import os
+import tempfile
+import unittest
+
+from oauth2client import multistore_file
+
+
+class _MockLockedFile(object):
+
+    def __init__(self, filename_str, error_code):
+        self.filename_str = filename_str
+        self.error_code = error_code
+        self.open_and_lock_called = False
+
+    def open_and_lock(self):
+        self.open_and_lock_called = True
+        raise IOError(self.error_code, '')
+
+    def is_locked(self):
+        return False
+
+    def filename(self):
+        return self.filename_str
+
+
+class MultistoreFileTests(unittest.TestCase):
+
+    def test_lock_file_raises_ioerror(self):
+        filehandle, filename = tempfile.mkstemp()
+        os.close(filehandle)
+
+        try:
+            for error_code in (errno.EDEADLK, errno.ENOSYS, errno.ENOLCK):
+                multistore = multistore_file._MultiStore(filename)
+                multistore._file = _MockLockedFile(filename, error_code)
+                # Should not raise even though the underlying file class did.
+                multistore._lock()
+                self.assertTrue(multistore._file.open_and_lock_called)
+        finally:
+            os.unlink(filename)
--- a/oauth2client/multistore_file.py
+++ b/oauth2client/multistore_file.py
@@ -290,6 +290,9 @@ class _MultiStore(object):
             elif e.errno == errno.ENOLCK:
                 logger.warn('File system is out of resources for writing the '
                             'credentials file (is your disk full?).')
+            elif e.errno == errno.EDEADLK:
+                logger.warn('Lock contention on multistore file, opening '
+                            'in read-only mode.')
             else:
                 raise
         if not self._file.is_locked():


thobrla:
Using threading.Lock / multiprocessing.Lock is substantially more complex because it's necessary to pass one per _MultistoreFile and then plumb them through.  We would need this code regardless to protect against multi-application calls, so starting out with this.

Also, would like this in a release soon if that is feasible.

dhermes:
@thobrla Can you add tests? At some point #212 will be complete and there will be a test coverage CI test which will fail if we go below 100% line coverage

dhermes:
In particular, don't you have some code which caused this to fail which could be used to protect against regression for all time?

thobrla:
Added tests.

tests/test_multistore_file.py
```diff
@@ -0,0 +1,44 @@
...
+
+        filehandle, filename = tempfile.mkstemp()
+        os.close(filehandle)
+
+        try:
+            for error_code in (errno.EDEADLK, errno.ENOSYS, errno.ENOLCK):
+
+                class MockLockedFile(object):
+
+                    def __init__(self, filename_str, error_code):
```
nathanielmanistaatgoogle: This binding of the name `error_code` shadows the outer binding of the name `error_code`. Right?

tests/test_multistore_file.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def test_lock_file_raises_ioerror(self):
+
+        filehandle, filename = tempfile.mkstemp()
+        os.close(filehandle)
+
+        try:
+            for error_code in (errno.EDEADLK, errno.ENOSYS, errno.ENOLCK):
+
+                class MockLockedFile(object):
```
nathanielmanistaatgoogle: Why is this class nested all the way in here rather than defined at module top-level?

tests/test_multistore_file.py
```diff
@@ -0,0 +1,44 @@
...
+
+                    def is_locked(self):
+                        return False
+
+                    def filename(self):
+                        return self.filename_str
+
+                multistore = multistore_file._MultiStore(filename)
+                multistore._file = MockLockedFile(filename, error_code)
+
```
nathanielmanistaatgoogle: After pulling the class out of this function, drop all the blank lines in this function except this one? (Or push back if your aesthetic sense guides you otherwise.)

tests/test_multistore_file.py
```diff
@@ -0,0 +1,44 @@
...
+                # Should not raise even though the underlying file class did.
+                multistore._lock()
+
+        finally:
+            os.unlink(filename)
+
```
nathanielmanistaatgoogle: Unnecessary end-of-file blank line? Or is GitHub lying to me with its visual presentation?

nathanielmanistaatgoogle:
I'm pretty unfamiliar with both the problem and the solution at play here; I'd feel more comfortable with this if @dhermes also reviewed and approved this change.

tests/test_multistore_file.py
thobrla: Because the mock's behavior is specific to this test, and I'm not intending to write a generic mock for all of multistore_file.  In we add more tests in a future CL, it can be abstracted, but I'm not in favor of trying to make it generic out of the gate.

That being said, it should exist outside of the try block here; originally I implemented just using the outer error_code.

tests/test_multistore_file.py
thobrla: Done.

tests/test_multistore_file.py
thobrla: Done.

tests/test_multistore_file.py
thobrla: Correct; moved this out of the loop.

dhermes:
FYI the test error is related to `tox==2.2.1` which was released on November 11, 2015. I can confirm this happen locally as well. (I think we have a "bad" config and old versions of `tox` let it slide.)

The issue is caused by https://github.com/google/oauth2client/blob/311a53fbcae550526a74057f27cb9df73a0af009/tox.ini#L86-L87

tests/test_multistore_file.py
```diff
@@ -0,0 +1,40 @@
...
+                return self.filename_str
+
+        filehandle, filename = tempfile.mkstemp()
+        os.close(filehandle)
+
+        try:
+            for error_code in (errno.EDEADLK, errno.ENOSYS, errno.ENOLCK):
+                multistore = multistore_file._MultiStore(filename)
+                multistore._file = MockLockedFile(filename, error_code)
+                # Should not raise even though the underlying file class did.
```
dhermes, line 39: Can you verify in the test that the underlying class/instance raised?

tests/test_multistore_file.py
```diff
@@ -0,0 +1,40 @@
...
+                multistore._lock()
+        finally:
+            os.unlink(filename)
```
dhermes, line 43: Prefer this over `remove()`? Also, it's a `tempfile`, so we don't need to remove it, the extra I/O just slows down the test suite. Am I missing something?

dhermes:
RE: Release process. Pushing a new version is as simple as pushing a new tag. My simple process for new release

1.  Actually increment the version and update the changelog in a PR. (#292 is a great example and includes a `bash` snippet I use to get a snapshot of all PRs merged since the previous release)
2.  Pull in all the changes after the version update PR is merged.
3.  Create the tag:

    ```
    $ git tag vX.Y.Z
    ```

4.  Push the tag to GitHub

    ```
    $ git push official vX.Y.Z
    ```

**NOTE**: I work with two remotes (`origin` and `official`):

```bash
$ git remote -vv
official        git@github.com:google/oauth2client.git (fetch)
official        git@github.com:google/oauth2client.git (push)
origin  git@github.com:dhermes/oauth2client.git (fetch)
origin  git@github.com:dhermes/oauth2client.git (push)
```

tests/test_multistore_file.py
nathanielmanistaatgoogle: Please place the class at module scope. It's not that the class needs a reason to move to module-scope, it's that module-scope is the default place for classes and they should only be promoted into the scope of other code elements when necessary (and "is only ever used from" is not sufficient reason).

nathanielmanistaatgoogle:
Please also squash commits; there's no reason for project history to include all our review back-and-forth.

tests/test_multistore_file.py
thobrla: Done

tests/test_multistore_file.py
thobrla, line 43: Done.

thobrla:
Planning to squash commits once I have reviewer approval, or would you prefer that I squash them with each iteration?

dhermes:
SGTM RE: Squashing after review is completed.

tests/test_multistore_file.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def filename(self):
+        return self.filename_str
+
+
+class MultistoreFileTests(unittest.TestCase):
+
+    def test_lock_file_raises_ioerror(self):
+        filehandle, filename = tempfile.mkstemp()
+        os.close(filehandle)
```
dhermes, line 33: Sorry I didn't ask before. What is the purpose of the `os.close` here? (I'm trying to soak up your OS knowledge.)

dhermes:
BTW This LGTM, Let's let @nathanielmanistaatgoogle weigh in before considering this ready to squash and merge.

tests/test_multistore_file.py
thobrla, line 33: Since we're not using the filehandle, there's no reason to leave it open (consuming a filehandle resource in the OS).

We need to use [tempfile.mkstemp()](https://docs.python.org/2/library/tempfile.html) to ensure that the temporary file (which the _Multistore constructor will open for reading via its name) is visible on the filesystem.  We're also responsible for cleaning the file up.  So removing the os.unlink() call was a mistake.

os.unlink and os.remove are identical, so that's purely stylistic.

dhermes, line 33: My previous comment was that the style of using `unlink` would be a negative in a codebase with lots of authors, since many would not know that it was an alias.

I'm also unclear why it matters that we're responsible for cleaning up the file. It's a tempfile, the OS will clean it up eventually.

As I said, the delete is "unnecessary" I/O and fast unit tests is an aspirational goal to encourage new devs to contribute to a project. (**UPDATE**: Or to at least keep from discouraging them.)

thobrla, line 33: It depends on where those authors are coming from; either way, both functions are documented in the [os module documentation](https://docs.python.org/2/library/os.html#os.unlink) and no preference is stated there.

The I/O is not unnecessary - the documentation for mkstemp explicitly states that it is the user's responsibility to clean it.  While it may currently be in a temporary directory that's cleaned by the OS, that's not guaranteed in the future. 

These unit tests already involve many file accesses, so I do not think that this will impact performance in a meaningful way.  If you disagree, provide some data explaining the unit test performance targets?

dhermes, line 33: Feel free to leave as-is.

> depends on where those authors are coming from

That was my point. "All" authors know what remove means, not all know that unlink is an alias and forcing them to Google it is unnecessary.

> provide some data 

No time to do that, not worth the battle. (Though these unit tests need a lot of work still, there are actually `time.sleep` statements elsewhere in the `tests/` subdir.)

dhermes:
@thobrla Can you rebase after #340 went in so that we can see the tests run?

nathanielmanistaatgoogle:
Change *content* LGTM; please produce a passing test run and single commit.

@dhermes please feel free to merge when satisfied.

dhermes:
@thobrla Travis finally ran the tests! All green.

Ping me when you've squashed commits?

dhermes:
JK Just realized they were already squashed. Good call.

thobrla:
Not sure why the tests took so long to run; build sat in a pending state for several hours before finally running.

@dhermes Can you pull this into a release?

dhermes:
Yeah Travis sometimes has issues or like doesn't like the `google` org / thinks it is too resource hungry. The longest we've waited for a job to run is 2.5 days, but usually it's immediate.

Can you assist with the release (send the PR with release notes and version bump as noted [above][1])? Then I'll accept the PR / push the new tag.

[1]: https://github.com/google/oauth2client/pull/336#issuecomment-156743551

PR #337, opened 2015-11-12
Avoid OAuth2Credentials id_token going out of sync after a token refresh.

danring:
After token refresh, `credential.id_token` no longer matches `_extract_id_token(credential.token_response['id_token'])`, as it does initially.  `access_token`, `token_expiry` are updated, but not `id_token`.

E.g., https://github.com/google/oauth2client/blob/master/oauth2client/client.py#L877 doesn't quite match https://github.com/google/oauth2client/blob/master/oauth2client/client.py#L2141

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -883,6 +883,10 @@ class OAuth2Credentials(Credentials):
                     seconds=int(d['expires_in'])) + datetime.datetime.utcnow()
             else:
                 self.token_expiry = None
+            if 'id_token' in d:
+                self.id_token = _extract_id_token(d['id_token'])
+            else:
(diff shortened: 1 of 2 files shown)


dhermes:
I like it. We need to resolve the test issue (an issue with our config / `tox`) and then re-run these tests.

dhermes:
@danring Do you mind re-basing so we can see the tests run? (I got #340 in to "fix" the `tox` config issue.)

dhermes:
@nathanielmanistaatgoogle WDYT?

nathanielmanistaatgoogle:
@dhermes: looks good; merge when you're satisfied too.

Issue #339, opened 2015-11-15
tox config is broken


dhermes:
Failures noticed in #336 and #337.

From http://tox.readthedocs.org/en/latest/changelog.html it looks like the 2.2.1 release notes explain the failure

> fix bug where `{envdir}` substitution could not be used in setenv if that env value is then used in `{basepython}`. Thanks Florian Bruhin.

Seems https://bitbucket.org/hpk42/tox/pull-requests/169/tries-to-fix-99/diff is the PR that fixed something but broke us

dhermes:
I should add that the offending line is

https://github.com/google/oauth2client/blob/311a53fbcae550526a74057f27cb9df73a0af009/tox.ini#L86-L87

and the issue is a circular reference between `setenv` and the use of `env:KEY:DEFAULT`.

dhermes:
@jonparrott Any idea on how to fix this? (I did the "cowardly" thing and downgraded `tox` temporarily.)

dhermes:
I accidentally stumbled on something which I think/hope will solve this:
https://github.com/GoogleCloudPlatform/gcloud-python/pull/1237

jonparrott:
All I've got is to just wait. We ended up pinning tox in [python-docs-samples](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/141) as well.

Consider tox isn't a *direct* dependency of library, pinning it for a while doesn't seem too offensive.

dhermes:
I don't think waiting will work. I think we are permanently broken. I tried making the change suggested by my other PR and get a circular dependency sort of error:

```python
...
  File "/usr/local/lib/python2.7/dist-packages/tox/config.py", line 959, in _replace_substitution
    val = self._substitute_from_other_section(sub_key)
  File "/usr/local/lib/python2.7/dist-packages/tox/config.py", line 944, in _substitute_from_other_section
    (section, item), self._subststack))
ValueError: ('testenv', 'gae_pythonpath') already in [('testenv:gae', 'setenv'), ('testenv', 'gae_pythonpath'), ('testenv:gae', 'setenv')]
```

It seems like it is getting a self-reference and it will "always" be impossible for `setenv` and `env` to coexist.

jonparrott:
@waprin who's more involved in the testing side than me.

dhermes:
@jonparrott By forcing me to articulate that 

> it will "always" be impossible for `setenv` and `env` to coexist.

[...]

PR #340, opened 2015-11-16
Pinning tox to version 2.1.1.

dhermes:
This is a temporary workaround due to the config parsing
change introduced in 2.2.0.

See #339 for more details.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -16,7 +16,7 @@
 
 set -ev
 
-pip install tox
+pip install tox==2.1.1
 if [[ "${TOX_ENV}" == "pypy" ]]; then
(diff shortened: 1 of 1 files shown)


dhermes:
@nathanielmanistaatgoogle This should be merged ASAP so we can actually see tests run on other PRs.

FYI @thobrla this should at least allow the tests to run in #336

dhermes:
@thobrla I'm going to pre-emptively merge this so you can get tests passing (to make #336 smooth sailing).

We can roll-back if @nathanielmanistaatgoogle has issues after the fact.

nathanielmanistaatgoogle:
LGTM; thank you for not blocking on me for something so small and straightforward.

dhermes:
Sure thing. Thanks for understanding.

PR #346, opened 2015-11-30
Adding empty oauth2client.contrib

### LINKED ISSUE #326: Create oauth2client.contrib package for non-core modules.
Similar to urllib3's contrib [[1](https://github.com/shazow/urllib3/tree/master/urllib3/contrib)] [[2](http://urllib3.readthedocs.org/en/latest/contrib.html)], we will move ``appengine``, ``flask_util``, ``django_orm``, and in incoming ``django_util`` into ``oauth2client.contrib``. This cleanly separates these framework/environment-specific modules away from the core library, and will make it easier to separate into a separate pypi package later if needed.

See #319 for initial discussion.

Should we have import aliases for backwards compatibility or should we explicitly break imports with this change?

theacodes:
Fixes #326, follow-up PRs will be submitted to move `flask_util`, `appengine`, and `django_orm`.

### DIFF
--- a/oauth2client/contrib/__init__.py
+++ b/oauth2client/contrib/__init__.py
@@ -1,0 +1,6 @@
+"""Contributed modules.
+
+Contrib contains modules that are not considered part of the core oauth2client
+library but provide additional functionality. These modules are intended to
+make it easier to use oauth2client.
+"""
--- a/docs/source/oauth2client.contrib.rst
+++ b/docs/source/oauth2client.contrib.rst
@@ -1,0 +1,10 @@
+oauth2client.contrib package
+============================
+
+Module contents
+---------------
+
+.. automodule:: oauth2client.contrib
+    :members:
+    :undoc-members:
+    :show-inheritance:
--- a/docs/source/oauth2client.rst
+++ b/docs/source/oauth2client.rst
@@ -1,6 +1,13 @@
 oauth2client package
 ====================
 
+Subpackages
+-----------
+
+.. toctree::
+
+    oauth2client.contrib
+
 Submodules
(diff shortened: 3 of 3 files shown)


nathanielmanistaatgoogle:
Looks good to me; waiting on test results.

jonparrott:
@dhermes' tricky docs thing got me again. It should pass tests now.

dhermes:
Do you see a less tricky way to make sure that contributors that add new code also document their new code?

jonparrott:
No, I think it's great. I've just been dumb enough to let me bite me twice now. :)

oauth2client/contrib/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""
```
nathanielmanistaatgoogle: Sadly there's not an example in the style guide to which I can refer, but the format of module doc strings should be

```
"""A single summary line on the same line as the opening triple-quote.

Zero or more additional paragraphs individually separated by single blank
lines.

The closing triple-quote goes at the end of the summary line if there
are zero additional paragraphs and goes on its own line if there are one or
more additional paragraphs.
"""
```

.

theacodes: Updated, let me know if you like the new version better.

nathanielmanistaatgoogle: Looks great; will merge after tests pass again.

theacodes: :+1: thank you.

jonparrott:
@nathanielmanistaatgoogle ping - travis is green. :)

PR #347, opened 2015-11-30
[DO NOT MERGE] Re-write Storage

theacodes:
This is a rough-cut to sanity check the concept. Docstrings and full tests still missing.

New:
- `client.Storage` is an abstract base class.
- `locked_storage.LockedStorage` provides locking based on context managers and defaults to `threading.Lock`.
- `locked_storage.threadsafe` is a class-level decorator to provide thread safety. This is applied to `FileStorage` and `KeyringStorage`.

Changed:
- All `Storage` subclasses have been updated to use the new base class.
- `file.Storage` renamed to `file_storage.FileStorage`.
- `keyring_storage.Storage` renamed to `keyring_storage.KeyringStorage`.
- `django_orm.Storage` renamed to `django_orm.DjangoOrmStorage`.

**Questionable changes**

As mentioned in [a comment on #344](https://github.com/google/oauth2client/pull/344#issuecomment-160785708), `Credentials` has the unfortunate behavior of using the Storage's lock to prevent concurrently refreshing the credentials. IMO, Refresh locking should be handled as a separate concern. For now, `Storage` implements a no-op `lock()`.

### DIFF
--- a/oauth2client/appengine.py
+++ b/oauth2client/appengine.py
@@ -464,7 +464,7 @@ class StorageByKeyName(Storage):
             db.delete(entity_key)
 
     @db.non_transactional(allow_existing=True)
-    def locked_get(self):
+    def get(self):
         """Retrieve Credential from datastore.
 
         Returns:
@@ -487,7 +487,7 @@ class StorageByKeyName(Storage):
         return credentials
 
     @db.non_transactional(allow_existing=True)
-    def locked_put(self, credentials):
+    def put(self, credentials):
         """Write a Credentials to the datastore.
 
         Args:
@@ -500,7 +500,7 @@ class StorageByKeyName(Storage):
             self._cache.set(self._key_name, credentials.to_json())
 
     @db.non_transactional(allow_existing=True)
-    def locked_delete(self):
+    def delete(self):
         """Delete Credential from datastore."""
 
         if self._cache:
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -19,6 +19,7 @@ Tools for interacting with OAuth 2.0 protected resources.
 
 import base64
 import collections
+import contextlib
 import copy
 import datetime
 import json
@@ -336,95 +337,38 @@ class Flow(object):
 class Storage(object):
     """Base class for all Storage objects.
 
-    Store and retrieve a single credential. This class supports locking
-    such that multiple processes and threads can operate on a single
-    store.
+    Store and retrieve a single credential.
     """
 
-    def acquire_lock(self):
-        """Acquires any lock necessary to access this Storage.
-
-        This lock is not reentrant.
-        """
-        pass
-
-    def release_lock(self):
-        """Release the Storage lock.
-
-        Trying to release a lock that isn't held will result in a
-        RuntimeError.
-        """
-        pass
-
-    def locked_get(self):
+    def get(self):
         """Retrieve credential.
-
-        The Storage lock must be held when this is called.
 
         Returns:
             oauth2client.client.Credentials
         """
         _abstract()
 
-    def locked_put(self, credentials):
+    def put(self, credentials):
         """Write a credential.
-
-        The Storage lock must be held when this is called.
 
         Args:
             credentials: Credentials, the credentials to store.
         """
         _abstract()
 
-    def locked_delete(self):
+    def delete(self):
         """Delete a credential.
 
         The Storage lock must be held when this is called.
         """
         _abstract()
 
-    def get(self):
-        """Retrieve credential.
-
-        The Storage lock must *not* be held when this is called.
-
-        Returns:
-            oauth2client.client.Credentials
-        """
-        self.acquire_lock()
-        try:
-            return self.locked_get()
-        finally:
-            self.release_lock()
-
-    def put(self, credentials):
-        """Write a credential.
-
-        The Storage lock must be held when this is called.
-
-        Args:
-            credentials: Credentials, the credentials to store.
-        """
-        self.acquire_lock()
-        try:
-            self.locked_put(credentials)
-        finally:
-            self.release_lock()
-
-    def delete(self):
-        """Delete credential.
-
-        Frees any resources associated with storing the credential.
-        The Storage lock must *not* be held when this is called.
-
-        Returns:
-            None
-        """
-        self.acquire_lock()
-        try:
-            return self.locked_delete()
-        finally:
-            self.release_lock()
+    def lock(self):
+        """TODO"""
+        @contextlib.contextmanager
+        def noop():
+            yield
+        return noop()
 
 
 def clean_headers(headers):
@@ -841,9 +785,8 @@ class OAuth2Credentials(Credentials):
         if not self.store:
             self._do_refresh_request(http_request)
         else:
-            self.store.acquire_lock()
-            try:
-                new_cred = self.store.locked_get()
+            with self.store.lock():
+                new_cred = self.store.get()
 
                 if (new_cred and not new_cred.invalid and
                         new_cred.access_token != self.access_token and
@@ -852,8 +795,6 @@ class OAuth2Credentials(Credentials):
                     self._updateFromCredential(new_cred)
                 else:
                     self._do_refresh_request(http_request)
-            finally:
-                self.store.release_lock()
 
     def _do_refresh_request(self, http_request):
         """Refresh the access_token using the refresh_token.
@@ -887,7 +828,7 @@ class OAuth2Credentials(Credentials):
             # re-authorize, so we unflag here.
             self.invalid = False
             if self.store:
-                self.store.locked_put(self)
+                self.store.put(self)
         else:
             # An {'error':...} response body means the token is expired or
             # revoked, so we flag the credentials as such.
@@ -901,7 +842,7 @@ class OAuth2Credentials(Credentials):
                         error_msg += ': ' + d['error_description']
                     self.invalid = True
                     if self.store:
-                        self.store.locked_put(self)
+                        self.store.put(self)
             except (TypeError, ValueError):
                 pass
             raise HttpAccessTokenRefreshError(error_msg, status=resp.status)
--- a/oauth2client/file_storage.py
+++ b/oauth2client/file_storage.py
@@ -19,7 +19,6 @@ credentials.
 """
 
 import os
-import threading
 
 from oauth2client.client import Credentials
 from oauth2client.client import Storage as BaseStorage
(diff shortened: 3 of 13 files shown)


jonparrott:
@dhermes I tried to keep this small, but I wanted to make sure all existing tests still pass. Sorry if I missed the mark.

dhermes:
Ha. `13 files != small` Which mark did you hope you hadn't missed?

jonparrott:
I wanted to land somewhere between "small" and "please don't make me look at this".

oauth2client/client.py
```diff
@@ -336,95 +337,38 @@ class Flow(object):
...
-    def locked_get(self):
+    def get(self):
         """Retrieve credential.
 
-        The Storage lock must be held when this is called.
-
         Returns:
             oauth2client.client.Credentials
         """
         _abstract()
```
dhermes, line 349: If we're going to re-factor we should aspire to nuke `_abstract()`, but maybe not in this PR?

dhermes:
> IMO, Refresh locking should be handled as a separate concern

How should it be handled? The user never even sees a `_refresh` occur, it just happens automatically on a 401 from the server.

oauth2client/appengine.py
```diff
@@ -464,7 +464,7 @@ def _delete_entity(self):
             db.delete(entity_key)
 
     @db.non_transactional(allow_existing=True)
-    def locked_get(self):
+    def get(self):
```
dhermes, line 467: Why get rid of the `locked_`?

oauth2client/client.py
```diff
@@ -841,9 +785,8 @@ def _refresh(self, http_request):
         if not self.store:
             self._do_refresh_request(http_request)
         else:
-            self.store.acquire_lock()
-            try:
-                new_cred = self.store.locked_get()
+            with self.store.lock():
```
dhermes, line 788: Massive :+1: to this pattern.

I'd wager that you need to very clearly set expectations for whether `get/set/delete` are expected to happen while locked or while not. At the very least someone should rewrite the docstrings for existing implementations to make this clear.

oauth2client/file_storage.py
```diff
@@ -32,34 +31,18 @@ class CredentialsFileSymbolicLinkError(Exception):
...
     def __init__(self, filename):
         self._filename = filename
-        self._lock = threading.Lock()
 
     def _validate_file(self):
         if os.path.islink(self._filename):
             raise CredentialsFileSymbolicLinkError(
                 'File: %s is a symbolic link.' % self._filename)
 
-    def acquire_lock(self):
```
dhermes, line 47: Shouldn't this be replaced by an implementation of `lock`?

oauth2client/locked_storage.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, storage, lock=None):
+        """TODO"""
+        self._storage = storage
+
+        if lock is None:
+            lock = threading.Lock()
+
+        self._lock = lock
+
+    def get(self, *args, **kwargs):
```
dhermes, line 35: OK now I'm confused. This makes it so that a generic storage can't be used with `Credentials.set_store`. I was imagining since not all storage types have a "locked" guarantee.

oauth2client/multistore_file.py
```diff
@@ -481,4 +474,7 @@ def _get_storage(self, key):
         Returns:
             A Storage object that can be used to get/set this cred
         """
-        return self._Storage(self, key)
```
dhermes, line 484: Should `self._Storage` be deleted?

tests/test_django_orm.py
```diff
@@ -137,7 +137,7 @@ def test_constructor(self):
         key_name = 'foo'
         key_value = 'bar'
         property_name = 'credentials'
-        storage = Storage(FakeCredentialsModel, key_name,
+        storage = DjangoOrmStorage(FakeCredentialsModel, key_name,
                           key_value, property_name)
```
dhermes, line 141: lint indent

tests/test_file.py
```diff
@@ -31,7 +31,7 @@
 from .http_mock import HttpMockSequence
 import six
 
-from oauth2client import file
```
dhermes, line 34: Massive :+1:. Who had the idea to rename? Is it worth providing the old names for compatibility for a generation?

tests/test_jwt.py
```diff
@@ -32,7 +32,7 @@
 from oauth2client.client import HAS_OPENSSL
 from oauth2client.client import HAS_CRYPTO
 from oauth2client import crypt
-from oauth2client.file import Storage
+from oauth2client.file_storage import FileStorage as Storage
```
dhermes, line 35: Not worth the effort of renaming throughout?

oauth2client/client.py
theacodes, line 349: Eh, I'll leave that up to @nathanielmanistaatgoogle.

oauth2client/appengine.py
theacodes, line 467: Matches the base class.

oauth2client/client.py
theacodes, line 788: Yeah this is a work in progress, I haven't quite decided on the behavior here. Will follow up in a top-level comment.

oauth2client/file_storage.py
theacodes, line 47: Not 100% sure yet, but the way this is currently done `File` is no longer threadsafe, but doing:

``` python
storage = LockedStorage(FileStorage())
```

Would be.

oauth2client/locked_storage.py
theacodes, line 35: Yeah, point of contention, following up with that.

oauth2client/multistore_file.py
theacodes, line 484: I haven't had a chance to fully scrub this file. This particular module is really complicated so I want to make sure I'm 100% confident in the Storage design before updating it.

tests/test_file.py
theacodes, line 34: Mine, it's been bugging me for a while. I personally think if we're going to break everything, we might as well not provide proxies for the old ones. That's @nathanielmanistaatgoogle and your call though.

tests/test_jwt.py
theacodes, line 35: Just trying to get tests to pass for now. Will more thoroughly address this.

jonparrott:
Okay, point of contention brought up by myself and @waprin. I could use some thoughts here.

I initially thought locking was only to prevent concurrent calls to `get`, `set`, and `delete`. It seemed better to pull locking into a separate concern from storage. However, this is not the case as locking is also used to prevent concurrent refreshing of the access token.

So, the previous design made these assumptions:

1. `Storage` is responsible for locking `get`, `put`, and `delete`.
2. `Credentials` are responsible for holding the storage lock during refresh.
3. A `Storage` instance is inherently designed to store a single `Credentials` instance..
4. `FileStorage` and `KeyringStorage` made the incorrect assumption of always using `threading.Lock` and providing no clear way for users to override them.
5. Other storage classes provided no way to provide locking.

I don't consider 1, 2, and 3 *bad*, but I'm willing to reconsider if someone else has better ideas. I do consider 4 & 5 the core of the problem here.

So I think we should design the replacement to Storage to:

1. Accept a context manager for locking and default to a no-op context manager.
2. Provide the public interface `get`, `put`, `delete`, and `lock`. `get`, `put`, and `delete` are responsible for locking and then calling the respective private method.
3. Private methods `_get`, `_put`, `_delete` are implemented by sub-classes. I'm not a fan of this, but I'm struggling to come up with a better solution.
3. `lock` is a context manager used by `Credentials.refresh` to hold the lock during refresh.

So basically more or less what I have now just with absorbing `LockedStorage` into the base class.

oauth2client/client.py
nathanielmanistaatgoogle, line 349: Absolutely nuke it now. An abstract/interface class should look mostly like

```
import abc

class MyAbstractClass(object):
    """My class doc string.

    More documentation.
    """
    __metaclass__ = abc.ABCMeta

    def MyAbstractMethod(self):
        """My abstract method doc string."""
        raise NotImplementedError()
```

.

tests/test_file.py
nathanielmanistaatgoogle, line 34: It's been bugging me too. I also think it's fine not to provide the old ones.

nathanielmanistaatgoogle:
I'm having a hard time following about the locks, but the thought of having one single lock that is used both within `Storage` and outside it does tickle some abstraction code smells to me. If `Storage` is just going to be an interface, it probably shouldn't make any specification other than "instances of this class must be safe for use from concurrent threads" or "instances of this class need not be safe for use from concurrent threads".

When we talk about some code being smart enough to want to ensure that only one credentials-refresh is going on, that seems smart but it also seems decoupled from the idea of locally storing and retrieving the credentials, and that decoupling, to me, suggests an independent lock.

What's the expected concurrent use of `get`, `put`, and `delete` that motivates a `Storage`-internal lock? Why should instances of `Storage` (for the moment considering just those three methods) have an internal lock?

jonparrott:
I'm starting to see why other oauth libraries don't provide storage.

> What's the expected concurrent use of get, put, and delete

If you think in terms of concrete storage classes, locking these methods individually is rarely needed. For example, database-backed storage classes don't need to this as the database itself handles the concurrency. The session-based storage for Flask, Django, and Bottle will use the framework's session object which in all cases is a context (thread/greenlet) local, so no need for locking there either. Local file storage may need this, though.

In terms of supporting locking during credential refresh, the use case here is that there are multiple processes (threads, processes, instances, etc.) that are using the same set of credentials. Imagine a multithreaded program that downloads all of the user's files from Google Drive. If this runs for more than an hour the credentials will be invalid and need to be refreshed. The first thread to notice this should refresh and store the credentials. The remaining threads will block until the refresh is done and then continue using the new credentials. Without the lock, each thread would end up refreshing the credentials.

Now that I'm getting a clearer picture of usage here, it seems that having storage just for the purpose of *storing* credentials seems pointless. You can do that with `.to_json()` and `.from_json()` and any database. It seems the purpose of storage is to ensure that refreshing credentials happen once and that all processes using those credentials can obtain the new access token.

jonparrott:
FYI, oauth2lib does appear to do any locking. It just provides a way for you to specify a function to save the new credentials. [link](https://requests-oauthlib.readthedocs.org/en/latest/oauth2_workflow.html#all-define-the-token-token-saver-and-needed-credentials)

```python
from requests_oauthlib import OAuth2Session

def token_saver(token):
    # write to database.
    pass

client = OAuth2Session(
     client_id, token=token, auto_refresh_url=refresh_url,
     auto_refresh_kwargs=extra, token_updater=token_saver)
```

dhermes:
@jonparrott Sorry I thought we were on the same page about why `Credentials.set_store` and `Credentials._refresh` needed to be best pals.

jonparrott:
@dhermes just when I think I understand this library it beats me over the head.

I'm not sure where to go from here. My proposal of adding a no-op `lock()` to the `Storage` base class seems like the easiest way to go forward without changing the fundamental relationship of `_refresh` and `Storage`.

dhermes:
Yeah. I think you should leave this alone for now, get your original PR merged and then we discuss the plan outside of that design.

jonparrott:
The `LockedStorage` part of my original PR is the issue. Should I just merge it's capabilities into the base class in the other PR and we can continue from there? Seems like the best way forward.

dhermes:
@nathanielmanistaatgoogle WDYT?

nathanielmanistaatgoogle:
@jonparrott: yes, I think merging `LockedStorage` into the base class in the other pull request is a fine way forward for now.

PR #349, opened 2015-12-02
More informative error when failing to open secrets file.

### LINKED ISSUE #348: clientsecrets._loadfile hides native OSError
Native error messages are masked by the InvalidClientSecrets error. _loadfile naively assumes all OSErrors are the result of missing file. Permission issues present as missing file issues and can be rather frustrating to diagnose.

dhermes:
Fixes #348.


dhermes:
PS Sorry if tests fail. I am with a temporary machine and it was becoming cumbersome.

dhermes:
@waprin You seem to know more about Django than I do :grinning: 

It seems 1.8.7 was release on Nov. 24 (2015) and is causing some test failures which weren't previously an issue.

(Though in Python 2.6, the only error is my use of `self.assertRaises` as a context manager.)

waprin:
@dhermes https://github.com/google/oauth2client/pull/350 adds a commit fixes the app isue for py27 /Django 1.9. It overlaps with the Django decorator PR which is still to be merged. The fact that django_orm did slightly hacky things to setup the Django env has caused these annoyances. Nathaniel seems skeptical of using the lots-of-files Django convention in the PR but I think these type of problems makes a case for just following Django conventions when possible (and might be easier to swallow once its in contrib package). 

As for your 2.6 problem, sounds like you just need to just not use assertRaises (also, surprised this library still needs to support 2.6). Given Django isn't supporting 3.3, maybe we should setup the tox environment to not run the Django tests in the 3.3 testenv, but I'm still a bit confused because the Django code looks like it does a version check and backports the feature. Investigating now.

dhermes:
Thanks.

PR #353, opened 2015-12-04
Move non-core modules to contrib.

theacodes:
Fixes #352

PR #355, opened 2015-12-04
Move tests that only apply to multistore_file out of test_file.py

theacodes:
Requisite for #353

### DIFF
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -32,9 +32,6 @@ from .http_mock import HttpMockSequence
 import six
 
 from oauth2client import file
-from oauth2client import locked_file
-from oauth2client import multistore_file
-from oauth2client import util
 from oauth2client.client import AccessTokenCredentials
 from oauth2client.client import OAuth2Credentials
 from six.moves import http_client
@@ -46,7 +43,8 @@ except:  # pragma: NO COVER
 
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 
-FILENAME = tempfile.mktemp('oauth2client_test.data')
+_filehandle, FILENAME = tempfile.mkstemp('oauth2client_test.data')
+os.close(_filehandle)
 
 
 class OAuth2ClientFileTests(unittest.TestCase):
@@ -63,8 +61,8 @@ class OAuth2ClientFileTests(unittest.TestCase):
         except OSError:
             pass
 
-    def create_test_credentials(self, client_id='some_client_id',
-                                expiration=None):
+    def _create_test_credentials(self, client_id='some_client_id',
+                                 expiration=None):
         access_token = 'foo'
         client_secret = 'cOuDdkfjxxnv+'
         refresh_token = '1/0/a.df219fjls0'
@@ -98,7 +96,7 @@ class OAuth2ClientFileTests(unittest.TestCase):
 
     def test_pickle_and_json_interop(self):
         # Write a file with a pickled OAuth2Credentials.
-        credentials = self.create_test_credentials()
+        credentials = self._create_test_credentials()
 
         f = open(FILENAME, 'wb')
         pickle.dump(credentials, f)
@@ -122,7 +120,7 @@ class OAuth2ClientFileTests(unittest.TestCase):
     def test_token_refresh_store_expired(self):
         expiration = (datetime.datetime.utcnow() -
                       datetime.timedelta(minutes=15))
-        credentials = self.create_test_credentials(expiration=expiration)
+        credentials = self._create_test_credentials(expiration=expiration)
 
         s = file.Storage(FILENAME)
         s.put(credentials)
@@ -145,7 +143,7 @@ class OAuth2ClientFileTests(unittest.TestCase):
         # from the store expires before the original request succeeds.
         expiration = (datetime.datetime.utcnow() +
                       datetime.timedelta(minutes=15))
-        credentials = self.create_test_credentials(expiration=expiration)
+        credentials = self._create_test_credentials(expiration=expiration)
 
         s = file.Storage(FILENAME)
         s.put(credentials)
@@ -174,7 +172,7 @@ class OAuth2ClientFileTests(unittest.TestCase):
     def test_token_refresh_good_store(self):
         expiration = (datetime.datetime.utcnow() +
                       datetime.timedelta(minutes=15))
-        credentials = self.create_test_credentials(expiration=expiration)
+        credentials = self._create_test_credentials(expiration=expiration)
 
         s = file.Storage(FILENAME)
         s.put(credentials)
@@ -189,7 +187,7 @@ class OAuth2ClientFileTests(unittest.TestCase):
     def test_token_refresh_stream_body(self):
         expiration = (datetime.datetime.utcnow() +
                       datetime.timedelta(minutes=15))
-        credentials = self.create_test_credentials(expiration=expiration)
+        credentials = self._create_test_credentials(expiration=expiration)
 
         s = file.Storage(FILENAME)
         s.put(credentials)
@@ -219,7 +217,7 @@ class OAuth2ClientFileTests(unittest.TestCase):
         self.assertEqual(credentials.access_token, valid_access_token)
 
     def test_credentials_delete(self):
-        credentials = self.create_test_credentials()
+        credentials = self._create_test_credentials()
 
         s = file.Storage(FILENAME)
         s.put(credentials)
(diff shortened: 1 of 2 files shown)


tests/test_file.py
```diff
@@ -241,179 +238,12 @@ def test_access_token_credentials(self):
 
         self.assertNotEquals(None, credentials)
         self.assertEquals('foo', credentials.access_token)
-        mode = os.stat(FILENAME).st_mode
+        os.stat(FILENAME).st_mode
```
dhermes: This was an unused variable clean up? It is actually used below on line 248. Previous maintainer probably made a mistake?

To have failures instead of errors, suggest

``` python
self.assertTrue(os.path.exists(FILENAME))

if os.name == 'posix':
    mode = os.stat(FILENAME).st_mode
    self.assertEquals('0o600', oct(stat.S_IMODE(mode)))
```

---

PS This is along the lines of "make it better if you must touch it" rather than you being responsible for something you did. Like a living breathing fixit.

tests/test_multistore_file.py
```diff
@@ -1,12 +1,19 @@
...
+import stat
 import tempfile
 import unittest
 
+from oauth2client import util
+from oauth2client.client import OAuth2Credentials
+from oauth2client import locked_file
 from oauth2client import multistore_file
 
+FILENAME = tempfile.mktemp('oauth2client_test.data')
```
dhermes: IIUC `mkstemp` is preferred.

tests/test_multistore_file.py
```diff
@@ -27,6 +34,32 @@ def filename(self):
 
 
 class MultistoreFileTests(unittest.TestCase):
+    def tearDown(self):
```
dhermes, line 39: I saw @waprin remove the space between class-def and first method.

Is that a thing? I've always left a space in the presence or absence of a docstring. No need to change, I'm just curious about why do one way or the other.

tests/test_multistore_file.py
```diff
@@ -27,6 +34,32 @@ def filename(self):
...
+        try:
+            os.unlink(FILENAME)
+        except OSError:
+            pass
+
+    def setUp(self):
```
dhermes, line 45: `setUp` and `tearDown` are new (i.e. not ported directly).

Why are they introduced?

tests/test_multistore_file.py
```diff
@@ -27,6 +34,32 @@ def filename(self):
...
+        try:
+            os.unlink(FILENAME)
+        except OSError:
+            pass
+
+    def create_test_credentials(self, client_id='some_client_id',
```
dhermes: Since a helper, can you start the name with an underscore? I realize the previous one was not, but again, better than we found it.

tests/test_multistore_file.py
waprin, line 39: No, I don't think it's a thing, I think the style guides leave it ambiguous and my formatter only works one way :-\ I auto-format on save and I can't figure out how to get Intellij not to do it. There is an issue on Jetbrains to let you configure it but it hasn't been implemented, I should just diasble auto-format for this project.

tests/test_multistore_file.py
```diff
@@ -41,3 +74,170 @@ def test_lock_file_raises_ioerror(self):
...
+        os.chmod(FILENAME, 0o400)
+
+        store = multistore_file.get_credential_storage(
+            FILENAME,
+            credentials.client_id,
+            credentials.user_agent,
+            ['some-scope', 'some-other-scope'])
+
+        store.put(credentials)
+        if os.name == 'posix':
```
dhermes, line 93: I don't understand why the `posix` checks were necessary. Do you?

tests/test_multistore_file.py
```diff
@@ -41,3 +74,170 @@ def test_lock_file_raises_ioerror(self):
...
+            SYMFILENAME = FILENAME + 'sym'
+            os.symlink(FILENAME, SYMFILENAME)
+            store = multistore_file.get_credential_storage(
+                SYMFILENAME,
+                'some_client_id',
+                'user-agent/1.0',
+                ['some-scope', 'some-other-scope'])
+            try:
+                store.get()
+                self.fail('Should have raised an exception.')
```
dhermes: Again hygiene, can you use `self.assertRaises` instead of `self.fail` with the catch?

I think you'll still want to keep the `try/finally` so you can unlink the symlinked file.

tests/test_multistore_file.py
```diff
@@ -41,3 +74,170 @@ def test_lock_file_raises_ioerror(self):
...
+
+        store.put(credentials)
+        stored_credentials = store.get()
+
+        self.assertNotEquals(None, stored_credentials)
+        self.assertEqual(credentials.access_token,
+                         stored_credentials.access_token)
+
+        # try retrieving with a dictionary
+        multistore_file.get_credential_storage_custom_string_key(
```
dhermes, line 181: Ditched `store_dict` since it wasn't used?

tests/test_multistore_file.py
theacodes: Fixed.

tests/test_multistore_file.py
```diff
@@ -41,3 +74,170 @@ def test_lock_file_raises_ioerror(self):
...
+        self.assertNotEquals(None, stored_credentials)
+        self.assertEqual(credentials.access_token,
+                         stored_credentials.access_token)
+
+        store.delete()
+        stored_credentials = store.get()
+
+        self.assertEquals(None, stored_credentials)
+
+    def test_multistore_file_backwards_compatibility(self):
```
dhermes, line 193: What purpose do we have to provide backwards compat?

Doing a `git blame` for the `backwards` compat goes back to 34c1ff543dd16edf58dcf5b336076cf27de3721a

Before then it was last touched in e2233cd7a27247dbad0d06cc25972bd4bec03d14 where it was introduced when the multistore implementation changed.

tests/test_file.py
theacodes: Yeah, I was just trying to keep my linter from yelling at me. Fixed to your suggestion.

(Also, I try to avoid doing much more than that only legacy code unless it's my intention. Sort of like not boiling the ocean, but I'll gladly apply heat when you want).

tests/test_multistore_file.py
theacodes, line 39: Fixed.

dhermes:
OK finished my review.

tests/test_multistore_file.py
theacodes, line 45: They're copied from `test_file`

tests/test_multistore_file.py
theacodes: Fixed in both test files.

tests/test_file.py
dhermes: Ha I agree with and appreciate the sentiment.

tests/test_multistore_file.py
theacodes, line 93: probably because non-posix doesn't return the same permission octals. Not sure though, because I know about as much about windows as I do taxidermy.

tests/test_multistore_file.py
theacodes, line 181: Yes, just avoiding linter errors.

tests/test_file.py
```diff
@@ -46,7 +43,7 @@
 
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 
-FILENAME = tempfile.mktemp('oauth2client_test.data')
+_, FILENAME = tempfile.mkstemp('oauth2client_test.data')
```
dhermes: Change to

``` python
_filehandle, FILENAME = tempfile.mkstemp('oauth2client_test.data')
os.close(_filehandle)
```

Recommended both by @tseaver and @thobrla

tests/test_file.py
```diff
@@ -63,8 +60,8 @@ def setUp(self):
         except OSError:
             pass
 
-    def create_test_credentials(self, client_id='some_client_id',
-                                expiration=None):
+    def _create_test_credentials(self, client_id='some_client_id',
```
dhermes, line 64: Happy you did this rename but I only was talking about the multistore tests.

tests/test_multistore_file.py
theacodes, line 193: No idea, are you suggesting we remove this?

FWIW, multistore_file is an extremely complicated module in this library, I'm not sure I understand its use case or implementation.

tests/test_multistore_file.py
```diff
@@ -1,12 +1,19 @@
...
+import stat
 import tempfile
 import unittest
 
+from oauth2client import util
+from oauth2client.client import OAuth2Credentials
+from oauth2client import locked_file
 from oauth2client import multistore_file
 
+_, FILENAME = tempfile.mkstemp('oauth2client_test.data')
```
dhermes: Ditto here with closing the file handle.

tests/test_multistore_file.py
dhermes, line 45: OK cool. Thanks (and sorry for never peeking at the whole source)

tests/test_multistore_file.py
dhermes, line 39: For some definition of the word fixed. Thanks for the context.

tests/test_multistore_file.py
theacodes: Fixed.

tests/test_multistore_file.py
dhermes, line 93: Haha. You can always create a Windows VM on Compute to check. (Not necessary now, this is someone else's legacy.)

tests/test_file.py
theacodes: Fixed.

tests/test_file.py
theacodes, line 64: Ah, well, I didn't see any reason not to.

tests/test_multistore_file.py
theacodes: Fixed.

tests/test_multistore_file.py
dhermes, line 193: It's fine I suppose, but unnecessary since people will have moved on from the old implementation (changed January 24, 2013, when I was still a ~~Googler~~ full-time Googler, aka futoogler). Luckily e2233cd7a27247dbad0d06cc25972bd4bec03d14 isn't very complicated.

`vanity` tells me it was many ~~moons~~ releases ago:

``` bash
$ vanity oauth2client
...
                  oauth2client-1.0.zip    2012-09-12        34947
               oauth2client-1.0.tar.gz    2012-09-12         9417
                  oauth2client-1.1.zip    2013-03-07         9189
               oauth2client-1.1.tar.gz    2013-03-07        24544
...
```

tests/test_file.py
dhermes, line 64: Totally. (Tubular!)

tests/test_multistore_file.py
dhermes, line 181: Tis good. Soon the linter and PEP8 will be turned on in `tox`

jonparrott:
@dhermes anything left on this one?

dhermes:
Need to resolve question of removing backwards compat tests. You can punt on it and file an issue?

jonparrott:
Let's punt it. I don't think this is the place to start removing things.

dhermes:
OK. LGTM if all comments have been resolved (I'm 99% sure they have, but please confirm)

Once you've confirmed, I or @nathanielmanistaatgoogle can merge it

jonparrott:
Looks like it, with the exception of a small comment about the use of the posix checks.

dhermes:
Good call. Can you file an issue?

PR #350, opened 2015-12-03
More informative error when failing to open secrets file.

waprin:
Django 1.9 should work so py27 should work at least.

Fixes #348.

PR #356, opened 2015-12-05
Improved unit-test line coverage and quality.

dhermes:
In particular
- Upgraded some unit test modules to `unittest2`.
- Got to 100% line coverage in `test_jwt`, `test_clientsecrets`,
  and `test_service_account` by using `self.assertRaises`
  instead of `self.fail()`
- In some places used `self.assertRaises` as a context manager so
  the thrown exception can be inspected / compared against. This
  is done instead of manually catching the exception and
  holding on to it (`unittest2` backports this feature from
  `unittest` that was not available in Python 2.6)
- Added license to `test_multistore_file`
- Removed legacy `python2.4` hashbang from `test_jwt` and
  `test_service_account`
- Added `unittest2` as a test dependency in `tox.ini`
- Updated `test_clientsecrets.test_load_by_filename_missing_file`
  to use the `self.assertRaises` context manager (this was
  a problem in #349 and #350)
- Updated `test_jwt` module docstring to reflect actual purpose

### DIFF
--- a/tests/test_clientsecrets.py
+++ b/tests/test_clientsecrets.py
@@ -18,7 +18,7 @@ import errno
 from io import StringIO
 import os
 import tempfile
-import unittest
+import unittest2
 
 from oauth2client._helpers import _from_bytes
 from oauth2client import GOOGLE_AUTH_URI
@@ -37,7 +37,7 @@ NONEXISTENT_FILE = os.path.join(
     os.path.dirname(__file__), 'afilethatisntthere.json')
 
 
-class Test__validate_clientsecrets(unittest.TestCase):
+class Test__validate_clientsecrets(unittest2.TestCase):
 
     def test_with_none(self):
         self.assertRaises(clientsecrets.InvalidClientSecretsError,
@@ -157,7 +157,7 @@ class Test__validate_clientsecrets(unittest.TestCase):
         self.assertEqual(result, (clientsecrets.TYPE_INSTALLED, client_info))
 
 
-class Test__loadfile(unittest.TestCase):
+class Test__loadfile(unittest2.TestCase):
 
     def test_success(self):
         client_type, client_info = clientsecrets._loadfile(VALID_FILE)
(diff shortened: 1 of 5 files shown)


dhermes:

tests/test_jwt.py
```diff
@@ -1,6 +1,4 @@
-#!/usr/bin/python2.4
-#
-# Copyright 2014 Google Inc. All rights reserved.
+# Copyright 2015 Google Inc. All rights reserved.
```
nathanielmanistaatgoogle: Don't "update" copyright years in license headers. New files should always get the year of their introduction, but existing license headers should not be altered.

tests/test_clientsecrets.py
```diff
@@ -210,32 +210,33 @@ def test_validate_error(self):
...
             # Test load(s)
-            try:
+            with self.assertRaises(
+                    clientsecrets.InvalidClientSecretsError) as exc_manager:
                 clientsecrets.loads(src)
-                self.fail(src + ' should not be a valid client_secrets file.')
-            except clientsecrets.InvalidClientSecretsError as e:
-                self.assertTrue(str(e).startswith(match))
+
+            first_caught_exception = exc_manager.exception
```
nathanielmanistaatgoogle: No need for a local field; use `exc_manager.exception` everywhere you wish now that you have it available for use.

(And analogous other places too.)

tests/test_jwt.py
dhermes: Ahh I see. I only did it because I was there. Will revert (it'll make the diff look nicer)

tests/test_clientsecrets.py
dhermes: SGTM

dhermes:
@nathanielmanistaatgoogle Changes addressed. PTAL

nathanielmanistaatgoogle:
Looks good; merge when tests pass for the edited commit.

dhermes:
Thanks!

PR #360, opened 2015-12-09
Adds a default value for run_flow's flags argument.

Tyler-Murphy:
This makes it easier to to use run_flow in other applications where the default flags are acceptable.

I think I've already signed a CLA, but let me know if this is mergeable and I haven't.

### DIFF
--- a/oauth2client/tools.py
+++ b/oauth2client/tools.py
@@ -112,7 +112,7 @@ class ClientRedirectHandler(BaseHTTPServer.BaseHTTPRequestHandler):
 
 
 @util.positional(3)
-def run_flow(flow, storage, flags, http=None):
+def run_flow(flow, storage, flags=argparser.parse_args(), http=None):
     """Core code for a command-line application.
 
     The ``run()`` function is called from your application and runs


nathanielmanistaatgoogle:
I'm not yet sure whether or not this is a change that should be made, but if it were: never make a function call in an expression being used as the default value of a function argument, because the expression is evaluated at function **definition** time, not function **call** time. Default values of arguments should almost always be `None`.

Tyler-Murphy:
I want it to be evaluated when the function is defined. The arguments already added to `argparser` when _CreateArgumentParser is called are the defaults I'm going for.

Is it better as something like the following?

```python
argparser = _CreateArgumentParser()
default_flags = argparser.parse_args()
```

where `default_flags` is used instead?

Or maybe this change doesn't belong here, since I'm not using the methods in tools.py on the command line.

oauth2client/tools.py
```diff
@@ -112,7 +112,7 @@ def log_message(self, format, *args):
 
 
 @util.positional(3)
-def run_flow(flow, storage, flags, http=None):
+def run_flow(flow, storage, flags=argparser.parse_args(), http=None):
```
dhermes, line 115: `@util.positional(3)` should also change if you add another conditional.

Tyler-Murphy, line 115: To `@util.positional(2)`?

dhermes, line 115: Yes, though I'm not endorsing this change or your use of a default like this.

I'm also not 100% sure but seem to remember that the number of positional args was something that needed to wait for a breaking release.

dhermes:
Closing since this went stale, #426 tries to take it to the finish line.

PR #362, opened 2015-12-14
Unpinning tox version in install script.

dhermes:
Regressions from tox 2.2.0 were fixed in >=2.3.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -16,7 +16,7 @@
 
 set -ev
 
-pip install tox==2.1.1
+pip install tox
 if [[ "${TOX_ENV}" == "pypy" ]]; then
     git clone https://github.com/yyuu/pyenv.git ${HOME}/.pyenv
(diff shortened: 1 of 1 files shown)


nathanielmanistaatgoogle:
Completely unpinned dependencies are a project's way of saying "we would like to be broken sometime in the future at a time not of our choosing and with no advance notice". Why not just move the pin forward?

dhermes:
But using someone else's library is a way of saying "we are a part of an open-source eco-system" and by using the latest version is a way of telling our peers "we want to help you kick the tires of your library and are excited about new features you release"

----

In other words, I disagree. I think using the latest version on PyPI allows us to upgrade automatically when it does smoothly and quickly detect breakages when it does not.

nathanielmanistaatgoogle:
Okay.

dhermes:
Haha that was easy :grinning: Thanks for the merge

PR #363, opened 2015-12-15
Fixing tests to ignore app engine.

theacodes:


dhermes:
I suppose the reason we didn't notice from CI is because the GAE coverages are under 100% so it probably even **increased** the global average coverage. All the more reason to get to 100% and have Travis fail when we don't reach it.

jonparrott:
> I suppose the reason we didn't notice from CI is because the GAE coverages are under 100% so it probably even increased the global average coverage. All the more reason to get to 100% and have Travis fail when we don't reach it.

Haha, brilliant. And I didn't catch it because I was running `tox -e py27,py34` instead of `cover`. Yay me.

dhermes:
@jonparrott The goal is to set the "right" default `tox` envs so that running `tox` without the `-e` flag before submitting is "nice".

Since `gcloud-python` takes awhile, I actually ignore the tab and make it tell me when it's done:

```
tox; espeak "DONE WITH STATUS $?"
```

jonparrott:
@dhermes oh yeah, absolutely, I'm at fault here.

dhermes:
Nobody's pointing fingers :grinning:

PR #370, opened 2016-01-04
Moving gce module to contrib.

dhermes:
Also fixing a broken import of `oauth2client.appengine` in `oauth2client.client`. This reveals an untested code-path that could have been found via 100% code coverage. (See #212)

---

Remaining modules not in `contrib`:

```
oauth2client/client.py
oauth2client/clientsecrets.py
oauth2client/crypt.py
oauth2client/devshell.py
oauth2client/file.py
oauth2client/_helpers.py
oauth2client/__init__.py
oauth2client/_openssl_crypt.py
oauth2client/_pycrypto_crypt.py
oauth2client/service_account.py
oauth2client/tools.py
oauth2client/util.py
oauth2client/xsrfutil.py
```

@waprin @jonparrott Do you think any of these should be moved to contrib? It seems that `devshell.py`, `file.py` and `xsrfutil.py` (only user is the `appengine` module) should go? BTW who is using `devshell`?


waprin:
I see one internal usage of devshell, I feel like it might be in here only because whoever does the migration to Github just included all the files. But I'm honestly not sure and not totally clear on how that migration process works. Either way your suggested changes all make sense to me.

jonparrott:
`file.py` should stay in core, I think. 

Everything else SGTM.

dhermes:
@nathanielmanistaatgoogle should I send another PR to move `xsrfutil` and `devshell` or shall I re-purpose this PR?

nathanielmanistaatgoogle:
Squash commits on this pull request, I think?

Let's do `xsrfutil` and `devshell` in a separate pull request.

dhermes:
Squashing is fine with me. Though I intentionally did it in two commits.

Good to merge once squashed?

nathanielmanistaatgoogle:
Looks good; please merge when tests repass.

PR #371, opened 2016-01-05
Factor out usage of utcnow() in client.

dhermes:
This is to enable better stubs in testing and eliminate two `sleep()` statements in unit tests. (The philosophy is "unit tests should be fast".)

Towards #357.

PR #372, opened 2016-01-05
Adds token_expiry to the Devshell credential.

### LINKED ISSUE #373: Devshell Credential does not set the token_expiry field
The credential generated in the Devshell environment does not set the token_expiry field.  This means you cannot determine how long the credential is valid for. 

The serialized `CredentialInfoResponse` proto contains a field for `expires_in`, so that should be translated into the `token_expiry` field.

wilford:
Fixes #373

### DIFF
--- a/oauth2client/devshell.py
+++ b/oauth2client/devshell.py
@@ -14,12 +14,17 @@
 
 """OAuth 2.0 utitilies for Google Developer Shell environment."""
 
+import datetime
 import json
 import os
 import socket
 
 from oauth2client._helpers import _to_bytes
 from oauth2client import client
+
+# Expose utcnow() at module level to allow for
+# easier testing (by replacing with a stub).
+_UTCNOW = datetime.datetime.utcnow
 
 DEVSHELL_ENV = 'DEVSHELL_CLIENT_PORT'
 
@@ -52,6 +57,7 @@ class CredentialInfoResponse(object):
     * Index 0 - user email
     * Index 1 - default project ID. None if the project context is not known.
     * Index 2 - OAuth2 access token. None if there is no valid auth context.
+    * Index 3 - Seconds until the access token expires. None if not present.
     """
 
     def __init__(self, json_string):
@@ -63,6 +69,7 @@ class CredentialInfoResponse(object):
         self.user_email = pbl[0] if pbl_len > 0 else None
         self.project_id = pbl[1] if pbl_len > 1 else None
         self.access_token = pbl[2] if pbl_len > 2 else None
+        self.expires_in = pbl[3] if pbl_len > 3 else None
 
 
 def _SendRecv():
@@ -117,6 +124,12 @@ class DevshellCredentials(client.GoogleCredentials):
     def _refresh(self, http_request):
         self.devshell_response = _SendRecv()
         self.access_token = self.devshell_response.access_token
+        expires_in = self.devshell_response.expires_in
+        if expires_in is not None:
+            delta = datetime.timedelta(seconds=expires_in)
+            self.token_expiry = _UTCNOW() + delta
+        else:
+            self.token_expiry = None
 
     @property
     def user_email(self):
--- a/tests/test_devshell.py
+++ b/tests/test_devshell.py
@@ -14,6 +14,7 @@
 
 """Tests for oauth2client.devshell."""
 
+import datetime
 import json
 import os
 import socket
@@ -22,6 +23,7 @@ import unittest
 
 import mock
 
+from oauth2client import devshell
 from oauth2client._helpers import _from_bytes
 from oauth2client._helpers import _to_bytes
 from oauth2client.client import save_to_well_known_file
@@ -32,6 +34,16 @@ from oauth2client.devshell import CredentialInfoResponse
 from oauth2client.devshell import DEVSHELL_ENV
 from oauth2client.devshell import DevshellCredentials
 from oauth2client.devshell import NoDevshellServer
+
+# A dummy value to use for the expires_in field
+# in CredentialInfoResponse.
+EXPIRES_IN = 1000
+DEFAULT_CREDENTIAL_JSON = json.dumps([
+    'joe@example.com',
+    'fooproj',
+    'sometoken',
+    EXPIRES_IN
+])
 
 
 class TestCredentialInfoResponse(unittest.TestCase):
@@ -51,16 +63,20 @@ class TestCredentialInfoResponse(unittest.TestCase):
         self.assertEqual(info_response.user_email, None)
         self.assertEqual(info_response.project_id, None)
         self.assertEqual(info_response.access_token, None)
+        self.assertEqual(info_response.expires_in, None)
 
     def test_constructor_full_list(self):
         user_email = 'user_email'
         project_id = 'project_id'
         access_token = 'access_token'
-        json_string = json.dumps([user_email, project_id, access_token])
+        expires_in = 1
+        json_string = json.dumps(
+            [user_email, project_id, access_token, expires_in])
         info_response = CredentialInfoResponse(json_string)
         self.assertEqual(info_response.user_email, user_email)
         self.assertEqual(info_response.project_id, project_id)
         self.assertEqual(info_response.access_token, access_token)
+        self.assertEqual(info_response.expires_in, expires_in)
 
 
 class Test_SendRecv(unittest.TestCase):
@@ -106,8 +122,7 @@ class _AuthReferenceServer(threading.Thread):
 
     def __init__(self, response=None):
         super(_AuthReferenceServer, self).__init__(None)
-        self.response = (response or
-                         '["joe@example.com", "fooproj", "sometoken"]')
+        self.response = response or DEFAULT_CREDENTIAL_JSON
         self.bad_request = False
 
     def __enter__(self):
@@ -195,19 +210,27 @@ class DevshellCredentialsTests(unittest.TestCase):
             creds = DevshellCredentials()
             self.assertEquals(None, creds.refresh_token)
 
-    def test_reads_credentials(self):
+    @mock.patch.object(devshell, '_UTCNOW')
+    def test_reads_credentials(self, utcnow):
+        NOW = datetime.datetime(1992, 12, 31)
+        utcnow.return_value = NOW
         with _AuthReferenceServer():
             creds = DevshellCredentials()
             self.assertEqual('joe@example.com', creds.user_email)
             self.assertEqual('fooproj', creds.project_id)
             self.assertEqual('sometoken', creds.access_token)
-
+            self.assertEqual(
+                NOW + datetime.timedelta(seconds=EXPIRES_IN),
+                creds.token_expiry)
+            utcnow.assert_called_once_with()
+            
     def test_handles_skipped_fields(self):
         with _AuthReferenceServer('["joe@example.com"]'):
             creds = DevshellCredentials()
             self.assertEqual('joe@example.com', creds.user_email)
             self.assertEqual(None, creds.project_id)
             self.assertEqual(None, creds.access_token)
+            self.assertEqual(None, creds.token_expiry)
 
     def test_handles_tiny_response(self):
         with _AuthReferenceServer('[]'):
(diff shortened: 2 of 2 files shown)


oauth2client/devshell.py
```diff
@@ -52,6 +53,7 @@ class CredentialInfoResponse(object):
     * Index 0 - user email
     * Index 1 - default project ID. None if the project context is not known.
     * Index 2 - OAuth2 access token. None if there is no valid auth context.
+    * Index 3 - Seconds till the access token expries. None if not present.
```
dhermes: - till -> until
- expries -> expires

oauth2client/devshell.py
```diff
@@ -117,6 +120,9 @@ def __init__(self, user_agent=None):
     def _refresh(self, http_request):
         self.devshell_response = _SendRecv()
         self.access_token = self.devshell_response.access_token
+        if self.devshell_response.expires_in != None:
```
dhermes: `!= None` -> `is not None`

oauth2client/devshell.py
```diff
@@ -117,6 +120,9 @@ def __init__(self, user_agent=None):
...
+          self.token_expiry = (datetime.datetime.utcnow() +
+            datetime.timedelta(seconds=self.devshell_response.expires_in))
```
dhermes: 4-space indent (PEP8)

tests/test_devshell.py
```diff
@@ -51,16 +52,19 @@ def test_constructor_empty_list(self):
...
         self.assertEqual(info_response.access_token, None)
+        self.assertEqual(info_response.expires_in, None)
 
     def test_constructor_full_list(self):
         user_email = 'user_email'
         project_id = 'project_id'
         access_token = 'access_token'
-        json_string = json.dumps([user_email, project_id, access_token])
+        expires_in = 1
+        json_string = json.dumps([user_email, project_id, access_token, expires_in])
```
dhermes: Line length longer than 80 chars

tests/test_devshell.py
```diff
@@ -202,12 +206,23 @@ def test_reads_credentials(self):
             self.assertEqual('fooproj', creds.project_id)
             self.assertEqual('sometoken', creds.access_token)
 
+            # Since we can't mock out datetime, assume that
```
dhermes: Why can't we mock out `datetime`? See #371 for an example of mocking

tests/test_devshell.py
```diff
@@ -202,12 +206,23 @@ def test_reads_credentials(self):
...
+            # The token_expiry time is within some regional
+            # range of now() + 1000.
+            sixtySecondsAfter = (datetime.datetime.utcnow() +
```
dhermes: `sixtySecondsAfter` -> `sixty_seconds_after` (var names should be PEP8)

tests/test_devshell.py
```diff
@@ -202,12 +206,23 @@ def test_reads_credentials(self):
...
+              datetime.timedelta(seconds=1060))
```
dhermes: Ditto here on 2-space indent

oauth2client/devshell.py
wilford: Done

oauth2client/devshell.py
wilford: Done.

oauth2client/devshell.py
wilford: Done.

tests/test_devshell.py
wilford: Done.

dhermes:
I don't see the updated/fixed code. Have you pushed again?

tests/test_devshell.py
wilford: Removed.

tests/test_devshell.py
wilford: Removed.

tests/test_devshell.py
wilford: Ah thanks. Done.

dhermes:
I'd also like to point out, from the [CONTRIBUTING][1] doc:

> 2) File an issue to notify the maintainers about what you're working on.
> ...
> 5) Send a pull request.

Getting a PR out of the blue with no context is optimal. We have no way of knowing why it is needed / what the context is.

Thanks for sending though! I just want to be able to have things go more smoothly in the future.

[1]: https://github.com/google/oauth2client/blob/ddc5a75ccda9b3cce4a6f4f366f4b45b9e1afa28/CONTRIBUTING.md

wilford:
Addressed reviewer comments.  Thanks!

wilford:
Ah sorry, just sent out the updated code.  Should I file an issue as to
what this addressing?

Thanks for the review.

dhermes:
Yes please. We were just discussing in #370 that we don't even how `devshell` is used, so it's very much out of context.

oauth2client/devshell.py
```diff
@@ -52,6 +57,7 @@ class CredentialInfoResponse(object):
     * Index 0 - user email
     * Index 1 - default project ID. None if the project context is not known.
     * Index 2 - OAuth2 access token. None if there is no valid auth context.
+    * Index 3 - Seconds untill the access token expires. None if not present.
```
dhermes: untill -> until

oauth2client/devshell.py
```diff
@@ -117,6 +124,9 @@ def __init__(self, user_agent=None):
     def _refresh(self, http_request):
         self.devshell_response = _SendRecv()
         self.access_token = self.devshell_response.access_token
+        if self.devshell_response.expires_in is not None:
+          self.token_expiry = (_UTCNOW() +
```
dhermes: 4-space indent inside the `if` block. (PEP8)

I'd recommend changing this whole section as follows (to make it easier on the eyes):

``` python
        expires_in = self.devshell_response.expires_in
        if expires_in is not None:
            delta = datetime.timedelta(seconds=expires_in)
            self.token_expiry = _UTCNOW() + delta
```

tests/test_devshell.py
```diff
@@ -195,19 +201,25 @@ def test_no_refresh_token(self):
...
+        NOW = datetime.datetime(1992, 12, 31)
+        utcnow.return_value = NOW
         with _AuthReferenceServer():
             creds = DevshellCredentials()
             self.assertEqual('joe@example.com', creds.user_email)
             self.assertEqual('fooproj', creds.project_id)
             self.assertEqual('sometoken', creds.access_token)
-
+            self.assertEqual(
+                NOW + datetime.timedelta(seconds=1000), creds.token_expiry)
```
dhermes: The 1000 is hard-coded both here and in the `_AuthReferenceServer()` definition. I'd suggest putting it somewhere as a global or class attribute so your code isn't so brittle.

tests/test_devshell.py
```diff
@@ -195,19 +201,25 @@ def test_no_refresh_token(self):
...
+
```
dhermes, line 226: You should check the calls to your stub. Something like

``` python
self.assertEqual(utcnow.mock_calls, [...])
```

or `utcnow.assert_called_with` or something similar. (Check out the [docs](https://docs.python.org/3/library/unittest.mock.html) for other options.)

oauth2client/devshell.py
wilford: Done.

oauth2client/devshell.py
wilford: Done.

tests/test_devshell.py
wilford: Done.

tests/test_devshell.py
wilford, line 226: Done.

tests/test_devshell.py
```diff
@@ -107,7 +116,7 @@ class _AuthReferenceServer(threading.Thread):
     def __init__(self, response=None):
         super(_AuthReferenceServer, self).__init__(None)
         self.response = (response or
-                         '["joe@example.com", "fooproj", "sometoken"]')
+            '["joe@example.com", "fooproj", "sometoken", ' + str(EXPIRES_IN) + ']')
```
dhermes: Thanks for doing the re-factor with the global, but sadly now this line is longer than 80 chars. (Sorry to be the PEP8 enforcer)

tests/test_devshell.py
```diff
@@ -195,19 +204,26 @@ def test_no_refresh_token(self):
...
+    def test_reads_credentials(self, utcnow):
+        NOW = datetime.datetime(1992, 12, 31)
+        utcnow.return_value = NOW
         with _AuthReferenceServer():
             creds = DevshellCredentials()
             self.assertEqual('joe@example.com', creds.user_email)
             self.assertEqual('fooproj', creds.project_id)
             self.assertEqual('sometoken', creds.access_token)
-
+            self.assertEqual(
```
dhermes, line 222: Line length. Instead:

``` python
            self.assertEqual(
                NOW + datetime.timedelta(seconds=EXPIRES_IN),
                creds.token_expiry)
```

tests/test_devshell.py
dhermes: At this point you'd probably be happier doing a

``` python
json.dumps([
    'joe@example.com',
    'fooproj',
    'sometoken',
    EXPIRES_IN,
])
```

somewhere.

wilford: Done.

tests/test_devshell.py
wilford, line 222: done.

dhermes:
OK this LGTM. Will wait for @nathanielmanistaatgoogle before merging.

@wilford Can you give an idea of
- what you use `devshell.py` for?
- why you noticed this change was necessary?

wilford:
devshell.py is used for retrieving Oauth2 access tokens when running inside
the Google Cloud Shell.  (You can demo the Cloud Shell by going to
https://console.cloud.google.com/ and clicking on the terminal icon in the
top right).  A user is automatically authenticated via their gaia browser
cookies when running inside the Cloud Shell, and we have a custom auth
channel for retrieving OAuth2 access tokens.  devshell.py uses this auth
channel to get these tokens when running inside the Cloud Shell environment.

I noticed this change was necessary because I wanted to introduce caching
of these tokens inside the Cloud Shell.  However, I realized we weren't
exposing the expiration times for the access tokens.

oauth2client/devshell.py
```diff
@@ -117,6 +124,10 @@ def __init__(self, user_agent=None):
     def _refresh(self, http_request):
         self.devshell_response = _SendRecv()
         self.access_token = self.devshell_response.access_token
+        expires_in = self.devshell_response.expires_in
+        if expires_in is not None:
```
nathanielmanistaatgoogle, line 128: Can `self.devshell_response.expires_in` be `None` at this point? Can `self.token_expiry` be non-`None` at this point? If both are possible, is this method correct in the way it would leave the previous non-`None` value unmolested?

wilford, line 128: As currently implemented, self.devshell_response.expires_in will not be none as we always fill in that value.  However, the value is marked as optional in the proto definition, so in theory yes it could be.

self.token_expiry can be non None if refresh() was previously called and set for the last value of the access token.  However, in the case where we get the same access token back, the expires_in value will also be less by whatever time delta has passed, so it will set it to the same token_expiry.

dhermes, line 128: @nathanielmanistaatgoogle The fact that `self.access_token` changes means we have a brand new token state.

@wilford I'd argue that `self.token_expiry = None` should occur in the case that `expires_in is None` (the other branch of this `if` statement`)

dhermes:
@nathanielmanistaatgoogle @wilford Can we push this through?

oauth2client/devshell.py
wilford, line 128: Set self.token_expiry to None in this case.

nathanielmanistaatgoogle:
Change content looks good and I've kicked the build to attempt a clean test result; to me this will be ready for merge when the test repasses.

dhermes:
Unless this has been rebased after #378 I'm fairly certain it will continue to fail.

nathanielmanistaatgoogle:
`All checks have passed` :-) Merge, or do you see further need for rebase?

dhermes:
Nope, it's fine to me.

Issue #357, opened 2015-12-05
Banish all uses of `time.sleep` in unit tests and mock out uses of sleep in tested code


dhermes:
Call me crazy, but this is a no-brainer right?

```python
$ # show the current commit
$ git log -1 HEAD --pretty=%H
2e5c7cb3f0631329dfd28f64edd4ef827df3d534
$ git grep 'time\.sleep'
oauth2client/locked_file.py:                time.sleep(delay)
oauth2client/locked_file.py:                    time.sleep(delay)
oauth2client/locked_file.py:                    time.sleep(delay)
tests/test_client.py:        time.sleep(S + 0.5)  # some margin to avoid flakiness
tests/test_service_account.py:        time.sleep(S + 0.5)  # some margin to avoid flakiness
```

-----

Also on this list should be any method that relies on the network being turned on. I used `nose.tools.timed(limit)` in #85 to try to find slow test cases, we should find them again and snuff them out.

pferate:
I'm thinking that this can be closed.
```
$ git grep "sleep"
oauth2client/contrib/_fcntl_opener.py:                time.sleep(delay)
oauth2client/contrib/_win32_opener.py:                time.sleep(delay)
oauth2client/contrib/locked_file.py:                time.sleep(delay)
tests/contrib/test_locked_file.py:    @mock.patch('time.sleep')
tests/contrib/test_locked_file.py:    def test_lock_retry_success(self, mock_sleep, mock_time, mock_open):
tests/contrib/test_locked_file.py:            mock_sleep.assert_called_with(1)
```
`_win32_opener.py` and `_fcntl_opener.py` are being mocked, and also may not be needed (per @jonparrott in #563).

dhermes:
SGTM though I'd prefer to leave this open until those files are actually gone

jonparrott:
Since there's already an issue open to delete those files and it's part of the same milestone, this issue is now essentially obsolete. I'm gonna close it.

PR #374, opened 2016-01-05
Using more mocks in implicit environ tests.

dhermes:
This was done mostly because the GCE environment check was taking too long to time-out.

---

Discovered while investigating #357.

---

Currently, the four updated tests run in:

```
$ nosetests \
>   tests/test_client.py:GoogleCredentialsTests.test_get_adc_from_environment_variable_service_account \
>   tests/test_client.py:GoogleCredentialsTests.test_get_adc_from_environment_variable_authorized_user \
>   tests/test_client.py:GoogleCredentialsTests.test_get_adc_from_environment_variable_malformed_file \
>   tests/test_client.py:GoogleCredentialsTests.test_get_application_default_environment_not_set_up
....
----------------------------------------------------------------------
Ran 4 tests in 1.004s

OK
```

and after this update:

```
$ nosetests \
>   tests/test_client.py:GoogleCredentialsTests.test_get_adc_from_environment_variable_service_account \
>   tests/test_client.py:GoogleCredentialsTests.test_get_adc_from_environment_variable_authorized_user \
>   tests/test_client.py:GoogleCredentialsTests.test_get_adc_from_environment_variable_malformed_file \
>   tests/test_client.py:GoogleCredentialsTests.test_get_application_default_environment_not_set_up
....
----------------------------------------------------------------------
Ran 4 tests in 0.005s

OK
```

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -101,6 +101,8 @@ ADC_HELP_MSG = (
     'https://developers.google.com/accounts/docs/'
     'application-default-credentials for more information.')
 
+_WELL_KNOWN_CREDENTIALS_FILE = 'application_default_credentials.json'
+
 # The access token along with the seconds in which it expires.
 AccessTokenInfo = collections.namedtuple(
     'AccessTokenInfo', ['access_token', 'expires_in'])
@@ -1419,9 +1421,6 @@ def _get_well_known_file():
     """Get the well known file produced by command 'gcloud auth login'."""
     # TODO(orestica): Revisit this method once gcloud provides a better way
     # of pinpointing the exact location of the file.
-
-    WELL_KNOWN_CREDENTIALS_FILE = 'application_default_credentials.json'
-
     default_config_dir = os.getenv(_CLOUDSDK_CONFIG_ENV_VAR)
     if default_config_dir is None:
         if os.name == 'nt':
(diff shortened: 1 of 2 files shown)


nathanielmanistaatgoogle:
Don't the tests now run faster because they're executing less code? It looks like without this change the tests go all the way "through" `oauth2client` and hit the "test circumstances" "below" `oauth2client`, but with this change the mocking is done within `oauth2client` and the tests execute less production code than before.

Sorry for all the weird quotes; I don't talk about this sort of thing often so I don't have a strong vocabulary of concepts for it.

Since I know you're minding test coverage I know that you haven't uncovered any production code, but isn't there an argument to be made for redundant and overlapping test coverage?

How easily could I propose a guideline that `oauth2client`'s tests should never mock out any part of `oauth2client`? One should never mock out any part of the system under test, right? So then it becomes a philosophical question of the extent to which `test_module.py` tests `module.py` and the extent to which it tests `oauth2client`, right?

dhermes:
> tests should never mock out any part of `oauth2client`

I disagree. There should be dedicated places to test a function, class or method. If another function, class or method relies on an **already tested** component, there is no need to re-test that part. **Especially** in cases where we go 3 or 4 layers deep. At that point, the mocks could begin to be larger than the actual testing code (or even the code under test).

nathanielmanistaatgoogle:
I'd like to persuade you otherwise at some point but won't make an obstacle of it today. :-) Merging.

dhermes:
Thanks.

PR #375, opened 2016-01-05
Moving xsrfutil module to contrib.

dhermes:
The only consumer of this utility is `oauth2client.contrib.appengine`. See discussion in #370 for reason of move.


nathanielmanistaatgoogle:
Looks good; please merge when tests pass.

Issue #298, opened 2015-09-03
Drop support for Python 2.6


dhermes:
See https://github.com/GoogleCloudPlatform/gcloud-python/issues/995 and https://github.com/google/oauth2client/pull/297 for discussion

craigcitro:
@thobrla

thobrla:
gsutil still supports 2.6 and depends on this

nathanielmanistaatgoogle:
@thobrla: what's gsutil's timeline for dropping 2.6 support?

thobrla:
There's no timeline.  We still get regular requests and reports from users on 2.6.

nathanielmanistaatgoogle:
:-(

dhermes:
@thobrla I recommend reading http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html (Nick even weighs in on our thread over in `gcloud-python`)

dhermes:
More support: `keyring` and `django` flat out don't work in Python 2.6.

nathanielmanistaatgoogle:
@thobrla: today is the twelfth day of Christmas; how possible would it be for us to make dropping Python 2.6 support the gift we give ourselves? Has anything changed on your side of the relationship? What if you had to pin your dependency on us to a less-than-current version?

thobrla:
The situation hasn't changed much - gsutil still has a substantial amount of users using Python 2.6.  If we had to pin on a less-than-current version that would work up until the point that we actually needed a fix/update in oauth2client, which I anticipate will happen at least once in the coming year, if not more.

That being said, I recognize we can't support this forever.  I'll discuss with our dev team and get back to you here.

nathanielmanistaatgoogle:
@thobrla: ping?

thobrla:
I'm still working on details with the gcloud CLI team, and probably won't have a fully-publishable plan for at least a few weeks.  The general consensus is that we'd like to deprecate support, but need to decide on the length of the deprecation period and migration plan.

thobrla:
FYI: gsutil announced deprecation for Python 2.6 today, with support ending on September 1, 2016.

nathanielmanistaatgoogle:
So... are we good to just hang out until 1 September? Is there anything else that would ask 2.6 support of us after then?

PR #376, opened 2016-01-05
Moving devshell module to contrib.

dhermes:
See discussion in #370 for context.


dhermes:
I should hold off on doing this until #372 is in.

dhermes:
@nathanielmanistaatgoogle It looks like a brand new release of `keyring` no longer works in Python 2.6:
https://travis-ci.org/google/oauth2client/builds/100392153

Luckily `keyring_storage` is now in `contrib`, so we can make it optional (and leave the `keyring` tests out of the Python 2.6 `tox` environment).

More ammo for #298.

dhermes:
@nathanielmanistaatgoogle I rebased after #372 got it.

nathanielmanistaatgoogle:
Change looks good; merge when tests repass.

PR #378, opened 2016-01-05
Removing keyring testing from py26 testing.

dhermes:
With the `6.0` release on January 5, 2016, the `keyring` package no longer supports Python 2.6 (it won't even install).

@nathanielmanistaatgoogle This was discovered in #376.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,8 +2,7 @@
 envlist = py26,py27,py33,py34,pypy,gae,cover
 
 [testenv]
-basedeps = keyring
-           mock>=1.3.0
+basedeps = mock>=1.3.0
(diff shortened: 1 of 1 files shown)


nathanielmanistaatgoogle:
Looks good; merge on test passage.

dhermes:
@nathanielmanistaatgoogle I killed the two merge Travis jobs so that this one would run faster

https://travis-ci.org/google/oauth2client/builds/100442900
https://travis-ci.org/google/oauth2client/builds/100442958

It seems Travis is only running one job at a time for us (possibly because the use of the environment matrix get run on separate VMs)

PR #382, opened 2016-01-07
Updating setup.py and MANIFEST to include all subdirs.

dhermes:
Inspired by #364. Wanted to get it in before it falls off the radar.

H/T to @trcook for bringing the problem to our attention.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include README.md
+recursive-exclude tests *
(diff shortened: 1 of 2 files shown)


nathanielmanistaatgoogle:
Change content looks good; are you inclined to include any more of the history or motivation in your commit message? If not, merge at your leisure.

dhermes:
No I'm not inclined to do so. The motivation should be obvious from the content: we left out directories in our source tree (and we had a hard-coded packages list, which is overall a bad idea).

Issue #384, opened 2016-01-15
GoogleCredentials and _ServiceAccountCredentials classes are missing to_json/from_json methods


dlorenc:
This prevents these classes from being serialized and used by things like oauth2client.multistore_file, which caches serialized credentials.

multistore_file is used by gsutil, so this is blocking switching gsutil to GoogleCredentials.from_stream.

PR #385, opened 2016-01-19
Add to/from json methods to Credentials classes

dlorenc:
This adds to_json and from_json methods to GoogleCredentials
and _ServiceAccountCredentials classes. This allows them to
be serialized by multistore_file.

Resovles: #384

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -198,6 +198,13 @@ class MemoryCache(object):
         self.cache.pop(key, None)
 
 
+def _parse_expiry(expiry):
+    if expiry and isinstance(expiry, datetime.datetime):
+        return expiry.strftime(EXPIRY_FORMAT)
+    else:
+        return None
+
+
 class Credentials(object):
     """Base class for all Credentials objects.
 
@@ -208,7 +215,7 @@ class Credentials(object):
     JSON string as input and returns an instantiated Credentials object.
     """
 
-    NON_SERIALIZED_MEMBERS = ['store']
+    NON_SERIALIZED_MEMBERS = frozenset(['store'])
 
     def authorize(self, http):
         """Take an httplib2.Http instance (or equivalent) and authorizes it.
@@ -265,9 +272,7 @@ class Credentials(object):
         for member in strip:
             if member in d:
                 del d[member]
-        if (d.get('token_expiry') and
-                isinstance(d['token_expiry'], datetime.datetime)):
-            d['token_expiry'] = d['token_expiry'].strftime(EXPIRY_FORMAT)
+        d['token_expiry'] = _parse_expiry(d.get('token_expiry'))
         # Add in information we will need later to reconsistitue this instance.
         d['_class'] = t.__name__
         d['_module'] = t.__module__
@@ -285,7 +290,7 @@ class Credentials(object):
             string, a JSON representation of this instance, suitable to pass to
             from_json().
         """
-        return self._to_json(Credentials.NON_SERIALIZED_MEMBERS)
+        return self._to_json(self.NON_SERIALIZED_MEMBERS)
 
     @classmethod
     def new_from_json(cls, s):
@@ -698,9 +703,6 @@ class OAuth2Credentials(Credentials):
         """
         self._retrieve_scopes(http.request)
         return self.scopes
-
-    def to_json(self):
-        return self._to_json(Credentials.NON_SERIALIZED_MEMBERS)
 
     @classmethod
     def from_json(cls, s):
@@ -1180,6 +1182,11 @@ class GoogleCredentials(OAuth2Credentials):
         print(response)
     """
 
+    NON_SERIALIZED_MEMBERS =  (
+        frozenset(['_private_key']) |
+        OAuth2Credentials.NON_SERIALIZED_MEMBERS)
+
+
     def __init__(self, access_token, client_id, client_secret, refresh_token,
                  token_expiry, token_uri, user_agent,
                  revoke_uri=GOOGLE_REVOKE_URI):
@@ -1221,6 +1228,32 @@ class GoogleCredentials(OAuth2Credentials):
         The Credentials type is preserved.
         """
         return self
+
+    @classmethod
+    def from_json(cls, s):
+        # TODO(issue 388): eliminate the circularity that is the reason for
+        # this non-top-level import.
+        from oauth2client.service_account import _ServiceAccountCredentials
+        data = json.loads(_from_bytes(s))
+
+        # We handle service_account._ServiceAccountCredentials since it is a
+        # possible return type of GoogleCredentials.get_application_default()
+        if (data['_module'] == 'oauth2client.service_account' and
+            data['_class'] == '_ServiceAccountCredentials'):
+            return _ServiceAccountCredentials.from_json(s)
+
+        token_expiry = _parse_expiry(data.get('token_expiry'))
+        google_credentials = cls(
+            data['access_token'],
+            data['client_id'],
+            data['client_secret'],
+            data['refresh_token'],
+            token_expiry,
+            data['token_uri'],
+            data['user_agent'],
+            revoke_uri=data.get('revoke_uri', None))
+        google_credentials.invalid = data['invalid']
+        return google_credentials
 
     @property
     def serialization_data(self):
--- a/oauth2client/service_account.py
+++ b/oauth2client/service_account.py
@@ -18,6 +18,8 @@ This credentials class is implemented on top of rsa library.
 """
 
 import base64
+import datetime
+import json
 import time
 
 from pyasn1.codec.ber import decoder
@@ -27,16 +29,23 @@ import rsa
 from oauth2client import GOOGLE_REVOKE_URI
 from oauth2client import GOOGLE_TOKEN_URI
 from oauth2client._helpers import _json_encode
+from oauth2client._helpers import _from_bytes
 from oauth2client._helpers import _to_bytes
 from oauth2client._helpers import _urlsafe_b64encode
 from oauth2client import util
 from oauth2client.client import AssertionCredentials
+from oauth2client.client import EXPIRY_FORMAT
 
 
 class _ServiceAccountCredentials(AssertionCredentials):
     """Class representing a service account (signed JWT) credential."""
 
     MAX_TOKEN_LIFETIME_SECS = 3600  # 1 hour in seconds
+
+    NON_SERIALIZED_MEMBERS =  (
+        frozenset(['_private_key']) |
+        AssertionCredentials.NON_SERIALIZED_MEMBERS)
+
 
     def __init__(self, service_account_id, service_account_email,
                  private_key_id, private_key_pkcs8_text, scopes,
@@ -107,6 +116,25 @@ class _ServiceAccountCredentials(AssertionCredentials):
             'private_key_id': self._private_key_id,
             'private_key': self._private_key_pkcs8_text
         }
+
+    @classmethod
+    def from_json(cls, s):
+        data = json.loads(_from_bytes(s))
+
+        credentials = cls(
+            service_account_id=data['_service_account_id'],
+            service_account_email=data['_service_account_email'],
+            private_key_id=data['_private_key_id'],
+            private_key_pkcs8_text=data['_private_key_pkcs8_text'],
+            scopes=[],
+            user_agent=data['_user_agent'])
+        credentials.invalid = data['invalid']
+        credentials.access_token = data['access_token']
+        token_expiry = data.get('token_expiry', None)
+        if token_expiry is not None:
+            credentials.token_expiry = datetime.datetime.strptime(
+                token_expiry, EXPIRY_FORMAT)
+        return credentials
 
     def create_scoped_required(self):
         return not self._scopes
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -618,6 +618,35 @@ class GoogleCredentialsTests(unittest2.TestCase):
             self.get_a_google_credentials_object().from_stream,
             credentials_file)
 
+    def test_to_from_json_authorized_user(self):
+        credentials_file = datafile(
+            os.path.join('gcloud', 'application_default_credentials_authorized_user.json'))
+        creds = GoogleCredentials.from_stream(credentials_file)
+        json = creds.to_json()
+        creds2 = GoogleCredentials.from_json(json)
+
+        self.assertEqual(creds.__dict__, creds2.__dict__)
+
+    def test_to_from_json_service_account(self):
+        self.maxDiff=None
+        credentials_file = datafile(
+            os.path.join('gcloud', _WELL_KNOWN_CREDENTIALS_FILE))
+        creds = GoogleCredentials.from_stream(credentials_file)
+
+        json = creds.to_json()
+        creds2 = GoogleCredentials.from_json(json)
+
+        self.assertEqual(creds.__dict__, creds2.__dict__)
+
+    def test_parse_expiry(self):
+        dt = datetime.datetime(2016, 1, 1)
+        parsed_expiry = client._parse_expiry(dt)
+        self.assertEqual('2016-01-01T00:00:00Z', parsed_expiry)
+
+    def test_bad_expiry(self):
+        dt = object()
+        parsed_expiry = client._parse_expiry(dt)
+        self.assertEqual(None, parsed_expiry)
 
 class DummyDeleteStorage(Storage):
     delete_called = False
@@ -771,6 +800,12 @@ class BasicCredentialsTests(unittest2.TestCase):
     def test_from_json_token_expiry(self):
         data = json.loads(self.credentials.to_json())
         data['token_expiry'] = None
+        instance = OAuth2Credentials.from_json(json.dumps(data))
+        self.assertTrue(isinstance(instance, OAuth2Credentials))
+
+    def test_from_json_bad_token_expiry(self):
+        data = json.loads(self.credentials.to_json())
+        data['token_expiry'] = 'foobar'
         instance = OAuth2Credentials.from_json(json.dumps(data))
         self.assertTrue(isinstance(instance, OAuth2Credentials))
 


oauth2client/client.py
```diff
@@ -1222,6 +1222,43 @@ def create_scoped(self, scopes):
...
+            data['token_expiry'],
+            data['token_uri'],
+            data['user_agent'],
+            revoke_uri=data.get('revoke_uri', None))
+        retval.invalid = data['invalid']
+        return retval
+
+    def to_json(self):
+        strip = (['_private_key'] +
+             GoogleCredentials.NON_SERIALIZED_MEMBERS)
```
dhermes: Indent should line up with `(`

oauth2client/client.py
```diff
@@ -1222,6 +1222,43 @@ def create_scoped(self, scopes):
...
+            data['_class'] == '_ServiceAccountCredentials'):
+            return _ServiceAccountCredentials.from_json(s)
+
+        if (data.get('token_expiry') and
+                not isinstance(data['token_expiry'], datetime.datetime)):
+            try:
+                data['token_expiry'] = datetime.datetime.strptime(
+                    data['token_expiry'], EXPIRY_FORMAT)
+            except ValueError:
+                data['token_expiry'] = None
```
dhermes: Rather than pushing through, shouldn't we raise this error the user as a malformed stored credential?

oauth2client/client.py
```diff
@@ -1222,6 +1222,43 @@ def create_scoped(self, scopes):
...
+        s = _from_bytes(s)
+        data = json.loads(s)
+
+        # A GoogleCredentials.from_stream return value might be a
+        # _ServiceAccountCredentials instance, so we should handle it in from_json.
+        if (data['_module'] == 'oauth2client.service_account' and
+            data['_class'] == '_ServiceAccountCredentials'):
+            return _ServiceAccountCredentials.from_json(s)
+
+        if (data.get('token_expiry') and
```
dhermes: Would probably be better to just use

``` python
expiry = data.get('token_expiry')
```

and then refer to `expiry` instead of `data['token_expiry']` after (twice).

oauth2client/client.py
```diff
@@ -1222,6 +1222,43 @@ def create_scoped(self, scopes):
...
+                not isinstance(data['token_expiry'], datetime.datetime)):
+            try:
+                data['token_expiry'] = datetime.datetime.strptime(
+                    data['token_expiry'], EXPIRY_FORMAT)
+            except ValueError:
+                data['token_expiry'] = None
+        retval = cls(
```
dhermes: ~~Isn't this already done on [`Credentials._to_json`](https://github.com/google/oauth2client/blob/8f8df1f5977ec76d50cb3dca9a31044cfd1fb9b4/oauth2client/client.py#L252)? If not, could it be re-purposed / upgraded to support this?~~

DERP, I just realized this is `from_json`

oauth2client/service_account.py
```diff
@@ -108,6 +112,29 @@ def serialization_data(self):
             'private_key': self._private_key_pkcs8_text
         }
 
+    def to_json(self):
+        strip = (['_private_key'] + self.NON_SERIALIZED_MEMBERS)
```
dhermes: Instead, add a class attribute to `_ServiceAccountCredentials`

``` python
NON_SERIALIZED_MEMBERS =  frozenset(
    ['_private_key'] +
    AssertionCredentials.NON_SERIALIZED_MEMBERS)
```

then you can just update `Credential.to_json` to use

```
self._to_json(self.NON_SERIALIZED_MEMBERS)
```

and just delete `OAuth2Credentials.to_json` (it is identical to `Credentials.to_json` anyhow).

We should also update `Credentials.NON_SERIALIZED_MEMBERS` to be a `frozenset`

dhermes: PS I realize this is a bit of a re-design but fits in the "leave it better than we found it" mindset

dlorenc: Done.

oauth2client/client.py
dlorenc: Refactored it a bit anyway.

oauth2client/client.py
dlorenc: Done.

oauth2client/client.py
dlorenc: Done.

dlorenc:
Thanks for the review, addressed most of the comments and pushed a new commit.

oauth2client/service_account.py
```diff
@@ -108,6 +117,29 @@ def serialization_data(self):
             'private_key': self._private_key_pkcs8_text
         }
 
+    # def to_json(self):
+    #     strip = (['_private_key'] + self.NON_SERIALIZED_MEMBERS)
+    #     return super(_ServiceAccountCredentials, self)._to_json(strip)
```
dhermes: Did you mean to leave in commented out code?

oauth2client/client.py
```diff
@@ -700,7 +704,7 @@ def retrieve_scopes(self, http):
         return self.scopes
 
     def to_json(self):
-        return self._to_json(Credentials.NON_SERIALIZED_MEMBERS)
+        return self._to_json(self.NON_SERIALIZED_MEMBERS)
```
dhermes: This method can be deleted, it is already implemented on the base class.

dhermes:

oauth2client/service_account.py
dlorenc: Deleted.

oauth2client/client.py
dlorenc: Deleted.

oauth2client/client.py
```diff
@@ -469,6 +467,12 @@ def clean_headers(headers):
     return clean
 
 
+def _parse_expiry(expiry):
+    if expiry and isinstance(expiry, datetime.datetime):
+        return expiry.strftime(EXPIRY_FORMAT)
+    return expiry
```
dhermes: I just realized this is bad if `expiry` is False-y but not `None`. Can you move the final `return expiry` since we really just want `None` to be returned (which is done by default)?

dlorenc: Done.

dhermes:
This LGTM but would like to give @nathanielmanistaatgoogle a chance to make a pass.

nathanielmanistaatgoogle:
"Resovles" in your commit message is a typo and should be "Resolves", right?

dhermes:
Yeah `resolves` is in https://help.github.com/articles/closing-issues-via-commit-messages/ but `Resovles` is not :grinning:

dlorenc:
Thanks for catching the typo. Fixed.

nathanielmanistaatgoogle:
(Will have other comments later; still working through the content of this change.)

oauth2client/client.py
```diff
@@ -469,6 +467,11 @@ def clean_headers(headers):
     return clean
 
 
+def _parse_expiry(expiry):
```
nathanielmanistaatgoogle: Please place this function earlier in the file (above) all of its use-sites.

oauth2client/client.py
```diff
@@ -469,6 +467,11 @@ def clean_headers(headers):
...
+    if expiry and isinstance(expiry, datetime.datetime):
+        return expiry.strftime(EXPIRY_FORMAT)
```
nathanielmanistaatgoogle: Never mix returning-an-expression and implicitly-return-None-by-naked-return-statement-or-falling-off-the-bottom-of-the-body in the same behavior. If you have any return statement of the form `return <expression>` anywhere in a behavior, then all of your return statements should be of the form `return <expression>` and execution should not be able to reach the end of the function without hitting a return statement.

In this case, place your `return None` inside an `else` rather than putting it at behavior top-level.

dhermes: @nathanielmanistaatgoogle This seems to be an arbitrary recommendation (that also happens to end up with more code)

oauth2client/client.py
```diff
@@ -1222,6 +1227,31 @@ def create_scoped(self, scopes):
         """
         return self
 
+    @classmethod
+    def from_json(cls, s):
+        from oauth2client.service_account import _ServiceAccountCredentials
```
nathanielmanistaatgoogle, line 1236: Import statements anywhere other than the top of a file need a very good reason to exist. What's the very good reason here?

dhermes, line 1236: Circular dependencies

oauth2client/client.py
nathanielmanistaatgoogle: It's not arbitrary, and yes, it lengthens the code.

oauth2client/client.py
nathanielmanistaatgoogle, line 1236: So there should be a comment here referencing the issue filed to resolve the circularity, at least?

oauth2client/client.py
dhermes: It's arbitrary in that I've never heard it recommended before. Maybe I meant subjective?

It seems to only protect Python beginners from misunderstanding, since `None` is the known return value.

oauth2client/client.py
```diff
@@ -1222,6 +1227,31 @@ def create_scoped(self, scopes):
...
+        data = json.loads(s)
+
+        # A GoogleCredentials.from_stream return value might be a
+        # _ServiceAccountCredentials instance, so we should handle it in from_json.
+        if (data['_module'] == 'oauth2client.service_account' and
+            data['_class'] == '_ServiceAccountCredentials'):
+            return _ServiceAccountCredentials.from_json(s)
+
+        token_expiry = _parse_expiry(data.get('token_expiry'))
+        retval = cls(
```
nathanielmanistaatgoogle: "retval" is always a bad name. Just call this "google_credentials"?

oauth2client/client.py
```diff
@@ -1222,6 +1227,31 @@ def create_scoped(self, scopes):
         """
         return self
 
+    @classmethod
+    def from_json(cls, s):
+        from oauth2client.service_account import _ServiceAccountCredentials
+        s = _from_bytes(s)
```
nathanielmanistaatgoogle: This one line isn't worth reassigning one of your parameters. How about just `data = json.loads(_from_bytes(s))` for the one line here?

oauth2client/client.py
```diff
@@ -1222,6 +1227,31 @@ def create_scoped(self, scopes):
...
+        data = json.loads(s)
+
+        # A GoogleCredentials.from_stream return value might be a
```
nathanielmanistaatgoogle: I don't understand this comment at all - we're implementing `from_json`, not `from_stream`. What does `from_stream` have to do with anything here?

oauth2client/service_account.py
```diff
@@ -108,6 +117,25 @@ def serialization_data(self):
             'private_key': self._private_key_pkcs8_text
         }
 
+    @classmethod
+    def from_json(cls, s):
+        s = _from_bytes(s)
+        data = json.loads(s)
+
+        retval = cls(
```
nathanielmanistaatgoogle: How about "service_account_credentials" instead of "retval"?

oauth2client/service_account.py
```diff
@@ -108,6 +117,25 @@ def serialization_data(self):
             'private_key': self._private_key_pkcs8_text
         }
 
+    @classmethod
+    def from_json(cls, s):
+        s = _from_bytes(s)
+        data = json.loads(s)
```
nathanielmanistaatgoogle: Same comment here about just doing this in one line instead of reassigning your parameter.

oauth2client/service_account.py
```diff
@@ -108,6 +117,25 @@ def serialization_data(self):
...
+        retval = cls(
+            service_account_id=data['_service_account_id'],
+            service_account_email=data['_service_account_email'],
+            private_key_id=data['_private_key_id'],
+            private_key_pkcs8_text=data['_private_key_pkcs8_text'],
+            scopes=[],
+            user_agent=data['_user_agent'])
+        retval.invalid = data['invalid']
+        retval.access_token = data['access_token']
+        if data.get('token_expiry', None):
```
nathanielmanistaatgoogle:
```
token_expiry = data.get('token_expiry', None)
if token_expiry is not None:
```

oauth2client/client.py
dlorenc: Done.

oauth2client/client.py
dlorenc: Done.

oauth2client/client.py
dlorenc: Done.

oauth2client/client.py
dlorenc: Done.

oauth2client/client.py
dlorenc: The goal of these methods is to make it so any object returned by "GoogleCredentials.from_stream" can be cached with from_json and to_json. Basically, this codepath should work:

creds = GoogleCredentials.from_stream(filename)
serialized_data = creds.to_json()
deserialized_object = GoogleCredentials.from_json(serialized_data)

since GoogleCredentials.from_stream returns _ServiceAccount instances, this method has to be able to deserialize _ServiceAccount instances.

Does that make sense?

oauth2client/service_account.py
dlorenc: Done.

oauth2client/service_account.py
dlorenc: Done.

oauth2client/service_account.py
dlorenc: Done.

oauth2client/client.py
dhermes: I think a more appropriate comment would be

``` python
# We handle service_account._ServiceAccountCredentials since it is a
# possible return type of GoogleCredentials.get_application_default()
```

dlorenc:
Addressed feedback. PTAL @nathanielmanistaatgoogle

oauth2client/client.py
nathanielmanistaatgoogle: I think this is the sticking point for me - I'm a strong adherent to the "a superclass' having any knowledge of any of its subclasses is an abstraction violation" design principle. Are you also an adherent? If so, do you agree that there's a problem here? If there's a problem here, are we making it much worse with this change or only a little worse? If there's a problem here, is there a fix we can manage now in this change or should we put it off until later?

oauth2client/client.py
```diff
@@ -1222,6 +1229,31 @@ def create_scoped(self, scopes):
...
+    def from_json(cls, s):
+        # This import needs to be here to avoid a circular dependency.
+        from oauth2client.service_account import _ServiceAccountCredentials
+        data = json.loads(_from_bytes(s))
+
+        # We handle service_account._ServiceAccountCredentials since it is a
+        # possible return type of GoogleCredentials.get_application_default()
+        if (data['_module'] == 'oauth2client.service_account' and
+            data['_class'] == '_ServiceAccountCredentials'):
+            return _ServiceAccountCredentials.from_json(s)
```
dhermes, line 1243: Re-surfacing the outdated comments here.

---

From me:

I think a more appropriate comment would be

```
# We handle service_account._ServiceAccountCredentials since it is a
# possible return type of GoogleCredentials.get_application_default()
```

---

From @nathanielmanistaatgoogle:

I think this is the sticking point for me - I'm a strong adherent to the "a superclass' having any knowledge of any of its subclasses is an abstraction violation" design principle. Are you also an adherent? If so, do you agree that there's a problem here? If there's a problem here, are we making it much worse with this change or only a little worse? If there's a problem here, is there a fix we can manage now in this change or should we put it off until later?

dhermes, line 1243: @nathanielmanistaatgoogle You're picking the wrong time to stick. The right time would've been when [`GoogleCredentials.get_application_default`](https://github.com/google/oauth2client/blob/8f8df1f5977ec76d50cb3dca9a31044cfd1fb9b4/oauth2client/client.py#L1340) was added.

It
- [can return](https://github.com/google/oauth2client/blob/8f8df1f5977ec76d50cb3dca9a31044cfd1fb9b4/oauth2client/client.py#L1253) an [`oauth2client.contrib.gce.AppAssertionCredentials`](https://github.com/google/oauth2client/blob/8f8df1f5977ec76d50cb3dca9a31044cfd1fb9b4/oauth2client/contrib/gce.py#L39)
- [can return](https://github.com/google/oauth2client/blob/8f8df1f5977ec76d50cb3dca9a31044cfd1fb9b4/oauth2client/client.py#L1236) an [`oauth2client.contrib.appengine.AppAssertionCredentials`](https://github.com/google/oauth2client/blob/8f8df1f5977ec76d50cb3dca9a31044cfd1fb9b4/oauth2client/contrib/appengine.py#L141)
- a [`GoogleCredentials`](https://github.com/google/oauth2client/blob/8f8df1f5977ec76d50cb3dca9a31044cfd1fb9b4/oauth2client/client.py#L1484)
- an [`oauth2client.service_account._ServiceAccountCredentials`](https://github.com/google/oauth2client/blob/8f8df1f5977ec76d50cb3dca9a31044cfd1fb9b4/oauth2client/client.py#L1493)

oauth2client/client.py
```diff
@@ -1222,6 +1229,31 @@ def create_scoped(self, scopes):
         """
         return self
 
+    @classmethod
+    def from_json(cls, s):
+        # This import needs to be here to avoid a circular dependency.
```
nathanielmanistaatgoogle: `TODO(issue 388): eliminate the circularity that is the reason for this non-top-level import.`

dhermes: -1 on TODO's in code. That is the value of having a world class issue tracking system. Up to you though.

oauth2client/client.py
nathanielmanistaatgoogle, line 1243: [:-(](http://sadtrombone.com/?autoplay=true)

oauth2client/client.py
nathanielmanistaatgoogle: I agree with you in the large because most TODOs don't refer the reader to a more authoritative place where the work item is tracked. The virtue of `TODO(<issue citation>):` is that it tells the reader of the code (who may not be familiar with every issue currently open in the issue tracker) that the problem they are seeing in the code is already considered a problem by the code's maintainers and tells them exactly where to find the specific issue covering the problem they're seeing.

@dlorenc: the problem with "This import needs to be here to avoid a circular dependency" is that it's tone-positive. It sounds like the speaker is saying "circular dependencies are usually wrong but here this is the right thing". A TODO here, in addition to the benefit laid out above, is that it's tone-negative - it tells the reader "we know this is a problem, just like circular dependencies are a problem in any code".

nathanielmanistaatgoogle:
Aside from what we're discussing about the TODO comment I'm happy with the content of this change. Please squash commits - a code review is just a code review; there's no need for this change to be more than one piece of project history (two after GitHub does its thing).

dlorenc:
Added the TODO and squashed. Thanks for the review!

nathanielmanistaatgoogle:
Looks good; only waiting for tests to repass.

PR #387, opened 2016-01-25
Look at the v1 endpoint for the GCE metadata server.

Galabar001:


oauth2client/contrib/gce.py
```diff
@@ -84,7 +84,8 @@ def _refresh(self, http_request):
         """
         query = '?scope=%s' % urllib.parse.quote(self.scope, '')
         uri = META.replace('{?scope}', query)
-        response, content = http_request(uri)
+        headers = { 'Metadata-Flavor' : 'Google' }
```
dhermes: Instead:

``` python
headers = {'Metadata-Flavor': 'Google'}
```

tests/contrib/test_gce.py
```diff
@@ -66,11 +66,13 @@ def _refresh_success_helper(self, bytes_response=False):
...
 
-        base_metadata_uri = ('http://metadata.google.internal/0.1/meta-data/'
-                             'service-accounts/default/acquire')
+        base_metadata_uri = (
+            'http://metadata.google.internal/computeMetadata/v1/instance/'
+            'service-accounts/default/acquire')
         escaped_scopes = urllib.parse.quote(' '.join(scopes), safe='')
         request_uri = base_metadata_uri + '?scope=' + escaped_scopes
-        http.request.assert_called_once_with(request_uri)
+        headers = { 'Metadata-Flavor' : 'Google' }
```
dhermes: Ditto here on the superfluous spaces

tests/contrib/test_gce.py
```diff
@@ -158,8 +160,9 @@ def test_get_access_token(self):
         self.assertEqual(None, token.expires_in)
 
         http.request.assert_called_once_with(
-            'http://metadata.google.internal/0.1/meta-data/service-accounts/'
-            'default/acquire?scope=dummy_scope')
+            'http://metadata.google.internal/computeMetadata/v1/instance/'
+            'service-accounts/default/acquire?scope=dummy_scope',
+            headers={'Metadata-Flavor' : 'Google'})
```
dhermes: Ditto on space before `:`

oauth2client/contrib/gce.py
dhermes: Did the header become required with `v1`? Is something broken?

Galabar001: The header is required for v1.

dhermes: I'm all about moving to the latest version, so this is more a curiosity: Was there some kind of announcement about deprecating the old URI or an encouragement for users to upgrade?

Are there implications for the [other place](https://github.com/google/oauth2client/blob/8f8df1f5977ec76d50cb3dca9a31044cfd1fb9b4/oauth2client/client.py#L116) this library talks to the metadata API?

oauth2client/contrib/gce.py
```diff
@@ -84,7 +84,8 @@ def _refresh(self, http_request):
         """
         query = '?scope=%s' % urllib.parse.quote(self.scope, '')
         uri = META.replace('{?scope}', query)
-        response, content = http_request(uri)
+        headers = { 'Metadata-Flavor': 'Google' }
```
nathanielmanistaatgoogle: Drop the space to the right of `{` and to the left of `}`.

oauth2client/contrib/gce.py
```diff
@@ -84,7 +84,8 @@ def _refresh(self, http_request):
...
+        response, content = http_request(uri, headers=headers)
```
nathanielmanistaatgoogle: Drop the introduction of the local field `headers` and in general be reticent about introducing a new name only to use it exactly once and on the immediately following line. Just use the dictionary literal expression in this call.

If `http_request(uri, headers={'Metadata-Flavor': 'Google'})` were a too-complex expression that spread across three lines, defining a named value for some intermediate part of it would be fine, but that's not the case here.

tests/contrib/test_gce.py
```diff
@@ -66,11 +66,13 @@ def _refresh_success_helper(self, bytes_response=False):
         credentials.refresh(http)
         self.assertEquals(access_token, credentials.access_token)
 
-        base_metadata_uri = ('http://metadata.google.internal/0.1/meta-data/'
-                             'service-accounts/default/acquire')
+        base_metadata_uri = (
```
nathanielmanistaatgoogle, line 69: Why not make use of `gce.META` here? I suppose it's a philosophical question: does the correctness of the `gce` module (which is the matter under test) depend upon this exact URL value, or does it depend upon the `gce` module using whatever value is stored in its `META` attribute?

Galabar001, line 69: My thought here is that it is usually better to repeat the URL here as that is the changed item (and if it changes in the future, we want to make sure the correct URL is being used).

tests/contrib/test_gce.py
```diff
@@ -66,11 +66,13 @@ def _refresh_success_helper(self, bytes_response=False):
...
+            'http://metadata.google.internal/computeMetadata/v1/instance/'
+            'service-accounts/default/acquire')
         escaped_scopes = urllib.parse.quote(' '.join(scopes), safe='')
         request_uri = base_metadata_uri + '?scope=' + escaped_scopes
-        http.request.assert_called_once_with(request_uri)
+        headers = { 'Metadata-Flavor': 'Google' }
```
nathanielmanistaatgoogle: As was done in the non-test file: drop the space to the right of `{` and to the left of `}` and elide the local "headers" field.

nathanielmanistaatgoogle:
Code change looks more or less good to me but please squash commits; I don't see any reason for this to enter project source control history as more than just one change.

@dhermes any further concerns from you?

Galabar001:
When you say "squash commits", how do I do that?

nathanielmanistaatgoogle:
However you want to produce one commit that contains all the change you want to make. [There are lots of different recipes out there](https://www.google.com/#q=git+squash); [this one](http://stackoverflow.com/questions/6934752/combining-multiple-commits-before-pushing-in-git/6934882#6934882) looks slightly more user-friendly than the rest.

Galabar001:
Does the following look correct:

https://github.com/google/oauth2client/commit/998637ab764f9d530d1730ed8e6a7f3f4993ba9b

nathanielmanistaatgoogle:
Looks correct code-wise. Please update the commit log message to conform to [these guidelines](http://chris.beams.io/posts/git-commit#seven-rules).

Galabar001:
How does this look:

https://github.com/google/oauth2client/commit/c9809c87f1352f951cc526b83e4b66ced8e8c3ae

nathanielmanistaatgoogle:
Looks good! Anything else before I merge it?

Galabar001:
I think we're good to go.

dhermes:
@Galabar001 this doesn't actually work. Are these endpoints live yet? Did you test on your own GCE instance?

@nathanielmanistaatgoogle I am working on a fix now. It might be worth have a system test that runs on GCE. Though that'd be a lot of work.

jonparrott:
I can help with system testing on GCE, if needed.

dhermes:
I am working on it. Don't need multiple cooks. Thanks though.

jonparrott:
@dhermes of course, just ping me if you need anything.

Galabar001:
I relied on unit tests -- should have tested it on a GCE instance.

Galabar001:
Let me know when this is ready to go.  We have many internal references (gce.py) in google3/ and I'd like to start updating thos.

Thanks!

dhermes:
I cc'ed you on #411, check it out.

jgeewax:
OK -- from now on, can we make sure there are tests on this one? @nathanielmanistaatgoogle 

I have no idea how long this has been broken, but it really bums me out that I found it out from reading a blog post of someone complaining about it :(

dhermes:
@jgeewax It's been broken for less than 24 hours (2.0.0 went out mid-day yesterday).

The amount of code required will be large, but I think worth it. From Travis we need code to:

- Spin up a GCE instance (maybe we should just have everything pre-installed on an image)
- Remotely execute code on that instance
- Get the result back

Definitely should only run that on merges to `master` and maybe only on tagged release commits?

@jonparrott WDYT?

jgeewax:
24 hours is awesome. The bigger fix then is the nonsense about "access to all Google Cloud APIs" to actually mean what it says (and include the `userinfo.email` scope).

Galabar001:
Sorry for this bug -- I've been making literally 100s of changes to clean
up this issue.  It looks like I added the wrong URL to the unit test, which
matched the wrong URL in the code.

First quick fix should be to update the unit test with the appropriate URL.

jonparrott:
@dhermes we have a jenkins instance running in GCE that we use for system testing samples. We can just add this repo. You can use `nose.plugins.attr` to mark it and add another tox environment for jenkins, or just have the test check if the metadata server is available and raise `skiptest` if it isn't.

Alternatively, I agree we can write a system test that creates GCE instance but it is a lot to maintain. There's a few tools/tricks we can use to make it easier.

dhermes:
I prefer the one that takes the least effort and least code to implement :grinning:

dhermes:
@Galabar001 The fix is in #411.

Galabar001:
I was surprised by the whole scope thing when I started looking into this.

jonparrott:
@dhermes Cool, so let's go with option one. You get the test there, I'll get it running in Jenkins. :+1:

dhermes:
Boom! I like the way you think. First things first is make `GAE_PYTHONPATH` optional in `tox`. (Can't remember who made it required... :grinning: )

jonparrott:
> Can't remember who made it required...

Probably my fault, but it shouldn't be required?

dhermes:
Not globally, I'll send a PR.

PR #389, opened 2016-01-28
Fall back to read-only credentials on EACCES errors

### LINKED ISSUE #386: errno.EACCES can propagate fatally outside of multistore_file
Report from a Windows gsutil user:

```
Encountered a problem: [Errno 13] Permission denied: 'C:\\Users\\Administrator\\
.gsutil\\credstore.lock'
```

My best guess is that it is occurring `locked_file.py` [here](https://github.com/google/oauth2client/blob/41eb75578f81876a53f1707cf1daf1eb89ddfc0b/oauth2client/contrib/locked_file.py#L291).

I suspect we need to add `errno.EACCES` to the list of exceptions handled in `multistore_file.py` [here](https://github.com/google/oauth2client/blob/41eb75578f81876a53f1707cf1daf1eb89ddfc0b/oauth2client/contrib/multistore_file.py#L286), since in theory we should be able to proceed by acquiring new credentials.

thobrla:
This change fixes a bug where multiple threads and/or processes
using multistore_file on Windows to access the same backing
store could raise IOError errno.EACCES to the calling application.
This change causes a fallback to read only mode and refreshes
credentials if necessary.

Fixes #386

### DIFF
--- a/oauth2client/contrib/multistore_file.py
+++ b/oauth2client/contrib/multistore_file.py
@@ -293,6 +293,8 @@ class _MultiStore(object):
             elif e.errno == errno.EDEADLK:
                 logger.warn('Lock contention on multistore file, opening '
                             'in read-only mode.')
+            elif e.errno == errno.EACCES:
+                logger.warn('Cannot access credentials file.')
             else:
                 raise
         if not self._file.is_locked():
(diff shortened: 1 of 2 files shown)


jonparrott:
Can you add a test case for this?

thobrla:
I did in test_multistore_file?

jonparrott:
Oh, of course. I get an 'F' for reading comprehension.

@dhermes @nathanielmanistaatgoogle this change looks good to me, please take a look.

PR #394, opened 2016-02-04
Use symbolic constants for HTTP codes

nathanielmanistaatgoogle:
I figured why not do something really easy for my first substantive change to this codebase. :-P

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -30,6 +30,7 @@ import tempfile
 import time
 import shutil
 import six
+from six.moves import http_client
 from six.moves import urllib
 
 import httplib2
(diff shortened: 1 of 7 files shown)


nathanielmanistaatgoogle:
(Oh, and _of course_ this may wait until after the more important work we're doing.)

oauth2client/client.py
```diff
@@ -73,7 +74,7 @@
 OOB_CALLBACK_URN = 'urn:ietf:wg:oauth:2.0:oob'
 
 # Google Data client libraries may need to set this to [401, 403].
-REFRESH_STATUS_CODES = [401]
+REFRESH_STATUS_CODES = [http_client.UNAUTHORIZED]
```
dhermes: Use a tuple instead of a list. (You break it you buy it)

dhermes:
LGTM pending Travis / small change from my comment

oauth2client/client.py
nathanielmanistaatgoogle: Done and repushed.

PR #395, opened 2016-02-04
Adding a pure Python crypt sub-module.

dhermes:
Resurrects #1.


nathanielmanistaatgoogle:
Change content looks good; waiting on test results.

PR #399, opened 2016-02-05
Removing SignedJwtAssertionCredentials.

### LINKED ISSUE #211: Clean up Service Account support

dhermes:
This completes the consolidation of the two service account credentials implementations.

In the process, also adding test coverage for some untested code paths within the crypto helpers.

Fixes #211.

**NOTE**: Has #398 as diffbase.

### DIFF
--- a/oauth2client/_openssl_crypt.py
+++ b/oauth2client/_openssl_crypt.py
@@ -68,6 +68,7 @@ class OpenSSLVerifier(object):
         Raises:
             OpenSSL.crypto.Error: if the key_pem can't be parsed.
         """
+        key_pem = _to_bytes(key_pem)
         if is_x509_cert:
             pubkey = crypto.load_certificate(crypto.FILETYPE_PEM, key_pem)
         else:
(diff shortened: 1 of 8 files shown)


nathanielmanistaatgoogle:
Tell me about the Python2.6 failures in this?

dhermes:
Ah phooey! Needed to be `unittest2` to use a context manager

nathanielmanistaatgoogle:
[I apologize that that's even an obstacle in our path over which to stumble.](https://github.com/google/oauth2client/issues/298)

dhermes:
Can't really change it. You could merge as-is and send a fixup right after or you could fork this branch

nathanielmanistaatgoogle:
Superseded by #400.

PR #400, opened 2016-02-05
Removing SignedJwtAssertionCredentials.

nathanielmanistaatgoogle:
This is #399 with `tests/test__pycrypto_crypt.py` changed to use `unittest2`.

### DIFF
--- a/oauth2client/_openssl_crypt.py
+++ b/oauth2client/_openssl_crypt.py
@@ -123,19 +123,17 @@ class OpenSSLSigner(object):
         return OpenSSLSigner(pkey)
 
 
-def pkcs12_key_as_pem(private_key_text, private_key_password):
-    """Convert the contents of a PKCS12 key to PEM using OpenSSL.
+def pkcs12_key_as_pem(private_key_bytes, private_key_password):
+    """Convert the contents of a PKCS#12 key to PEM using pyOpenSSL.
 
     Args:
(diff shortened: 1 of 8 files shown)


dhermes:
LGTM, though I meant to update the `s` to `json_str` change actually be `json_data` (based on #396 or #397)

nathanielmanistaatgoogle:
I can change `json_str` to `json_data`.

Any idea why the tests haven't yet kicked off for this?

nathanielmanistaatgoogle:
@googlebot you're being silly and failing to recognize an authored-by-@dhermes-committed-by-@nathanielmanistaatgoogle commit as legitimate to merge.

dhermes:
1. ISTM the `google` org gets rate limited (also having 8 Travis envs probably doesn't help)
1. I've had the CLA check fail like this before. I think it's a corner case the server isn't implemented to handle?

nathanielmanistaatgoogle:
Changed `json_str` to `json_data` in `oauth2client/client.py`; please take another quick look.

Any way to poke Travis to make it notice that a different commit has been pushed? I cancelled the test run of 81da3192423237b6.

dhermes:
You've gotta figure out what the rate limit is. Need to check with whoever owns the org

dhermes:
Still LGTM

PR #403, opened 2016-02-10
Use the latest version of the GCE metadata server endpoint

Galabar001:
We are updating all legacy uses (0.1 and v1beta1) of the GCE
server to use the latest (v1) endpoint.


nathanielmanistaatgoogle:
It's odd to see a source change without a corresponding change to test code - what's up with that?

Galabar001:
Can you point me to the tests for this particular file?

Galabar001:
Let me know what you think of the new tests.

nathanielmanistaatgoogle:
I think the new test methods look alright but I'd like @dhermes to also review.

Please squash commits and please amend your commit log message to conform to [the guidelines](chris.beams.io/posts/git-commit/#seven-rules).

Galabar001:
Let me know how this new commit looks.

oauth2client/client.py
```diff
@@ -837,6 +838,9 @@ def _generate_refresh_request_headers(self):
         if self.user_agent is not None:
             headers['user-agent'] = self.user_agent
 
+        if _METADATA_V1_URL_PREFIX in self.token_uri:
+            headers[_METADATA_FLAVOR_HEADER] = _DESIRED_METADATA_FLAVOR
```
dhermes, line 842: This seems a bit aggressive in that it may capture URIs not indended. Why not just ensure that all the callers send the header?

Galabar001, line 842: It doesn't look like the OAuth2Credentials class takes a headers argument.  Are you suggesting adding one (seems reasonable, but I want to be sure before I dive into it)?

Galabar001, line 842: Or is there some other way for the user to specify a header to be sent with the token URI?

dhermes, line 842: There is no other way on the `oauth2client.client.OAuth2Credentials` class, but this header should only be relevant to the `oauth2client.client.contrib.gce.AppAssertionCredentials` class. You should just over-ride `_generate_refresh_request_headers` there.

Galabar001, line 842: Looking at some of our code, it seems that we are directly instantiating an `oauth2client.client.OAuth2Credentials` object with a metadata URL.  How does the `AppAssertionCredentials` come into play at that point?

dhermes, line 842: Care to point to which code you are referring to?

dhermes:
@Galabar001 Can we keep moving forward on this?

Galabar001:
LGTM.

Galabar001:
LGTM meant for other change (test_gce.py).

Galabar001:
I'm still trying to understand where this is used by internal clients.

Galabar001:
We've got a client that does something like:

from oauth2client.client import OAuth2Credentials

METADATA_URL = 'http://metadata/0.1/meta-data/service-accounts'

class GoogleComputeCredentials(OAuth2Credentials):
   ...
   def __init__(...):
   ...
   query = urllib.urlencode({'scopes': scopes})
    token_uri = '%s/%s/acquire?%s' % (METADATA_URL, service_account, query)
    super(GoogleComputeCredentials, self).__init__(
        '',
        None,
        None,
        None,
        None,
        token_uri,
        user_agent)

Galabar001:
Were can we add the needed headers in this case?

Galabar001:
/Were/Where/

dhermes:
@Galabar001 why would `GoogleComputeCredentials` exist when `oauth2client.contrib.gce.AppAssertionCredentials` already does?

Galabar001:
Good question.  This particular code is overriding _do_refresh_request (take a look at compute_client.py internally).

So, it may be the case that anyone using the OAuth2Credentials directly is either doing their own thing with _do_refresh_request or they are using AppAssertionCredentials (although I can't be sure about this).

dhermes:
Just chatted with @Galabar001 and figured it the impetus behind this. Issue resolved, no change needed here.

PR #404, opened 2016-02-11
Release 2.0.0

nathanielmanistaatgoogle:


nathanielmanistaatgoogle:
Per [your comment over in this issue](https://github.com/google/oauth2client/issues/401#issuecomment-181062999) do we want to provide migration instructions in `CHANGELOG.md` or elsewhere?

dhermes:
I think `CHANGELOG.md` is fine. Did you want to get the JWT scopeless creds in before a release?

nathanielmanistaatgoogle:
I wanted to but haven't - other things have just been piling on top of me for the last six days.

I've been wanting to get this library on a regular boring uneventful release schedule and if that's the case then whenever scopeless lands (next week!) it won't be any more than one week from release.

Would you like to write the migration instructions?

dhermes:
You can just add to the change log:

----

For JSON keys, you can create a credential via

```python
from oauth2client.service_account import ServiceAccountCredentials
credentials = ServiceAccountCredentials.from_json_keyfile_name(
    key_filename, scopes=[...])
```

You can still rely on

```python
from oauth2client.client import GoogleCredentials
credentials = GoogleCredentials.get_application_default()
```

returning these credentials when you set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable.

For `.p12` keys, construct via

```python
credentials = ServiceAccountCredentials.from_p12_keyfile(
    svc_acct_email, key_filename, scopes=[...])
```

though we urge you to use JSON keys (rather than `.p12` keys) if you can.

This is equivalent to the previous method

```python
from oauth2client.client import SignedJwtAssertionCredentials

with open(key_filename, 'rb') as file_obj:
    private_key = file_obj.read()

credentials = SignedJwtAssertionCredentials(
    svc_acct_email, private_key, scope=[...])
```

nathanielmanistaatgoogle:
@dhermes: I'm tempted to hold off on this until the integration of #406; do you agree?

dhermes:
Yes I agree.

nathanielmanistaatgoogle:
Updated the changelog; see it rendered at https://github.com/nathanielmanistaatgoogle/oauth2client/blob/release-2.0.0/CHANGELOG.md.

nathanielmanistaatgoogle:
I believe this is now ready to be merged?

dhermes:
Yeah it looks fine.

PR #410, opened 2016-02-16
Use symbolic constant rather than literal value

nathanielmanistaatgoogle:
This cost me a few minutes of debugging because I was looking for use of the symbolic constant and it never occurred to me to look for use of the literal string value.

Another reason to get excited about using [`enum`](https://pypi.python.org/pypi/enum34) for enumerated types!


dhermes:
LGTM

PR #411, opened 2016-02-17
Fixing broken token acquire endpoint on GCE.

dhermes:
I wanted to get this out there before writing / running tests, so there may be a follow on commit.

See #387 and https://github.com/GoogleCloudPlatform/gcloud-python/issues/1484 for context.

/cc @nathanielmanistaatgoogle @jonparrott @jgeewax @tseaver @Galabar001

### DIFF
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -34,7 +34,7 @@ logger = logging.getLogger(__name__)
 
 # URI Template for the endpoint that returns access_tokens.
 META = ('http://metadata.google.internal/computeMetadata/v1/instance/'
-        'service-accounts/default/acquire{?scope}')
+        'service-accounts/default/token{?scope}')
 
 
 class AppAssertionCredentials(AssertionCredentials):
@@ -90,11 +90,11 @@ class AppAssertionCredentials(AssertionCredentials):
         content = _from_bytes(content)
         if response.status == http_client.OK:
             try:
-                d = json.loads(content)
+                token_content = json.loads(content)
             except Exception as e:
                 raise HttpAccessTokenRefreshError(str(e),
                                                   status=response.status)
-            self.access_token = d['accessToken']
+            self.access_token = token_content['access_token']
         else:
             if response.status == http_client.NOT_FOUND:
                 content += (' This can occur if a VM was created'
--- a/tests/contrib/test_gce.py
+++ b/tests/contrib/test_gce.py
@@ -12,10 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Tests for oauth2client.contrib.gce.
-
-Unit tests for oauth2client.contrib.gce.
-"""
+"""Unit tests for oauth2client.contrib.gce."""
 
 import json
 from six.moves import http_client
@@ -54,7 +51,7 @@ class AppAssertionCredentialsTests(unittest.TestCase):
 
     def _refresh_success_helper(self, bytes_response=False):
         access_token = u'this-is-a-token'
-        return_val = json.dumps({u'accessToken': access_token})
+        return_val = json.dumps({u'access_token': access_token})
         if bytes_response:
             return_val = _to_bytes(return_val)
         http = mock.MagicMock()
@@ -69,7 +66,7 @@ class AppAssertionCredentialsTests(unittest.TestCase):
 
         base_metadata_uri = (
             'http://metadata.google.internal/computeMetadata/v1/instance/'
-            'service-accounts/default/acquire')
+            'service-accounts/default/token')
         escaped_scopes = urllib.parse.quote(' '.join(scopes), safe='')
         request_uri = base_metadata_uri + '?scope=' + escaped_scopes
         http.request.assert_called_once_with(
@@ -153,7 +150,7 @@ class AppAssertionCredentialsTests(unittest.TestCase):
         http = mock.MagicMock()
         http.request = mock.MagicMock(
             return_value=(mock.Mock(status=http_client.OK),
-                          '{"accessToken": "this-is-a-token"}'))
+                          '{"access_token": "this-is-a-token"}'))
 
         credentials = AppAssertionCredentials(['dummy_scope'])
         token = credentials.get_access_token(http=http)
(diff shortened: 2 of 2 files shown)


oauth2client/contrib/gce.py
```diff
@@ -58,6 +58,9 @@ def __init__(self, scope, **kwargs):
             scope: string or iterable of strings, scope(s) of the credentials
                    being requested.
         """
+        # if scope:
+        #     raise ValueError('GCE will not respect your scopes so what '
+        #                      'is the point?')
         self.scope = util.scopes_to_string(scope)
```
tseaver, line 61: Is this a stray commit?  If we truly cannot usefully pass in the `scope`, then why not remove it altogether?

dhermes, line 61: Not a stray commit. Just editorializing by me to elicit comments. (Commit message is `Pretend commit but really we should.`)

I'm just worried that if we remove `scope` from the constructor it will break lots of people.

dhermes:
@nathanielmanistaatgoogle PTAL. This is a serious regression and we should deal with it quickly.

oauth2client/contrib/gce.py
```diff
@@ -58,6 +58,9 @@ def __init__(self, scope, **kwargs):
             scope: string or iterable of strings, scope(s) of the credentials
                    being requested.
         """
+        # if scope:
```
nathanielmanistaatgoogle: No commented-out code. Should this be an issue?

nathanielmanistaatgoogle:
Thank you for emphasizing the importance of this; I'm happy to be around to respond to it for the next several hours.

nathanielmanistaatgoogle:
Please squash commits.

oauth2client/contrib/gce.py
dhermes: See the remarks one line below (Is this thing on?)

dhermes:
Will squash commits post review. Any issues other than the fake 4th commit?

oauth2client/contrib/gce.py
nathanielmanistaatgoogle: Let's do something real, though - how about logging a warning, or better yet making use of the [warnings module](https://docs.python.org/2/library/warnings.html)?

dhermes:

oauth2client/contrib/gce.py
```diff
@@ -51,13 +59,17 @@ class AppAssertionCredentials(AssertionCredentials):
     """
 
     @util.positional(2)
-    def __init__(self, scope, **kwargs):
+    def __init__(self, scope='', **kwargs):
```
nathanielmanistaatgoogle: Do you have to change the `@util.positional(2)` decoration on this method if you change one of its positional arguments to be optional? How safe a change is this to make? Can we not live with the parameter being positional until the next major version bump of the library?

oauth2client/contrib/gce.py
```diff
@@ -51,13 +59,17 @@ class AppAssertionCredentials(AssertionCredentials):
...
         """Constructor for AppAssertionCredentials
 
         Args:
             scope: string or iterable of strings, scope(s) of the credentials
```
nathanielmanistaatgoogle, line 58: How about just "any type, ignored" for a parameter description? Or "NoneType, a warning will be emitted for any non-None value but otherwise ignored"?

nathanielmanistaatgoogle:
The commit log message of 23b998ae7390f329 is weird for the way it contains a verb tense problem ("in the process also [...] warning a user") and the hash, author, and date of an earlier draft commit.

oauth2client/contrib/gce.py
dhermes: Good catch, I wasn't even paying attention to that. How do you feel about I just dropping `util.positional()`?

oauth2client/contrib/gce.py
dhermes, line 58: Any Truth-y value. Will update.

dhermes, line 58: Update: The type matters so long as we call

``` python
```

dhermes:
@nathanielmanistaatgoogle Fixed the commit message

oauth2client/contrib/gce.py
nathanielmanistaatgoogle, line 58: I just realized the scope (heh) of this change is larger than I had been thinking, because this change doesn't eliminate the assignment of the instance-scope `scope` field from the instance constructor.

This means we're dug into a hole of some sort, right? And of some depth?

How about an implementation comment explaining why it is that users must not pass a non-empty value of the `scope` parameter but we're forced to assign it to an instance-scope attribute?

Is it because the attribute is a public attribute?

Apologies for the confusion. Also I'll be away from my screen for the next few hours.

dhermes, line 58: We're not in a hole. GCE never honored the scopes and now we're just adding a warning for it (and also not sending it in the 

```
http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token
```

request. The only other change is to make it so that users don't **need** to pass scopes (since they don't do anything). But the rest of the behavior (i.e. for old code that does pass scopes) will work just fine.

dhermes:
@nathanielmanistaatgoogle I'm going to break this in two. Will ping you once updated.

dhermes:
@nathanielmanistaatgoogle PTAL. Sent out #419 for the contentious pieces.

nathanielmanistaatgoogle:
Looks good; merge when you like.

dhermes:
@nathanielmanistaatgoogle Do you want to cut a `2.0.0b` or `2.0.1` release with this fix for the regression? Should we wait until #419 to cut a release?

jonparrott:
Please cut a `2.0.0b` ASAP, as this is breaking a lot of stuff. @nathanielmanistaatgoogle we need to cut a new release of `google-api-python-client`, at least one to pin the oauth2client dependency because it's currently *very* broken. This is bad. :(

dhermes:
Sorry, just double checked [PEP 440][1] and I think the correct version would be `2.0.0-post1`. Can someone confirm? (Going to get working on the release PR right now.)

[1]: https://www.python.org/dev/peps/pep-0440/#post-release-separators

jonparrott:
Confirmed, it should be `-post1`.

PR #413, opened 2016-02-17
Implement ServiceAccountCredentials.from_p12_keyfile_buffer().

### LINKED ISSUE #412: ServiceAccountCredentials should support directly passing in key file contents
(breaking out my comment from #401 as separate feature request)

Previously, `SignedJwtAssertionCredentials` allowed you to pass in the key file contents directly instead of passing in a file path (as shown [here](https://developers.google.com/api-client-library/python/auth/service-accounts)) It would be great if `ServiceAccountCredentials` also supported this, for both JSON and p12 keys.

In the meantime, this has prevented us from upgrading to v2.0.0, as our keys are pulled in dynamically at runtime and it's much easier for us to directly pass in the contents instead of having to first write to disk, pass in the path, and then do all the necessary cleanup.

dhermes:
Fixes #412.

/cc @allenpc

### DIFF
--- a/oauth2client/service_account.py
+++ b/oauth2client/service_account.py
@@ -216,6 +216,38 @@ class ServiceAccountCredentials(AssertionCredentials):
         return cls._from_parsed_json_keyfile(keyfile_dict, scopes)
 
     @classmethod
+    def _from_p12_keyfile_contents(cls, service_account_email,
+                                   private_key_pkcs12,
+                                   private_key_password=None, scopes=''):
+        """Factory constructor from JSON keyfile.
+
+        Args:
+            service_account_email: string, The email associated with the
+                                   service account.
+            private_key_pkcs12: string, The contents of a PKCS#12 keyfile.
+            private_key_password: string, (Optional) Password for PKCS#12
+                                  private key. Defaults to ``notasecret``.
+            scopes: List or string, (Optional) Scopes to use when acquiring an
+                    access token.
+
+        Returns:
+            ServiceAccountCredentials, a credentials object created from
+            the keyfile.
+
+        Raises:
+            NotImplementedError if pyOpenSSL is not installed / not the
+            active crypto library.
+        """
+        if private_key_password is None:
+            private_key_password = _PASSWORD_DEFAULT
+        signer = crypt.Signer.from_string(private_key_pkcs12,
+                                          private_key_password)
+        credentials = cls(service_account_email, signer, scopes=scopes)
+        credentials._private_key_pkcs12 = private_key_pkcs12
+        credentials._private_key_password = private_key_password
+        return credentials
+
+    @classmethod
     def from_p12_keyfile(cls, service_account_email, filename,
                          private_key_password=None, scopes=''):
         """Factory constructor from JSON keyfile.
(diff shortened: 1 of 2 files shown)


tests/test_service_account.py
```diff
@@ -173,6 +173,30 @@ def test_from_p12_keyfile_explicit(self):
...
+            # Seek back to the beginning so the buffer can be
+            # passed to the constructor.
+            file_obj.seek(0)
+            creds = ServiceAccountCredentials.from_p12_keyfile_buffer(
+                service_account_email, file_obj,
+                private_key_password=private_key_password,
+                scopes=scopes)
+        # Check the created object.
+        self.assertIsInstance(creds, ServiceAccountCredentials)
+        self.assertEqual(creds.client_id, None)
```
nathanielmanistaatgoogle: Since this is already a `unittest2`-using file, `assertIsNone`, please (here and below).

dhermes: @nathanielmanistaatgoogle PTAL

allenpc:
Looking good from my side :+1:

dhermes:
Couldn't decided between `from_p12_keyfile_buffer`, `from_p12_keyfile_stream` or something else. Any preference?

nathanielmanistaatgoogle:
`from_p12_key_text`?

dhermes:
It isn't text, it's a `read()`-able object

oauth2client/service_account.py
```diff
@@ -248,6 +247,62 @@ def from_p12_keyfile(cls, service_account_email, filename,
...
+
+    @classmethod
+    def from_p12_keyfile_buffer(cls, service_account_email, file_buffer,
+                                private_key_password=None, scopes=''):
+        """Factory constructor from JSON keyfile.
+
+        Args:
+            service_account_email: string, The email associated with the
+                                   service account.
+            file_buffer: file-like object, Buffer that implements ``read()``
```
nathanielmanistaatgoogle: `file-like object that implements`.

dhermes: The style is `{type}, Sentence here that is a description`. You're saying `{type}` should be really long?

nathanielmanistaatgoogle:
Drop `file` (without "-like"); the argument may have come from the filesystem or it may not have. `stream` might be as good as anything else; Python doesn't seem to have a consistent single term for what we're talking about although "file-like" is the most common that I've seen. `from_p12_key_stream`?

oauth2client/service_account.py
nathanielmanistaatgoogle: Unfortunately, yes, I think so?

dhermes:
Are you saying we should rename `from_p12_key_buffer` --> `from_p12_key_stream`?

nathanielmanistaatgoogle:
I'm getting a little lost; let me back up a bit: for the methods that don't themselves read from the file system, why is it important that their users have the idea of a file called to mind when contemplating whether or not to use them?

dhermes:
It's not important.

nathanielmanistaatgoogle:
(I'm content to drop the discussion if I'm annoying you, by the way.)

I think `from_p12_key_buffer` and `from_p12_key_stream` are both fine as names?

oauth2client/service_account.py
```diff
@@ -248,6 +247,62 @@ def from_p12_keyfile(cls, service_account_email, filename,
...
+    @classmethod
+    def from_p12_keyfile_buffer(cls, service_account_email, file_buffer,
+                                private_key_password=None, scopes=''):
+        """Factory constructor from JSON keyfile.
+
+        Args:
+            service_account_email: string, The email associated with the
+                                   service account.
+            file_buffer: stream, A buffer that implements ``read()``
+                         and contains the PKCS#12 key contents.
```
dhermes, line 287: @nathanielmanistaatgoogle I updated this. PTAL

nathanielmanistaatgoogle:
Looks good; squash commits and merge at your leisure. :+1:

PR #415, opened 2016-02-17
Making GAE_PYTHON optional in tox config.

dhermes:
Fallback is `${GIT_ROOT}/google_appengine`.

~~The goal is to also make these pull the latest version of the SDK (right now it is downloaded once and frozen in time).~~


jonparrott:
What was the issue with these commands being in `tox`, why did we have to move them to their own scripts?

dhermes:
The issue is that `GAE_PYTHONPATH` is a **requirement**, rather than an optional flag that can be used to over-ride. Just running `tox` is not possible in a "clean" environment.

jonparrott:
(as an aside, I love tox but hate using an ini file for configuration)

dhermes:
@jonparrott I am aware of the default via `{env:VARNAME:DEFAULT_VAL}` but

- was unclear on how to make this path relative to the root of the git repo
- am "future-proofing" by giving us a place to manually update the SDK when needed

jonparrott:
Fair enough, just wanted to make you aware of it. I'm fine with scripts as long as @nathanielmanistaatgoogle is fine with it.

Also, in terms of the GAE SDK - see https://github.com/GoogleCloudPlatform/python-repo-tools/blob/master/gcp_python_repo_tools/appengine_sdk.py - we're creating this in tandem with re-organizing the tests over at python-docs-samples. Eventually we should use that here to keep the SDK up-to-date (once I add updating functionality to it).

dhermes:
Nice! I'm very happy to let you handle it.

nathanielmanistaatgoogle:
As for my own being fine with scripts: uh... I really don't want to have to be. Look at the way this change adds so much more than it removes. How backed-into-a-corner are we?

dhermes:
OK I nuked the scripts. There should still be some unification done in `build_docs.sh` and `GAE_PYTHONPATH` should be used with `fetch_gae_sdk` / `update_gae_sdk` in a consistent way.

tox.ini
```diff
@@ -111,11 +111,10 @@ deps = {[testenv]basedeps}
 commands =
     nosetests \
       --with-gae \
+      --gae-lib-root={env:GAE_PYTHONPATH:google_appengine} \
```
dhermes, line 114: @jonparrott I had `/usr/local/google_appengine` symlink-ed to `${HOME}/google-cloud-sdk/platform/google_appengine` on one machine and not at all on another. Since that is the [default](https://github.com/Trii/NoseGAE/blob/b7c91f8c45d431b7b9fdcca33f40546a2bc12fe0/nosegae.py#L37), I got spurious local success that failed on Travis (and on my other machine).

dhermes:
@jonparrott PTAL. It turns out a `{` inside of a `{` is a no-no. I tried `{env:GAE_PYTHONPATH:{toxinidir}/google_appengine}` and it failed in strange ways.

jonparrott:
Yeah `toxinidir` is almost always `cwd`, unless you explicitly change it.

jonparrott:
This LGTM.

PR #416, opened 2016-02-17
GCE only system test

dhermes:
~~NOTE: Has #415 as a diffbase.~~

### DIFF
--- a/scripts/run_gce_system_tests.py
+++ b/scripts/run_gce_system_tests.py
@@ -1,0 +1,58 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+import unittest
+
+import httplib2
+from six.moves import http_client
+from six.moves import urllib
+
+from oauth2client import GOOGLE_TOKEN_INFO_URI
+from oauth2client.client import GoogleCredentials
+from oauth2client.contrib.gce import AppAssertionCredentials
+
+
+class TestComputeEngine(unittest.TestCase):
+
+    def test_application_default(self):
+        default_creds = GoogleCredentials.get_application_default()
+        self.assertIsInstance(default_creds, AppAssertionCredentials)
+
+    def test_token_info(self):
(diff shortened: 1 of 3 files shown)


scripts/run_gce_system_tests.py
```diff
@@ -0,0 +1,51 @@
...
+import unittest
+
+import httplib2
+from six.moves import http_client
+
+from oauth2client.client import GoogleCredentials
+from oauth2client.contrib.gce import AppAssertionCredentials
+
+
+SCOPE = ('https://www.googleapis.com/auth/plus.login',
```
theacodes: what's this used for?

dhermes: Copy-pasta. Nice find. (I want #212 to be done so we can turn on `pylint` and 100% coverage so the machines can yell at the humans.)

scripts/run_gce_system_tests.py
```diff
@@ -0,0 +1,51 @@
...
+USER_INFO = 'https://www.googleapis.com/oauth2/v2/userinfo'
+
+
+class TestComputeEngine(unittest.TestCase):
+
+    def test_application_default(self):
+        default_creds = GoogleCredentials.get_application_default()
+        self.assertIsInstance(default_creds, AppAssertionCredentials)
+
+    def test_user_info(self):
```
theacodes: userinfo requires special scopes, but you could use tokeninfo instead.

dhermes: :+1: I was just thinking about this when I deleted the scopes.

theacodes: Example here: https://github.com/google/oauth2client/blob/master/oauth2client/client.py#L988

dhermes: @jonparrott I have to be AFK for a bit. You want to just fork this PR and run with it?

theacodes: How long is a bit? I likely won't be able to do that today, but happy to poke at it tomorrow.

dhermes: 3+ hours. I can jump on it once I get home tonight.

dhermes:
@jonparrott PTAL

jonparrott:
this LGTM, @nathanielmanistaatgoogle?

nathanielmanistaatgoogle:
Why isn't Travis showing the test having been run? Is a `travis.yml` update due in this change?

dhermes:
Travis doesn't run in GCE. This Travis environment will fail anywhere but on GCE.

nathanielmanistaatgoogle:
Oh right...

dhermes:
See my and @jonparrott [previous discussion][1]. (And for future people who make it here.)

[1]: https://github.com/google/oauth2client/pull/387#issuecomment-185349768

dhermes:
OK @jonparrott LMK if I can do anything to help getting the Jenkins test going.

jonparrott:
Will do. Setting it up now.

dhermes:
w00t!

PR #419, opened 2016-02-19
Stop sending scopes in token requests on GCE

dhermes:
~~**NOTE**: Has #411 as diffbase.~~

### DIFF
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -19,6 +19,8 @@ Utilities for making it easier to use OAuth 2.0 on Google Compute Engine.
 
 import json
 import logging
+import warnings
+
 from six.moves import http_client
 from six.moves import urllib
 
@@ -34,7 +36,13 @@ logger = logging.getLogger(__name__)
 
 # URI Template for the endpoint that returns access_tokens.
 META = ('http://metadata.google.internal/computeMetadata/v1/instance/'
-        'service-accounts/default/token{?scope}')
+        'service-accounts/default/token')
+_SCOPES_WARNING = """\
+You have requested explicit scopes to be used with a GCE service account.
+Using this argument will have no effect on the actual scopes for tokens
+requested. These scopes are set at VM instance creation time and
+can't be overridden in the request.
+"""
 
 
 class AppAssertionCredentials(AssertionCredentials):
@@ -51,13 +59,19 @@ class AppAssertionCredentials(AssertionCredentials):
     """
 
     @util.positional(2)
-    def __init__(self, scope, **kwargs):
+    def __init__(self, scope='', **kwargs):
         """Constructor for AppAssertionCredentials
 
         Args:
             scope: string or iterable of strings, scope(s) of the credentials
-                   being requested.
+                   being requested. Using this argument will have no effect on
+                   the actual scopes for tokens requested. These scopes are
+                   set at VM instance creation time and won't change.
         """
+        if scope:
+            warnings.warn(_SCOPES_WARNING)
+        # This is just provided for backwards compatibility, but is not
+        # used by this class.
         self.scope = util.scopes_to_string(scope)
         self.kwargs = kwargs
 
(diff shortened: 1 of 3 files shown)


oauth2client/contrib/gce.py
```diff
@@ -51,13 +59,17 @@ class AppAssertionCredentials(AssertionCredentials):
     """
 
     @util.positional(2)
```
nathanielmanistaatgoogle, line 61: Continuing the discussion started in #411: I'm still working to understand this decorator and how it behaves and why its use is a good thing but I think the right thing for now is to leave this as it is.

oauth2client/contrib/gce.py
```diff
@@ -51,13 +59,17 @@ class AppAssertionCredentials(AssertionCredentials):
...
         Args:
             scope: string or iterable of strings, scope(s) of the credentials
-                   being requested.
+                   being requested. Using this argument will have no effect on
+                   the actual scopes for tokens requested. These scopes are
+                   set at VM instance creation time and won't change.
         """
+        if scope:
+            warnings.warn(_SCOPES_WARNING)
         self.scope = util.scopes_to_string(scope)
```
nathanielmanistaatgoogle, line 75: Continuing the "finding ourselves in a hole" discussion from #411: the `scope` attribute now looks silly from a class design point of view in that the class takes it from users, creates a derived product of it that its users could have just as easily created themselves and that may be identical to the original value, and then turns around and sets that derived value on a public instance-scope attribute so that it is available again to the users that supplied it in case they misplaced it. If this were a brand-new class, we'd say to never do that attribute assignment, right? So is that attribute assignment there just to honor existing users who may potentially be out there now making use of the `scope` attribute of their `AppAssertionCredentials` instances? If so, is that worth an implementation comment?

The `scope` instance attribute plays no role in this class' inheritance hierarchy or inherited behavior, right? Right? Pretty please with sugar on top please please please?

oauth2client/contrib/gce.py
dhermes, line 61: OK. The decorator only has [an opinion](https://github.com/google/oauth2client/blob/0c7996099f9e3d74f7b9a6d32b6f69f44987638a/oauth2client/util.py#L48) about the **max** number of positional, so we are good to make `scope` optional?

oauth2client/contrib/gce.py
dhermes, line 75: This class had no children, only parents. And as a child at the bottom of the hierarchy, it has no use for scope because of GCE (essentially what is in the warning message).

I can add a comment. How does this work:

``` python
# This is just provided for backwards compatibility, but is not used by this class.
```

oauth2client/contrib/gce.py
```diff
@@ -107,7 +117,7 @@ def serialization_data(self):
             'Cannot serialize credentials for GCE service accounts.')
 
     def create_scoped_required(self):
-        return not self.scope
+        return False
```
dhermes, line 122: @nathanielmanistaatgoogle Note this has also changed since obviously no scopes are ever required (still won't prevent GCE tokens from not having the right scopes, it's just not something this class can do anything about).

oauth2client/contrib/gce.py
nathanielmanistaatgoogle, line 75: Looks like a good comment; thanks.

oauth2client/contrib/gce.py
nathanielmanistaatgoogle, line 61: Yes, I think so.

oauth2client/contrib/gce.py
```diff
@@ -51,13 +59,19 @@ class AppAssertionCredentials(AssertionCredentials):
...
             scope: string or iterable of strings, scope(s) of the credentials
-                   being requested.
+                   being requested. Using this argument will have no effect on
+                   the actual scopes for tokens requested. These scopes are
+                   set at VM instance creation time and won't change.
         """
+        if scope:
+            warnings.warn(_SCOPES_WARNING)
+        # This is just provided for backwards compatibility, but is not
+        # used by this class.
```
dhermes, line 74: @nathanielmanistaatgoogle Added comment

dhermes:
@nathanielmanistaatgoogle What remains to be done here?

nathanielmanistaatgoogle:
To my knowledge nothing remains to be done here. Merge?

dhermes:
Good deal.

PR #420, opened 2016-02-19
Implementing ServiceAccountCredentials.create_delegated().

### LINKED ISSUE #418: ServiceAccountCredentials needs a way to provide the 'sub' parameter
'sub' is required to do domain-wide delegation of authority. The historical sample was:

```python
credentials = SignedJwtAssertionCredentials(client_email, private_key,
    'https://www.googleapis.com/auth/sqlservice.admin',
    sub='user@example.org')
```

Right now the only decent way to do this with the factory method is to do:

```python
credentials._kwargs['sub'] = 'user@example.org'
```

We should either extend the factory methods to accept `kwargs` or add a `create_delegated` method.

/cc @dhermes

dhermes:
Fixes #418.

FYI @nathanielmanistaatgoogle I'd love to remove the `**kwargs` from the `ServiceAccountCredentials` constructor. AFAIK this is the only extra field that people use.

### DIFF
--- a/oauth2client/service_account.py
+++ b/oauth2client/service_account.py
@@ -407,3 +407,38 @@ class ServiceAccountCredentials(AssertionCredentials):
         result._private_key_pkcs12 = self._private_key_pkcs12
         result._private_key_password = self._private_key_password
         return result
+
+    def create_delegated(self, sub):
+        """Create credentials that act as domain-wide delegation of authority.
+
+        Use the ``sub`` parameter as the subject to delegate on behalf of
+        that user.
+
+        For example::
+
+          >>> account_sub = 'foo@email.com'
+          >>> delegate_creds = creds.create_delegated(account_sub)
+
+        Args:
+            sub: string, An email address that this service account will
+                 act on behalf of (via domain-wide delegation).
+
+        Returns:
+            ServiceAccountCredentials, a copy of the current service account
+            updated to act on behalf of ``sub``.
+        """
+        new_kwargs = dict(self._kwargs)
+        new_kwargs['sub'] = sub
+        result = self.__class__(self._service_account_email,
+                                self._signer,
+                                scopes=self._scopes,
+                                private_key_id=self._private_key_id,
+                                client_id=self.client_id,
+                                user_agent=self._user_agent,
+                                **new_kwargs)
+        result.token_uri = self.token_uri
+        result.revoke_uri = self.revoke_uri
+        result._private_key_pkcs8_pem = self._private_key_pkcs8_pem
+        result._private_key_pkcs12 = self._private_key_pkcs12
+        result._private_key_password = self._private_key_password
+        return result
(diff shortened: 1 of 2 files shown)


oauth2client/service_account.py
```diff
@@ -407,3 +407,38 @@ def create_scoped(self, scopes):
...
+
+        Args:
+            sub: string, An email address that this service account will
+                 act on behalf of (via domain-wide delegation).
+
+        Returns:
+            ServiceAccountCredentials, a copy of the current service account
+            updated to act on behalf of ``sub``.
+        """
+        new_kwargs = copy.copy(self._kwargs)
```
theacodes: Why not `self._kwargs.copy()`?

jonparrott:
I'm wondering if we should renamed `_kwargs` to something better and make it public, e.g. `credentials.assertion_data`.

oauth2client/service_account.py
dhermes: My brain. `copy.copy` isn't even a deep copy, wonder what I was thinking.

dhermes:
I vote we just remove `_kwargs` in total.

jonparrott:
> I vote we just remove _kwargs in total.

I'm for it, but being familiar with the sense of humor of the universe we'd probably have to release `2.0.2` when the one person doing weird things with the assertion data lets us know. :stuck_out_tongue:

dhermes:
I think that's OK. We won't be deleting the old versions that support it.

jonparrott:
Fair enough. This LGTM once you sort out the `copy`.

dhermes:
Sorted out the `.copy()`, folded it into the original commit. Will wait for some more eyes before committing.

oauth2client/service_account.py
```diff
@@ -407,3 +407,38 @@ def create_scoped(self, scopes):
...
+
+        Args:
+            sub: string, An email address that this service account will
+                 act on behalf of (via domain-wide delegation).
+
+        Returns:
+            ServiceAccountCredentials, a copy of the current service account
+            updated to act on behalf of ``sub``.
+        """
+        new_kwargs = self._kwargs.copy()
```
nathanielmanistaatgoogle: After the back-and-forth that the two of you had, how much would you dislike me if I asked for `new_kwargs = dict(self._kwargs)`? That would be what I'd prefer if they're all the same to you.

dhermes: Fine with me. Any other comments or concerns?

dhermes: Folded into the original commit.

nathanielmanistaatgoogle:
Looks good; thanks for waiting on me.

Travis is slow; either for everyone or perhaps just the `google` org. Yesterday I waited ten hours for results that when they ran took three minutes. I promise no pushback if you have run the tests several times locally and choose to merge before the official premerge test results decide to come around.

dhermes:
It's just the `google` org. `GoogleCloudPlatform` has no issues. Will run once more locally before merging

dhermes:
OK everything passing locally. Merging.

PR #421, opened 2016-02-19
Adding common sign_blob() service account types.

dhermes:
Also adding `service_account_email()` property.

See https://github.com/GoogleCloudPlatform/gcloud-python/issues/922 for some details on why `sign_blob` won't work on GCE.

I'm a bit worried about `service_account_id` on the GAE credentials class. In the [implementation](https://code.google.com/p/googleappengine/source/browse/trunk/python/google/appengine/api/app_identity/app_identity.py?r=530#497) it seems to be **almost** synonymous with the svc. account email / name, but not quite (i.e. there may be some legacy behavior). ISTM the reason it's in the constructor is so that it could be [passed to `app_identity`](https://github.com/google/oauth2client/blob/a16a4785c243965aef49065062368781a7caeca3/oauth2client/contrib/appengine.py#L196-L197). Anyhow, our `service_account_name` could either shadow or conflict with `service_account_id` if a user passed one in. Though I don't imagine anyone in the history of this library has passed in a value for `service_account_id`.

Context: We use these utilities in a haphazard way in `gcloud-python` when creating a signed URL (for cloud storage objects). This just puts the auth utils in the auth library in a unified way.

/cc @tseaver @jgeewax @jonparrott

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -1617,6 +1617,18 @@ class AssertionCredentials(GoogleCredentials):
         """
         self._do_revoke(http_request, self.access_token)
 
+    def sign_blob(self, blob):
+        """Cryptographically sign a blob (of bytes).
+
+        Args:
+            blob: bytes, Message to be signed.
+
+        Returns:
+            tuple, A pair of the private key ID used to sign the blob and
+            the signed contents.
+        """
+        raise NotImplementedError('This method is abstract.')
+
 
 def _RequireCryptoOrDie():
     """Ensure we have a crypto library, or throw CryptoUnavailableError.
--- a/oauth2client/contrib/appengine.py
+++ b/oauth2client/contrib/appengine.py
@@ -166,6 +166,7 @@ class AppAssertionCredentials(AssertionCredentials):
         self.scope = util.scopes_to_string(scope)
         self._kwargs = kwargs
         self.service_account_id = kwargs.get('service_account_id', None)
+        self._service_account_email = None
 
         # Assertion type is no longer used, but still in the
         # parent class signature.
@@ -209,6 +210,34 @@ class AppAssertionCredentials(AssertionCredentials):
 
     def create_scoped(self, scopes):
         return AppAssertionCredentials(scopes, **self._kwargs)
+
+    def sign_blob(self, blob):
+        """Cryptographically sign a blob (of bytes).
+
+        Implements abstract method
+        :meth:`oauth2client.client.AssertionCredentials.sign_blob`.
+
+        Args:
+            blob: bytes, Message to be signed.
+
+        Returns:
+            tuple, A pair of the private key ID used to sign the blob and
+            the signed contents.
+        """
+        return app_identity.sign_blob(blob)
+
+    @property
+    def service_account_email(self):
+        """Get the email for the current service account.
+
+        Returns:
+            string, The email associated with the Google App Engine
+            service account.
+        """
+        if self._service_account_email is None:
+            self._service_account_email = (
+                app_identity.get_service_account_name())
+        return self._service_account_email
 
 
 class FlowProperty(db.Property):
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -21,6 +21,7 @@ import json
 import logging
 import warnings
 
+import httplib2
 from six.moves import http_client
 from six.moves import urllib
 
@@ -35,14 +36,40 @@ __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 logger = logging.getLogger(__name__)
 
 # URI Template for the endpoint that returns access_tokens.
-META = ('http://metadata.google.internal/computeMetadata/v1/instance/'
-        'service-accounts/default/token')
+_METADATA_ROOT = ('http://metadata.google.internal/computeMetadata/v1/'
+                  'instance/service-accounts/default/')
+META = _METADATA_ROOT + 'token'
+_DEFAULT_EMAIL_METADATA = _METADATA_ROOT + 'email'
 _SCOPES_WARNING = """\
 You have requested explicit scopes to be used with a GCE service account.
 Using this argument will have no effect on the actual scopes for tokens
 requested. These scopes are set at VM instance creation time and
 can't be overridden in the request.
 """
+
+
+def _get_service_account_email(http_request=None):
+    """Get the GCE service account email from the current environment.
+
+    Args:
+        http_request: callable, (Optional) a callable that matches the method
+                      signature of httplib2.Http.request, used to make
+                      the request to the metadata service.
+
+    Returns:
+        tuple, A pair where the first entry is an optional response (from a
+        failed request) and the second is service account email found (as
+        a string).
+    """
+    if http_request is None:
+        http_request = httplib2.Http().request
+    response, content = http_request(
+        _DEFAULT_EMAIL_METADATA, headers={'Metadata-Flavor': 'Google'})
+    if response.status == http_client.OK:
+        content = _from_bytes(content)
+        return None, content
+    else:
+        return response, content
 
 
 class AppAssertionCredentials(AssertionCredentials):
@@ -78,6 +105,7 @@ class AppAssertionCredentials(AssertionCredentials):
         # Assertion type is no longer used, but still in the
         # parent class signature.
         super(AppAssertionCredentials, self).__init__(None)
+        self._service_account_email = None
 
     @classmethod
     def from_json(cls, json_data):
@@ -123,3 +151,44 @@ class AppAssertionCredentials(AssertionCredentials):
 
     def create_scoped(self, scopes):
         return AppAssertionCredentials(scopes, **self.kwargs)
+
+    def sign_blob(self, blob):
+        """Cryptographically sign a blob (of bytes).
+
+        This method is provided to support a common interface, but
+        the actual key used for a Google Compute Engine service account
+        is not available, so it can't be used to sign content.
+
+        Args:
+            blob: bytes, Message to be signed.
+
+        Raises:
+            NotImplementedError, always.
+        """
+        raise NotImplementedError(
+            'Compute Engine service accounts cannot sign blobs')
+
+    @property
+    def service_account_email(self):
+        """Get the email for the current service account.
+
+        Uses the Google Compute Engine metadata service to retrieve the email
+        of the default service account.
+
+        Returns:
+            string, The email associated with the Google Compute Engine
+            service account.
+
+        Raises:
+            AttributeError, if the email can not be retrieved from the Google
+            Compute Engine metadata service.
+        """
+        if self._service_account_email is None:
+            failure, email = _get_service_account_email()
+            if failure is None:
+                self._service_account_email = email
+            else:
+                raise AttributeError('Failed to retrieve the email from the '
+                                     'Google Compute Engine metadata service',
+                                     failure, email)
+        return self._service_account_email
--- a/oauth2client/service_account.py
+++ b/oauth2client/service_account.py
@@ -320,10 +320,27 @@ class ServiceAccountCredentials(AssertionCredentials):
                                      key_id=self._private_key_id)
 
     def sign_blob(self, blob):
+        """Cryptographically sign a blob (of bytes).
+
+        Implements abstract method
+        :meth:`oauth2client.client.AssertionCredentials.sign_blob`.
+
+        Args:
+            blob: bytes, Message to be signed.
+
+        Returns:
+            tuple, A pair of the private key ID used to sign the blob and
+            the signed contents.
+        """
         return self._private_key_id, self._signer.sign(blob)
 
     @property
     def service_account_email(self):
+        """Get the email for the current service account.
+
+        Returns:
+            string, The email associated with the service account.
+        """
         return self._service_account_email
 
     @property
--- a/tests/contrib/test_appengine.py
+++ b/tests/contrib/test_appengine.py
@@ -116,13 +116,28 @@ class TestAppAssertionCredentials(unittest.TestCase):
 
     class AppIdentityStubImpl(apiproxy_stub.APIProxyStub):
 
-        def __init__(self):
+        def __init__(self, key_name=None, sig_bytes=None,
+                     svc_acct=None):
             super(TestAppAssertionCredentials.AppIdentityStubImpl,
                   self).__init__('app_identity_service')
+            self._key_name = key_name
+            self._sig_bytes = sig_bytes
+            self._sign_calls = []
+            self._svc_acct = svc_acct
+            self._get_acct_name_calls = 0
 
         def _Dynamic_GetAccessToken(self, request, response):
             response.set_access_token('a_token_123')
             response.set_expiration_time(time.time() + 1800)
+
+        def _Dynamic_SignForApp(self, request, response):
+            response.set_key_name(self._key_name)
+            response.set_signature_bytes(self._sig_bytes)
+            self._sign_calls.append(request.bytes_to_sign())
+
+        def _Dynamic_GetServiceAccountName(self, request, response):
+            response.set_service_account_name(self._svc_acct)
+            self._get_acct_name_calls += 1
 
     class ErroringAppIdentityStubImpl(apiproxy_stub.APIProxyStub):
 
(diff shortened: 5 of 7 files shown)


oauth2client/client.py
```diff
@@ -1617,6 +1617,23 @@ def _revoke(self, http_request):
...
+        Returns:
+            tuple, A pair of the private key ID used to sign the blob and
+            the signed contents.
+        """
+        _abstract()
+
+    @property
+    def service_account_email(self):
+        """Get the email for the current service account."""
+        _abstract()
```
nathanielmanistaatgoogle: Just `raise NotImplementedError()`. Consistency is often worthwhile but not in this specific case.

oauth2client/client.py
```diff
@@ -1617,6 +1617,23 @@ def _revoke(self, http_request):
...
+        """Cryptographically sign a blob (of bytes).
+
+        Args:
+            blob: bytes, Message to be signed.
+
+        Returns:
+            tuple, A pair of the private key ID used to sign the blob and
+            the signed contents.
+        """
+        _abstract()
```
nathanielmanistaatgoogle: Just `raise NotImplementedError()` here too.

dhermes: I agree. Thanks.

oauth2client/client.py
```diff
@@ -1617,6 +1617,23 @@ def _revoke(self, http_request):
...
+
+    @property
+    def service_account_email(self):
```
nathanielmanistaatgoogle: Why should this method be added to `AssertionCredentials`? Why is it the case that every `AssertionCredentials` instance will always have a service account email?

dhermes: I was on the fence on this one, but this abstract class has three subclasses and they all have a `service_account_email`, so I went with putting it here as a signal that `isinstance(foo, AppAssertionCredentials)` would mean `foo.service_account_email` is there.

oauth2client/client.py
dhermes: Done

oauth2client/client.py
dhermes: Done

oauth2client/client.py
```diff
@@ -1617,6 +1617,23 @@ def _revoke(self, http_request):
...
+        Returns:
+            tuple, A pair of the private key ID used to sign the blob and
+            the signed contents.
+        """
+        raise NotImplementedError
+
+    @property
+    def service_account_email(self):
+        """Get the email for the current service account."""
+        raise NotImplementedError
```
dhermes: @nathanielmanistaatgoogle To resurface the discussion that got buried. From you:

> Why should this method be added to `AssertionCredentials`? Why is it the case that every `AssertionCredentials` instance will always have a service account email?

From me:

> I was on the fence on this one, but this abstract class has three subclasses and they all have a `service_account_email`, so I went with putting it here as a signal that `isinstance(foo, AppAssertionCredentials)` would mean `foo.service_account_email` is there.

nathanielmanistaatgoogle: I'm not thrilled about `AssertionCredentials` being a public class, but since it is, we have to consider not just our own library's known subclasses but also all possible subclasses that may be out there that others have implemented. I suspect that the only basis we could ever have for being certain that this property would make sense for all `AppAssertionCredentials` everywhere would come from something about the `AppAssertionCredentials` class itself. But there's nothing here in the class motivating that - it's just coincidence that our three subclasses happen to have this other quality in common. So I think this abstract method should be dropped from this change.

There's one other thing we _could_ do, though: if we believe that there's a fundamental "sameness" to the way our three classes present their service account email and we'd like to present that sameness as part of the library's class taxonomy, we could define a separate purely-virtual `class ServiceAccountEmailHaver` (final name TBD :-P) and have our three classes implement that.

oauth2client/client.py
```diff
@@ -1617,6 +1617,23 @@ def _revoke(self, http_request):
...
+        Args:
+            blob: bytes, Message to be signed.
+
+        Returns:
+            tuple, A pair of the private key ID used to sign the blob and
+            the signed contents.
+        """
+        raise NotImplementedError
+
+    @property
```
nathanielmanistaatgoogle: Perhaps a moot point by now but for my own curiosity: why `@property` and not `@abstractproperty`?

Fun fact about me: I don't like, and dispute the premise of, both `abstractproperty` and `property`.

dhermes: There is no such thing as `abstractproperty`?

oauth2client/client.py
dhermes: I'll just delete the property, seems easier :grinning:

dhermes:
@nathanielmanistaatgoogle Removed the abstract `service_account_email` property. What remains to be reviewed?

oauth2client/client.py
nathanielmanistaatgoogle: Slight confusion on my part; sorry. [Here's abstractproperty](https://docs.python.org/2/library/abc.html#abc.abstractproperty).

oauth2client/client.py
```diff
@@ -1617,6 +1617,18 @@ def _revoke(self, http_request):
...
+        """Cryptographically sign a blob (of bytes).
+
+        Args:
+            blob: bytes, Message to be signed.
+
+        Returns:
+            tuple, A pair of the private key ID used to sign the blob and
+            the signed contents.
+        """
+        raise NotImplementedError
```
nathanielmanistaatgoogle: I prefer the explicit `()` in `raise NotImplementedError()`; humor me if you don't care either way?

oauth2client/contrib/appengine.py
```diff
@@ -210,6 +211,35 @@ def create_scoped_required(self):
     def create_scoped(self, scopes):
         return AppAssertionCredentials(scopes, **self._kwargs)
 
+    def sign_blob(self, blob):
+        """Cryptographically sign a blob (of bytes).
```
nathanielmanistaatgoogle, line 215: This doesn't differ at all from the overridden method doc string, right? It's a straight copy/paste?

How about `"""See AppAssertionCredentials.sign_blob for specification."""` instead?

oauth2client/contrib/appengine.py
```diff
@@ -210,6 +211,35 @@ def create_scoped_required(self):
...
+            tuple, A pair of the private key ID used to sign the blob and
+            the signed contents.
+        """
+        return app_identity.sign_blob(blob)
+
+    @property
+    def service_account_email(self):
+        """Get the email for the current service account.
+
+        Uses the ``app_identity`` service to retrieve the service
```
nathanielmanistaatgoogle: This is a lot of internal implementation to promise to external users of this method - is that intended? Why do users of this method need to be guaranteed that the method is implemented this way?

oauth2client/contrib/gce.py
```diff
@@ -123,3 +127,60 @@ def create_scoped_required(self):
 
     def create_scoped(self, scopes):
         return AppAssertionCredentials(scopes, **self.kwargs)
+
+    def sign_blob(self, blob):
+        """Cryptographically sign a blob (of bytes).
+
+        This method is provided to support a common interface, but
```
nathanielmanistaatgoogle, line 158: The "common interface" part of this is no longer true, correct?

oauth2client/contrib/gce.py
```diff
@@ -123,3 +127,60 @@ def create_scoped_required(self):
...
+        the actual key used for a Google Compute Engine service account
+        is not available, so it can't be used to sign content.
+
+        Args:
+            blob: bytes, Message to be signed.
+
+        Raises:
+            NotImplementedError, always.
```
nathanielmanistaatgoogle, line 166: This method is starting to smell like it shouldn't exist? Or should it exist, and something's gone over my head?

oauth2client/contrib/gce.py
```diff
@@ -123,3 +127,60 @@ def create_scoped_required(self):
...
+        """
+        raise NotImplementedError(
+            'Compute Engine service accounts cannot sign blobs')
+
+    @property
+    def service_account_email(self):
+        """Get the email for the current service account.
+
+        Uses the Google Compute Engine metadata service to retrieve the email
```
nathanielmanistaatgoogle, line 175: Reconsider this - I'm open to it being important for users to know how the value is obtained by this object, but the caching behavior again seems like an unnecessary promise.

oauth2client/contrib/gce.py
```diff
@@ -123,3 +127,60 @@ def create_scoped_required(self):
...
+        Returns:
+            string, The email associated with the Google Compute Engine
+            service account.
+        """
+        if self._service_account_email is None:
+            self._service_account_email = _get_service_account_email()
+        return self._service_account_email
+
+
+def _get_service_account_email(http_request=None):
```
nathanielmanistaatgoogle: Please locate this newly-introduced code element earlier in the file than any of its use sites.

oauth2client/service_account.py
```diff
@@ -320,10 +320,24 @@ def _generate_assertion(self):
                                      key_id=self._private_key_id)
 
     def sign_blob(self, blob):
+        """Cryptographically sign a blob (of bytes).
```
nathanielmanistaatgoogle, line 323: `"""See AppAssertionCredentials.sign_blob for specification."""` here too, right?

Copy/pasted documentation always invites tragedy. Tragedy sometimes has a more important meeting to attend and doesn't show up, but it is at least _invited_.

nathanielmanistaatgoogle:
(Completed a full review pass.)

oauth2client/contrib/appengine.py
dhermes, line 215: The over-ridden method doesn't do anything though. That could be confusing.

oauth2client/contrib/appengine.py
dhermes: Side-bar removed.

oauth2client/contrib/gce.py
dhermes, line 158: Well `creds.sign_blob(blob)` is the common interface, it just never crosses the finish line by allowing

``` python
key_id, signed_bytes = creds.sign_blob(blob)
```

oauth2client/contrib/gce.py
dhermes, line 166: The desire is for `credentials.sign_blob(blob)` to be possible. Essentially it's choosing between `NotImplementedError` and `AttributeError`. Which duck is best?

The reason I did it this way was so that users could get a helpful message

```
Compute Engine service accounts cannot sign blobs
```

rather than just an `AttributeError`. (In `gcloud-python` we have a [note](https://github.com/GoogleCloudPlatform/gcloud-python/blob/a77e7fae62237edbba25b5da37deb673f6ef92b5/gcloud/storage/blob.py#L158-L166) about this in our docstring.)

oauth2client/contrib/gce.py
dhermes, line 175: Axed the caching part.

oauth2client/contrib/gce.py
dhermes: Done.

oauth2client/service_account.py
dhermes, line 323: Ditto on previous comment.

oauth2client/contrib/gce.py
```diff
@@ -45,6 +48,31 @@
...
+    """
+    if http_request is None:
+        http_request = httplib2.Http().request
+    response, content = http_request(
+        _DEFAULT_EMAIL_METADATA, headers={'Metadata-Flavor': 'Google'})
+    if response.status == http_client.OK:
+        content = _from_bytes(content)
+        return content
+    else:
+        raise RuntimeError(response, content)
```
dhermes: @nathanielmanistaatgoogle I'm sorry I didn't point this out sooner. I don't think `RuntimeError` is appropriate here, but I didn't want to implement a do-nothing

``` python
class SneezingError(Exception): pass
```

as is the tradition in `oauth2client`. WDYT?

oauth2client/contrib/appengine.py
nathanielmanistaatgoogle, line 215: If it truly did nothing, it would be dead code and we'd eliminate it.

What can an abstract method, the body of which raises `NotImplementedError`, truly be said to **do**? _Specify_ behavior of concrete methods (specifically, those which override the abstract method). And it's to exactly that specification that we want to direct the user of this method. So when we say "See AppAssertionCredentials.sign_blob for specification", we're directing the user of this method to that specification, rather than any method body. We could say "See specification of AppAssertionCredentials.sign_blob for specification of this method", but that would be an unnecessary mouthful.

Have you ever written a Java method and written `// inherited doc is fine` on it rather than giving it a proper JavaDoc specification? That would be along the lines of what we want to do here.

oauth2client/contrib/gce.py
nathanielmanistaatgoogle, line 166: Agreed. `NotImplementedError` is best here.

nathanielmanistaatgoogle:
Looks good except for the copied-and-pasted doc strings.

oauth2client/contrib/gce.py
nathanielmanistaatgoogle: Well, this ultimately gets exposed as the `service_account_email` property, doesn't it? What exception should an object raise if an attempt to access an attribute of it makes a request over the network that fails?

oauth2client/contrib/appengine.py
dhermes, line 215: I'm not sure why Java wisdom applies here? If someone shows up on [the `appengine` docs](http://oauth2client.readthedocs.org/en/latest/source/oauth2client.contrib.appengine.html) page and the doc says "haha, you suck, go [somewhere else](http://oauth2client.readthedocs.org/en/latest/source/oauth2client.client.html)", isn't that a negative experience?

PS It seems the GAE docs are back to not building. Seems to be flaky within RTD, will try to track down.

oauth2client/contrib/gce.py
dhermes: We could have this method return `None` and then have the `@property` return an `AttributeError`?

oauth2client/contrib/appengine.py
nathanielmanistaatgoogle, line 215: Perhaps I shouldn't have mentioned Java... but since I did, here's what happens there: if you leave an overriding method undocumented, Javadoc renders the overridden documentation in its place. So the documentation only has to exist one place in code and its visible in two (or N, as the case might be).

We don't have that for Python, so what I'm pushing for is the "only maintain the documentation in one place" option, coming as it does at the cost of "users have to navigate from one place to another". I think the dangers of having to maintain multiple copies of text generally outweigh the inconvenience of users having to manually navigate from one method to another, so I've adopted a personal rule and habit around it, and am applying that rule here. I see that you feel otherwise in this particular case, but do you also feel otherwise in general?

oauth2client/contrib/gce.py
nathanielmanistaatgoogle: I was cheeky in asking the question; my feeling is that there's no good option for reporting when accessing an attribute of an object triggers a network fault (and now you know the beginning of why I'm against `@property`...).

Given that there's no good option, I think `AttributeError` describing the fault encountered (so including `response` and `content`) is the least bad option.

oauth2client/contrib/appengine.py
dhermes, line 215: I agree that the two definitions will begin to float away from one another. [Reading about](http://stackoverflow.com/questions/2025562/inherit-docstrings-in-python-class-inheritance) ways to inherit docstrings now.

dhermes, line 215: @nathanielmanistaatgoogle After [reading](http://stackoverflow.com/questions/8100166/inheriting-methods-docstrings-in-python) a bit more, I don't think it is worth it given the existing tooling.

oauth2client/contrib/gce.py
```diff
@@ -45,6 +48,30 @@
...
+    """
+    if http_request is None:
+        http_request = httplib2.Http().request
+    response, content = http_request(
+        _DEFAULT_EMAIL_METADATA, headers={'Metadata-Flavor': 'Google'})
+    if response.status == http_client.OK:
+        content = _from_bytes(content)
+        return None, content
+    else:
+        return response, content
```
dhermes, line 72: @nathanielmanistaatgoogle PTAL

oauth2client/contrib/appengine.py
nathanielmanistaatgoogle, line 215: Might we compromise on adding something like "Implements abstract method AppAssertionCredentials.sign_blob" to this doc string? I'm comfortable backpedalling to allowing the copied doc string but there's got to be something informing future users and maintainers about the relationship between this code element and another code element elsewhere.

(If you agree, alter the similar "ditto" copied doc comment elsewhere in this change?)

dhermes:
@nathanielmanistaatgoogle See the 4th commit for the "implements" note.

nathanielmanistaatgoogle:
All looks great; squash and merge. Thanks for the conversation.

dhermes:
:+1: Will wait to merge until Travis is done in 10 hours.

PR #424, opened 2016-02-20
Helpful error message in P12 factory in absence of pyOpenSSL.

### LINKED ISSUE #417: Warn user if .p12 support is not available because of missing crypto libraries
If you try to create `ServiceAccountCredentials` without having `PyOpenSSL` installed, a `UnicodeDecodeError` is raised:
````python
>>> from oauth2client import service_account
>>> service_account.ServiceAccountCredentials.from_p12_keyfile(CLIENT_ID, PATH_TO_P12_KEY_FILE)
  File "<input>", line 1, in <module>
  File "lib/python2.7/site-packages/oauth2client/service_account.py", line 245, in from_p12_keyfile
    private_key_password)
  File "lib/python2.7/site-packages/oauth2client/_pure_python_crypt.py", line 167, in from_string
    key = _from_bytes(key)  # pem expects str in Py3
  File "lib/python2.7/site-packages/oauth2client/_helpers.py", line 87, in _from_bytes
    if isinstance(value, six.binary_type) else value)
  File "lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0x82 in position 1: invalid start byte
````
It would be better to use the existing (but unused, from what I can tell) `_PKCS12_ERROR` string inside `service_account.py` to let the user know what's going on.

### PR DESCRIPTION
**Pull request by dhermes**
Fixes #417.

@nathanielmanistaatgoogle I'm not sure why this didn't make it in the re-write. I definitely had tests for this and an `if crypt.Signer` check.

Sorry @ddoskind that I absent-mindedly let it slip in #398 (it was in #392 but I dropped the `_get_signer` helper since only the factories need to make a signer now rather than the constructor back then)

### DIFF
--- a/oauth2client/service_account.py
+++ b/oauth2client/service_account.py
@@ -240,6 +240,8 @@ class ServiceAccountCredentials(AssertionCredentials):
         """
         if private_key_password is None:
             private_key_password = _PASSWORD_DEFAULT
+        if crypt.Signer is not crypt.OpenSSLSigner:
+            raise NotImplementedError(_PKCS12_ERROR)
         signer = crypt.Signer.from_string(private_key_pkcs12,
                                           private_key_password)
         credentials = cls(service_account_email, signer, scopes=scopes)
--- a/oauth2client/_pycrypto_crypt.py
+++ b/oauth2client/_pycrypto_crypt.py
@@ -120,9 +120,7 @@ class PyCryptoSigner(object):
             pkey = RSA.importKey(parsed_pem_key)
         else:
             raise NotImplementedError(
-                'PKCS12 format is not supported by the PyCrypto library. '
-                'Try converting to a "PEM" '
-                '(openssl pkcs12 -in xxxxx.p12 -nodes -nocerts > '
-                'privatekey.pem) '
-                'or using PyOpenSSL if native code is an option.')
+                'No key in PEM format was detected. This implementation '
+                'can only use the PyCrypto library for keys in PEM '
+                'format.')
         return PyCryptoSigner(pkey)
--- a/tests/test_service_account.py
+++ b/tests/test_service_account.py
@@ -165,6 +165,21 @@ class ServiceAccountCredentialsTests(unittest2.TestCase):
             self.assertEqual(creds._private_key_password, private_key_password)
         self.assertEqual(creds._scopes, ' '.join(scopes))
 
+    def _p12_not_implemented_helper(self):
+        service_account_email = 'name@email.com'
+        filename = data_filename('privatekey.p12')
+        with self.assertRaises(NotImplementedError):
+            ServiceAccountCredentials.from_p12_keyfile(
+                service_account_email, filename)
+
+    @mock.patch('oauth2client.crypt.Signer', new=crypt.PyCryptoSigner)
+    def test_from_p12_keyfile_with_pycrypto(self):
+        self._p12_not_implemented_helper()
+
+    @mock.patch('oauth2client.crypt.Signer', new=crypt.RsaSigner)
+    def test_from_p12_keyfile_with_rsa(self):
+        self._p12_not_implemented_helper()
+
     def test_from_p12_keyfile_defaults(self):
         self._from_p12_keyfile_helper()
 

### CODE REVIEW

**Comment by nathanielmanistaatgoogle** [comment]
Looks good; merge when you please.

File: oauth2client/service_account.py [hunk 1]
```diff
@@ -240,6 +240,8 @@ def _from_p12_keyfile_contents(cls, service_account_email,
         """
         if private_key_password is None:
             private_key_password = _PASSWORD_DEFAULT
+        if crypt.Signer is not crypt.OpenSSLSigner:
```
codingjoe, line 243: This line breaks the `RsaSigner`

codingjoe:
Hi,

this seems to have introduced a bug. You can no longer use any other signer as the OpenSSL one. 

This seems strange, is that intended behavior?

Cheers,
Joe

dhermes:
The `rsa` library (and the ASN.1 helpers) were never able to parse P12 keys, but they work just fine with JSON keys.

1. Did you actually have them working with P12?
1. Do you have a need for P12 keys instead of JSON?

oauth2client/service_account.py
dhermes, line 243: Expand the diff, you'll see this is only in the `_from_p12_keyfile_contents` helper

codingjoe:
I use Django on Heroku. On Heroku you can't store files except in your repository. Adding a private key to the repository is not the best idea I guess ;)
That's why I added it to the environment.

I wrote that little bit of code that worked for me until this change:
```python
def build_googleapiclient(service_email=None, private_key=None,
                          user_email=None):
    service_email = service_email or settings.GOOGLE_DRIVE_STORAGE_SERVICE_EMAIL
    key = private_key or settings.GOOGLE_DRIVE_STORAGE_KEY

    credentials = ServiceAccountCredentials._from_p12_keyfile_contents(
        service_email,
        key,
        scopes="https://www.googleapis.com/auth/drive",
    )
    credentials = credentials.create_delegated(user_email)
    http = httplib2.Http()
    http = credentials.authorize(http)

    return build('drive', 'v2', http=http)
```

I know it's not nice to use a private API and than complain, but hey ;)

Since my private key start with `-----BEGIN RSA PRIVATE KEY-----` I naturally assumed it's an RSA key and not P12.

dhermes:
`-----BEGIN RSA PRIVATE KEY-----` is not a P12 key, it is likely a PCKS#1 key in PEM format (and if it's from Google you probably got it by converting a P12 key to it).

`rsa` and `crypt.RsaSigner` are just libraries, they don't (necessarily) relate to the key type. Any of the three possible signers will work with that key.

I'd first and foremost just recommend getting a JSON key file from Google, since it's unclear where you got the PKCS#1 key from. (A JSON keyfile is a JSON-ified dict, with a field that holds a PKCS#8 key in PEM format.)

If you want to proceed "as-is" by just constructing any old signer:

```python
signer = crypt.Signer.from_string(settings.GOOGLE_DRIVE_STORAGE_KEY)
creds = ServiceAccountCredentials(
    settings.GOOGLE_DRIVE_STORAGE_SERVICE_EMAIL, signer,
    scopes='https://www.googleapis.com/auth/drive',
    sub=user_email)
# OPTIONAL HACKY SNIPPET FOR SERIALIZATION
creds._private_key_pkcs8_pem = settings.GOOGLE_DRIVE_STORAGE_KEY 
# DEFINITELY A LIE SINCE YOUR STORAGE KEY IS PKCS#1, NOT PKCS#8
```

----

Can you give an idea / snippet of what you were doing before (with `SignedJwtAssertionCredentials`, before the 2.0.0 release)?

codingjoe:
@dhermes: sure, reversion says we used this one :)

Is the JSON key file available now for server2server communication? We needed a user that can modify out Drive regardless of the current user.
```python
def build_googleapiclient(service_email=None, private_key=None,
                          user_email=None):
    service_email = service_email or settings.GOOGLE_DRIVE_STORAGE_SERVICE_EMAIL
    key = private_key or settings.GOOGLE_DRIVE_STORAGE_KEY

    kwargs = {}
    if user_email:
        kwargs['sub'] = user_email
    credentials = SignedJwtAssertionCredentials(
        service_email,
        key,
        scope="https://www.googleapis.com/auth/drive",
        **kwargs
    )
    http = httplib2.Http()
    http = credentials.authorize(http)

    return build('drive', 'v2', http=http)
```

dhermes:
The key you have is fine. Yes a JSON keyfile is  a keyfile for a Google service account.

The snippets I mentioned above will work for you (I'm 99.9% sure your PKCS#1 key is just converted from an original P12 key downloaded from Google).

codingjoe:
@dhermes thanks! I just saw that you can now create json files for service accounts. I just created new creds for the account in json:
```python
def build_googleapiclient(user_email=None):
    key = json.loads(settings.GOOGLE_DRIVE_STORAGE_KEYP)

    credentials = ServiceAccountCredentials._from_parsed_json_keyfile(
        key,
        scopes="https://www.googleapis.com/auth/drive",
    )
    if user_email is not None:
        credentials = credentials.create_delegated(user_email)
    http = httplib2.Http()
    http = credentials.authorize(http)

    return build('drive', 'v2', http=http)
```
This works fine for me. Forget the rest ;)

dhermes:
Good deal! I strongly encourage you just use

```python
ServiceAccountCredentials.from_json_keyfile_name(
    settings.GOOGLE_DRIVE_STORAGE_KEYP,
    scopes='https://www.googleapis.com/auth/drive')
```

instead of using a non-public interface.

codingjoe:
@dhermes that is a nice Idea, but again, I can't store files on Heroku that are not in the git repository.
You're not encouraging me to put a PK in my repo are ;)

dhermes:
Sorry I misread the `json.loads` as `json.load`. I wasn't thinking about your environment.

Still, you don't need the non-public method. Use http://oauth2client.readthedocs.org/en/latest/source/oauth2client.service_account.html#oauth2client.service_account.ServiceAccountCredentials.from_json_keyfile_dict

codingjoe:
Good point, thx!

PR #425, opened 2016-02-20
Fix GCE system tests

theacodes:
tokeninfo will not tell you the email address for a GCE account. It's sufficient to verify that the token is valid and unexpired.

### DIFF
--- a/scripts/run_gce_system_tests.py
+++ b/scripts/run_gce_system_tests.py
@@ -48,8 +48,6 @@ class TestComputeEngine(unittest.TestCase):
 
         content = content.decode('utf-8')
         payload = json.loads(content)
-        self.assertTrue(payload['email'].endswith(
-            '-compute@developer.gserviceaccount.com'))
         self.assertEqual(payload['access_type'], 'offline')
         self.assertLessEqual(int(payload['expires_in']), 3600)
 


dhermes:
@jonparrott FWIW I ran this on an actual VM and it passed. It must be the case that token info won't have the email address when the email scope isn't turned on when the VM is created?

jonparrott:
Probably. That's the case for the instance that our Jenkins instance is on.

dhermes:
OK. How do you feel about just checking conditionally if `email` is in the payload? (Maybe that's bad, the actual email address is also likely brittle.)

jonparrott:
You're able to rename service accounts, so the email address is brittle. I'm okay with just verifying that it's a valid token - this case would've adequately caught the error that spawned this whole system test.

dhermes:
:+1:

jonparrott:
Jenkins is passing. I'll set it up to email our @google addresses on
failure.

dhermes:
No GitHub integration? I'd be happy to take a stab at setting up a hook?

jonparrott:
We can't because it's firewalled. We're looking into the possibility of
having a separate app handle the hooks. @waprin

dhermes:
SG

PR #426, opened 2016-02-20
Allowing default flags in run_flow().

dhermes:
Resurrects #360. I double checked two different default behaviors to verify this is the "right" way:

``` python
>>> # Method 1
>>> flags = argparser.parse_args()
>>> flags
Namespace(auth_host_name='localhost', auth_host_port=[8080, 8090], logging_level='ERROR', noauth_local_webserver=False)
>>> # Method 2 (same result, but takes more code)
>>> import argparse
>>> parser = argparse.ArgumentParser(parents=[argparser])
>>> flags = parser.parse_args()
>>> flags
Namespace(auth_host_name='localhost', auth_host_port=[8080, 8090], logging_level='ERROR', noauth_local_webserver=False)
```

### DIFF
--- a/oauth2client/tools.py
+++ b/oauth2client/tools.py
@@ -113,7 +113,7 @@ class ClientRedirectHandler(BaseHTTPServer.BaseHTTPRequestHandler):
 
 
 @util.positional(3)
-def run_flow(flow, storage, flags, http=None):
+def run_flow(flow, storage, flags=None, http=None):
     """Core code for a command-line application.
 
     The ``run()`` function is called from your application and runs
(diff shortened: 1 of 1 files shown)


oauth2client/tools.py
```diff
@@ -113,7 +113,7 @@ def log_message(self, format, *args):
 
 
 @util.positional(3)
```
dhermes, line 115: @nathanielmanistaatgoogle As in #360 I'm unsure if ~~this~~ the number of positional args needs to be reduced.

nathanielmanistaatgoogle:
Test coverage?

dhermes:
I was under the assumption that there were no tests for this module, let me double check.

dhermes:
OK I was wrong, but correct in spirit. The [tests][1] for the `tools` module are essentially 0, and testing `run_flow()` would increase the size of this PR by a factor of 20, so it should probably be its own thing.

[1]: https://github.com/google/oauth2client/blob/d5f41392b20599fec55b8ac8611646c1ad93a5d7/tests/test_tools.py

oauth2client/tools.py
nathanielmanistaatgoogle, line 115: I'm pretty sure that there's no change needed here.

dhermes, line 115: :+1:

nathanielmanistaatgoogle:
Okay. I was caught by the -0.03% coverage decrease and for once I looked at it and said "wait, no, that's a real thing and not just noise, this change introduces new code that is definitely not covered".

Merge when you like.

dhermes:
FWIW I **really** want some help to sprint through #212 and make any drop below 100% not allowed

PR #429, opened 2016-02-20
Moving util.dict_to_tuple_key() into only module that uses it.

dhermes:
When I filed #428 I compared the modules to see what was superfluous and noticed that this method only had one user. This change may not be "allowed" since it removes a public interface.

### DIFF
--- a/tests/contrib/test_multistore_file.py
+++ b/tests/contrib/test_multistore_file.py
@@ -48,6 +48,30 @@ class _MockLockedFile(object):
         return self.filename_str
 
 
+class Test__dict_to_tuple_key(unittest2.TestCase):
+
+    def test_key_conversions(self):
+        key1, val1 = 'somekey', 'some value'
+        key2, val2 = 'another', 'something else'
+        key3, val3 = 'onemore', 'foo'
+        test_dict = {
+            key1: val1,
+            key2: val2,
+            key3: val3,
+        }
+        tuple_key = multistore_file._dict_to_tuple_key(test_dict)
+
+        # the resulting key should be naturally sorted
+        expected_output = (
+            (key2, val2),
+            (key3, val3),
+            (key1, val1),
+        )
+        self.assertTupleEqual(expected_output, tuple_key)
+        # check we get the original dictionary back
+        self.assertDictEqual(test_dict, dict(tuple_key))
+
+
 class MultistoreFileTests(unittest2.TestCase):
 
     def tearDown(self):
@@ -257,3 +281,7 @@ class MultistoreFileTests(unittest2.TestCase):
         store2.delete()
         keys = multistore_file.get_all_credential_keys(FILENAME)
         self.assertEquals([], keys)
+
+
+if __name__ == '__main__':  # pragma: NO COVER
+    unittest2.main()
--- a/oauth2client/contrib/multistore_file.py
+++ b/oauth2client/contrib/multistore_file.py
@@ -73,6 +73,21 @@ class NewerCredentialStoreError(Error):
     """The credential store is a newer version than supported."""
 
 
+def _dict_to_tuple_key(dictionary):
+    """Converts a dictionary to a tuple that can be used as an immutable key.
+
+    The resulting key is always sorted so that logically equivalent
+    dictionaries always produce an identical tuple for a key.
+
+    Args:
+        dictionary: the dictionary to use as the key.
+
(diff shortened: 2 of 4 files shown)


dhermes:
For some reason this didn't trigger the Travis hook?

tests/contrib/test_multistore_file.py
```diff
@@ -48,6 +48,24 @@ def filename(self):
         return self.filename_str
 
 
+class Test__dict_to_tuple_key(unittest.TestCase):
```
nathanielmanistaatgoogle: `unittest2.TestCase`, then...

tests/contrib/test_multistore_file.py
```diff
@@ -48,6 +48,24 @@ def filename(self):
...
+
+    def test_key_conversions(self):
+        d = {'somekey': 'some value', 'another': 'something else',
+             'onemore': 'foo'}
+        tuple_key = multistore_file._dict_to_tuple_key(d)
+
+        # the resulting key should be naturally sorted
+        self.assertEqual(
```
nathanielmanistaatgoogle: ... `assertTupleEqual` here and...

tests/contrib/test_multistore_file.py
```diff
@@ -48,6 +48,24 @@ def filename(self):
...
+            (('another', 'something else'),
+             ('onemore', 'foo'),
+             ('somekey', 'some value')),
+            tuple_key)
+
+        # check we get the original dictionary back
+        self.assertEqual(d, dict(tuple_key))
```
nathanielmanistaatgoogle: ... `assertDictEqual` here. Right?

tests/contrib/test_multistore_file.py
dhermes: This test module [still uses](https://github.com/google/oauth2client/blob/1f1821653d63c68e657965da4525f04a846c18d0/tests/contrib/test_multistore_file.py#L22) `unittest` (i.e. has no need for backports). Uses of `unittest2` have been on an as-needed basis. You think it's worth switching over wholesale?

tests/contrib/test_multistore_file.py
dhermes: Ah I'd guess this is a backport.

tests/contrib/test_multistore_file.py
```diff
@@ -48,6 +48,24 @@ def filename(self):
         return self.filename_str
 
 
+class Test__dict_to_tuple_key(unittest.TestCase):
+
+    def test_key_conversions(self):
+        d = {'somekey': 'some value', 'another': 'something else',
```
nathanielmanistaatgoogle: For maximum orthodoxy, never repeat your test data values. Consider:

```
first_key = 'somekey'
<...>
third_value = 'foo'
test_dictionary = {
    first_key: first_value,
    second_key: second_value,
    third_key: third_value,
}
```

and then later:

```
expected_tuple = (
    (first_key, first_value),
    (second_key, second_value),
    (third_key, third_value),
)
```

.

Okay, okay, okay: orthodoxy is fussy and sometimes verbose. But at least develop a habit of _feeling bad_ when repeating test data values?

dhermes:
@nathanielmanistaatgoogle Don't forget to weigh in on whether or not moving this is allowed (since it is a public interface)

tests/contrib/test_multistore_file.py
dhermes: I didn't even read the test. Copy-pasta for life!

nathanielmanistaatgoogle:
I endorse this change, despite it technically regressing the public interface of the library. The code element being withdrawn has nothing to do with authentication or cryptography or security or anything else that is part of the intended surface area of the library.

I was actually just the other day defending another library's many many underscores and highly restricted public interface and I think this a good example of what I was describing in that conversation about how too often things are accidentally public because underscores are cosmetically unattractive and library authors are slow to adopt a habit of "everything hidden by default when first written and then public only after thoughtful consideration that it needs to be public to serve the library's users". :-)

dhermes:
I am a big fan of that approach.

I addressed the review comments. PTAL.

dhermes:
And we even got Travis to fire on the new commit!

nathanielmanistaatgoogle:
Looks great; squash, wait a reasonable time for test results (it is Saturday after all), and merge. :-)

PR #431, opened 2016-02-21
Getting to 100% line coverage in tests.

dhermes:


tests/test_client.py
```diff
@@ -357,17 +347,16 @@ def test_save_to_well_known_file_service_account(self):
         self.assertEqual('ABCDEF', d['private_key_id'])
         os.remove(temp_credential_file)
 
-    def test_save_well_known_file_with_non_existent_config_dir(self):
+    @mock.patch('os.path.isdir', return_value=False)
+    def test_save_well_known_file_with_non_existent_config_dir(self,
```
nathanielmanistaatgoogle, line 351: I don't understand the change on this line. Mustn't `self` be the exactly one parameter of a test method?

tests/test_file.py
```diff
@@ -177,7 +177,7 @@ def test_token_refresh_good_store(self):
         new_cred.access_token = 'bar'
         s.put(new_cred)
 
-        credentials._refresh(lambda x: x)
+        credentials._refresh(None)
```
nathanielmanistaatgoogle, line 180: What's the reason for this change?

tests/test_client.py
dhermes, line 351: `@mock.patch` (which is now in the standard library in Python 3) adds sugar for more arguments. Also, I've used it in PRs that you've reviewed before.

tests/test_file.py
dhermes, line 180: The `good_store` part of `test_token_refresh_good_store` means the refresh method is not needed, so `coverage` was (correctly) flagging `lambda x: x` as something which was never executed.

tests/test_client.py
nathanielmanistaatgoogle, line 351: Alright, I suppose I'll have to go off and learn it then.

dhermes, line 351: It's very useful. It took me awhile but I'm happy I did.

Issue #433, opened 2016-02-23
Code quality: Remove / reduce functions and classes not defined at global scope


dhermes:
Some examples:

- all the [`ndb` helpers][1] in the `appengine` module (changes coming)
- the class created in [`OAuth2Decorator.callback_handler`][2]
- version specific [Django config][3]
- locked file [`fcntl` helpers][4]
- locked file [`win32` helpers][5]
- class-within-a-class storage [in `multistore_file`][6]
- monkey-patched [request method][7] (this is a whole project in itself)

Also worth noting, but probably not worth changing are decorators:

- appengine decorators return values ([`oauth_required`][8] and [`oauth_aware`][9])
- [decorators][10] in `util`
- [decorators][11] in `django_util` ([also][12])
- [decorators][13] in `flask_util`

[1]: https://github.com/google/oauth2client/blob/349942b0db8e96c336802dc421fd4f1670e4b7e8/oauth2client/contrib/appengine.py#L94
[2]: https://github.com/google/oauth2client/blob/349942b0db8e96c336802dc421fd4f1670e4b7e8/oauth2client/contrib/appengine.py#L916
[3]: https://github.com/google/oauth2client/blob/349942b0db8e96c336802dc421fd4f1670e4b7e8/oauth2client/contrib/django_util/apps.py#L29
[4]: https://github.com/google/oauth2client/blob/349942b0db8e96c336802dc421fd4f1670e4b7e8/oauth2client/contrib/locked_file.py#L179
[5]: https://github.com/google/oauth2client/blob/349942b0db8e96c336802dc421fd4f1670e4b7e8/oauth2client/contrib/locked_file.py#L248
[6]: https://github.com/google/oauth2client/blob/349942b0db8e96c336802dc421fd4f1670e4b7e8/oauth2client/contrib/multistore_file.py#L213
[7]: https://github.com/google/oauth2client/blob/349942b0db8e96c336802dc421fd4f1670e4b7e8/oauth2client/client.py#L592
[8]: https://github.com/google/oauth2client/blob/349942b0db8e96c336802dc421fd4f1670e4b7e8/oauth2client/contrib/appengine.py#L762
[9]: https://github.com/google/oauth2client/blob/349942b0db8e96c336802dc421fd4f1670e4b7e8/oauth2client/contrib/appengine.py#L829
[10]: https://github.com/google/oauth2client/blob/349942b0db8e96c336802dc421fd4f1670e4b7e8/oauth2client/util.py#L120-L122
[...]

PR #434, opened 2016-02-23
Move conditionally defined ndb helpers into their own module.

dhermes:
Towards #433.

### DIFF
--- a/oauth2client/contrib/_appengine_ndb.py
+++ b/oauth2client/contrib/_appengine_ndb.py
@@ -1,0 +1,163 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Google App Engine utilities helper.
+
+Classes that directly require App Engine's ndb library. Provided
+as a separate module in case of failure to import ndb while
+other App Engine libraries are present.
+"""
+
+import logging
+
+from google.appengine.ext import ndb
+
+from oauth2client import client
+
+
+NDB_KEY = ndb.Key
+"""Key constant used by :mod:`oauth2client.contrib.appengine`."""
+
+NDB_MODEL = ndb.Model
+"""Model constant used by :mod:`oauth2client.contrib.appengine`."""
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class SiteXsrfSecretKeyNDB(ndb.Model):
+    """NDB Model for storage for the sites XSRF secret key.
+
+    Since this model uses the same kind as SiteXsrfSecretKey, it can be
+    used interchangeably. This simply provides an NDB model for interacting
+    with the same data the DB model interacts with.
+
+    There should only be one instance stored of this model, the one used
+    for the site.
+    """
+    secret = ndb.StringProperty()
+
+    @classmethod
+    def _get_kind(cls):
+        """Return the kind name for this class."""
+        return 'SiteXsrfSecretKey'
+
+
+class FlowNDBProperty(ndb.PickleProperty):
+    """App Engine NDB datastore Property for Flow.
+
+    Serves the same purpose as the DB FlowProperty, but for NDB models.
+    Since PickleProperty inherits from BlobProperty, the underlying
+    representation of the data in the datastore will be the same as in the
+    DB case.
+
+    Utility property that allows easy storage and retrieval of an
+    oauth2client.Flow
+    """
+
+    def _validate(self, value):
+        """Validates a value as a proper Flow object.
+
+        Args:
+            value: A value to be set on the property.
+
+        Raises:
+            TypeError if the value is not an instance of Flow.
+        """
+        _LOGGER.info('validate: Got type %s', type(value))
+        if value is not None and not isinstance(value, client.Flow):
+            raise TypeError('Property %s must be convertible to a flow '
+                            'instance; received: %s.' % (self._name,
+                                                         value))
+
+
+class CredentialsNDBProperty(ndb.BlobProperty):
+    """App Engine NDB datastore Property for Credentials.
+
+    Serves the same purpose as the DB CredentialsProperty, but for NDB
+    models. Since CredentialsProperty stores data as a blob and this
+    inherits from BlobProperty, the data in the datastore will be the same
+    as in the DB case.
+
+    Utility property that allows easy storage and retrieval of Credentials
+    and subclasses.
+    """
+
+    def _validate(self, value):
+        """Validates a value as a proper credentials object.
+
+        Args:
+            value: A value to be set on the property.
+
+        Raises:
+            TypeError if the value is not an instance of Credentials.
+        """
+        _LOGGER.info('validate: Got type %s', type(value))
+        if value is not None and not isinstance(value, client.Credentials):
+            raise TypeError('Property %s must be convertible to a '
+                            'credentials instance; received: %s.' %
+                            (self._name, value))
+
+    def _to_base_type(self, value):
+        """Converts our validated value to a JSON serialized string.
+
+        Args:
+            value: A value to be set in the datastore.
+
+        Returns:
+            A JSON serialized version of the credential, else '' if value
+            is None.
+        """
+        if value is None:
+            return ''
+        else:
+            return value.to_json()
+
+    def _from_base_type(self, value):
+        """Converts our stored JSON string back to the desired type.
+
+        Args:
+            value: A value from the datastore to be converted to the
+                   desired type.
+
+        Returns:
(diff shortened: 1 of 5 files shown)


oauth2client/contrib/_appengine_ndb.py
```diff
@@ -0,0 +1,167 @@
...
+Classes that directly require App Engine's ndb library. Provided
+as a separate module in case of failure to import ndb while
+other App Engine libraries are present.
+"""
+
+import logging
+
+from google.appengine.ext import ndb
+
+from oauth2client.client import Credentials
```
nathanielmanistaatgoogle: Since this is a new file, could I ask for adherence to the style guide's direction to import only modules and not classes, functions, or other elements of modules?

oauth2client/contrib/_appengine_ndb.py
```diff
@@ -0,0 +1,167 @@
...
+
+        Args:
+            value: A value to be set on the property.
+
+        Raises:
+            TypeError if the value is not an instance of Flow.
+        """
+        _LOGGER.info('validate: Got type %s', type(value))
+        if value is not None and not isinstance(value, Flow):
+            raise TypeError('Property %s must be convertible to a flow '
```
nathanielmanistaatgoogle, line 79: Implicit string concatenation is one of Python's worst features and one should always make an effort to avoid it. If you line-wrap right after `(` and then again after `(` can you fit the format string in one piece?

oauth2client/contrib/_appengine_ndb.py
dhermes: :+1: Works for me.

oauth2client/contrib/_appengine_ndb.py
```diff
@@ -0,0 +1,167 @@
...
+                            'instance; received: %s.' % (self._name,
+                                                         value))
+
+
+# TODO(dhermes): Turn this into a JsonProperty and overhaul the Credentials
```
nathanielmanistaatgoogle: Any desire to file this work item as an issue, so that this comment is `TODO(issue NNN):`? Wasn't it you who told me a few weeks ago that you don't care for TODOs in code?

oauth2client/contrib/appengine.py
```diff
@@ -49,9 +49,19 @@
 # TODO(dhermes): Resolve import issue.
```
nathanielmanistaatgoogle: Is this `TODO` being addressed by this very pull request? Or does this `TODO` refer to some other work item?

oauth2client/contrib/_appengine_ndb.py
dhermes, line 79: Too long to do it without concat. Tops out at 87 columns

``` python
        if value is not None and not isinstance(value, Flow):
            raise TypeError(
                'Property %s must be convertible to a flow instance; received: %s.' % (
                    self._name, value))
```

RE: "one of Python's worst features" That's just like, [your opinion man](https://www.youtube.com/watch?v=pWdd6_ZxX8c).

oauth2client/contrib/appengine.py
dhermes: This `TODO` is from 2012.

oauth2client/contrib/_appengine_ndb.py
dhermes: Yup. This `TODO` is from 2012 when I implemented the `ndb` support.

oauth2client/contrib/appengine.py
```diff
@@ -49,9 +49,19 @@
...
 # This is a temporary fix for a Google internal issue.
 try:
-    from google.appengine.ext import ndb
```
nathanielmanistaatgoogle, line 52: Feel free to push back and call me fussy, but I'd rather see this as

```
try:
    from oauth2client.contrib import _appengine_ndb
except ImportError:
    _appengine_ndb = None

if _appengine_ndb is None:
    <assignment-to-None statements>
else:
    <assignment-to-not-None statements>
```

. Just yesterday or the day before I was directed to the formal documentation: ["Imported names should always be considered an implementation detail"](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces) (i.e. not part of the public API of the module in which they are imported).

oauth2client/contrib/appengine.py
dhermes: Also, to answer your question, no it isn't being addressed. This PR just reduces the amount of conditionally defined code in `oauth2client` and gets 100% code coverage the the `ndb` stuff in the process.

I have no idea why the import failed (`This is a temporary fix for a Google internal issue.`) and don't have a way to fix it now.

oauth2client/contrib/appengine.py
dhermes, line 52: Which imported names are you referring to here?

oauth2client/contrib/_appengine_ndb.py
nathanielmanistaatgoogle, line 79: Thanks for trying... and doesn't PEP8 call for eight-space continuation indent? Wouldn't that push it out to 91?

oauth2client/contrib/appengine.py
nathanielmanistaatgoogle, line 52: As currently drafted (so on the right side of the diff; sorry that this conversation is happening on the left): aren't the names `CredentialsNDBProperty`, `FlowNDBProperty`, and `SiteXsrfSecretKeyNDB` intended for application use?

oauth2client/contrib/_appengine_ndb.py
dhermes, line 79: AFAIK 8 spaces is only for cases where its ambiguous.

oauth2client/contrib/_appengine_ndb.py
dhermes: Also, yes I'll create the issue.

oauth2client/contrib/appengine.py
dhermes, line 52: Ahhh got it. I'd imagine the comment above is for "someone else's library" but here we are in full control. You can find examples of the BDFL doing this in real life ([e.g.](https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/ecc346e86b2b3790ef82c2454d03a789de24030f/ndb/__init__.py#L22) in `ndb`).

dhermes, line 52: Anyhow, under the Google style-guide rule of "only import modules", there is only one way to go here. Will send update.

oauth2client/contrib/_appengine_ndb.py
dhermes: Created #439

oauth2client/contrib/appengine.py
dhermes: Filed #440.

dhermes:
@nathanielmanistaatgoogle PTAL, I addressed the comments.

oauth2client/contrib/_appengine_ndb.py
```diff
@@ -0,0 +1,163 @@
...
+Classes that directly require App Engine's ndb library. Provided
+as a separate module in case of failure to import ndb while
+other App Engine libraries are present.
+"""
+
+import logging
+
+from google.appengine.ext import ndb
+
+from oauth2client import client as client_mod
```
nathanielmanistaatgoogle: Drop this rename - the name doesn't conflict with any other name used in the text of this module?

nathanielmanistaatgoogle:
You addressed it just fine; a counterargument is a perfectly valid form of addressing a comment. :-)

Looks great otherwise (if you really want to keep the `as client_mod`, well it's in an internal place so that's fine). Squash, retest, and merge. :-)

oauth2client/contrib/_appengine_ndb.py
dhermes: I did it because `client` can be a module name or an object name (I was being defensive). I've been bit before in unit-tests where I import `client` and then create `client = client.Client()` (OOPS). No conflict here so I'll change it.

dhermes:
Did the rename. Will squash and wait until the sun goes down for Travis

oauth2client/contrib/appengine.py
```diff
@@ -62,6 +61,21 @@
...
+    FlowNDBProperty = None
+    _NDB_KEY = None
+    _NDB_MODEL = None
+    SiteXsrfSecretKeyNDB = None
+else:
+    CredentialsNDBModel = _appengine_ndb.CredentialsNDBModel
+    CredentialsNDBProperty = _appengine_ndb.CredentialsNDBProperty
+    FlowNDBProperty = _appengine_ndb.FlowNDBProperty
+    _NDB_KEY = _appengine_ndb.NDB_KEY
+    _NDB_MODEL = _appengine_ndb.NDB_MODEL
```
dhermes, line 76: FYI @nathanielmanistaatgoogle I just had to fix an issue here. These used to refer to non-existent values

``` python
_NDB_KEY = _appengine_ndb._NDB_KEY
_NDB_MODEL = _appengine_ndb._NDB_MODEL
```

which caused [build 1000](https://travis-ci.org/google/oauth2client/builds/111547666) to fail. (Neat that it's build 1000.)

I feel silly for not running locally after the tiny review change, since it was hours to wait on Travis.

Le sigh.

nathanielmanistaatgoogle, line 76: I should have caught that in review; thanks for letting me know.

If we were lint-compliant we would have gotten a "suspicious access of private element" warning, wouldn't we have? In a perfect world. :-P

dhermes, line 76: First #212, then `tox -e lint` will follow

PR #438, opened 2016-02-24
fix an error in the sample code

jeremi:
credential is under the oauth object in the request


nathanielmanistaatgoogle:
@waprin on this also as the original author of this code. (If I'm reading history correctly, right?)

jonparrott:
This LGTM, but I will wait for @waprin to chime in.

waprin:
Lgtm, ty

PR #452, opened 2016-03-04
Fix flask required decorator to redirect on expired credentials.

theacodes:


jonparrott:
/cc @dhermes or @nathanielmanistaatgoogle 

@elibixby ran into this in his application. Apparently I was mistaken in just checking `credentials.invalid` which doesn't actually get set to `True` until a request fails.

@waprin a similar change is probably need to the django helper.

waprin:
acknowledged, I'll take a look at Django.

tests/contrib/test_flask_util.py
```diff
@@ -384,6 +404,35 @@ def index():
...
+            utcnow.return_value = datetime.datetime(1990, 5, 29)
+
+            with self.app.test_client() as client:
+                with client.session_transaction() as session:
+                    session['google_oauth2_credentials'] =\
+                        credentials.to_json()
+
+                response = client.get('/protected')
+                self.assertEqual(response.status_code, httplib.OK)
+                self.assertTrue('Hello' in response.data.decode('utf-8'))
```
nathanielmanistaatgoogle: `self.assertIn`?

tests/contrib/test_flask_util.py
```diff
@@ -384,6 +404,35 @@ def index():
...
             self.assertTrue('Hello' in response.data.decode('utf-8'))
 
+        # Expired credentials with refresh token, should allow.
+        credentials.token_expiry = datetime.datetime(1990, 5, 28)
+        with mock.patch('oauth2client.client._UTCNOW') as utcnow:
+            utcnow.return_value = datetime.datetime(1990, 5, 29)
+
+            with self.app.test_client() as client:
+                with client.session_transaction() as session:
+                    session['google_oauth2_credentials'] =\
```
nathanielmanistaatgoogle: Use parentheses rather than backslash for continuation in cases like this.

["Do not use backslash line continuation."](https://google.github.io/styleguide/pyguide.html?showone=Line_length#Line_length)

tests/contrib/test_flask_util.py
```diff
@@ -350,8 +351,27 @@ def test_with_credentials(self):
                 credentials.refresh_token)
             self.assertEqual(self.oauth2.user_id, '123')
             self.assertEqual(self.oauth2.email, 'user@example.com')
+
```
nathanielmanistaatgoogle: Stray keystroke inadvertently creates new blank line in file?

oauth2client/contrib/flask_util.py
```diff
@@ -443,7 +443,12 @@ def credentials(self):
 
     def has_credentials(self):
         """Returns True if there are valid credentials for the current user."""
-        return self.credentials and not self.credentials.invalid
```
dhermes, line 446: Sidebar: If not for this use case, what is `invalid` useful for?

tests/contrib/test_flask_util.py
dhermes: Yeah, backslash continuation is double gross

oauth2client/contrib/flask_util.py
elibixby, line 446: Agreed, also surprisingly, invalid is not even set before a failed request even if there is no refresh token present. At the very least, the client can detect if the access token is expired, and there isn't a refresh token, and set invalid preemptively.

theacodes, line 446: AFAIT, `invalid` is more misleading than useful.

jonparrott:
Thanks for the review, just FYI I won't be able to fix the few nits for a few days. Will ping when ready.

oauth2client/contrib/flask_util.py
dhermes, line 446: @nathanielmanistaatgoogle Is there any way to do a survey of how and if people are using features like this?

nathanielmanistaatgoogle, line 446: I suspect it's basically impossible.

In case you needed more understanding of why I try so hard to minimize the public interface of a library released into the wild...

tests/contrib/test_flask_util.py
theacodes: Fixed.

tests/contrib/test_flask_util.py
theacodes: _Shame cube_

Fixed.

tests/contrib/test_flask_util.py
theacodes: Fixed.

jonparrott:
Comments addressed.

dhermes:
LoL Python 2.6. You need to swap out for `unittest2` in `tests.contrib.test_flask_util`.

jonparrott:
> LoL Python 2.6

ugh. Fixed.

tests/contrib/test_flask_util.py
```diff
@@ -384,6 +403,35 @@ def index():
...
+            utcnow.return_value = datetime.datetime(1990, 5, 29)
+
+            with self.app.test_client() as client:
+                with client.session_transaction() as session:
+                    session['google_oauth2_credentials'] =\
+                        credentials.to_json()
+
+                response = client.get('/protected')
+            self.assertEqual(response.status_code, httplib.FOUND)
+            self.assertTrue('oauth2authorize' in response.headers['Location'])
```
nathanielmanistaatgoogle: `self.assertIn`.

tests/contrib/test_flask_util.py
```diff
@@ -384,6 +403,35 @@ def index():
...
+            self.assertTrue('protected' in response.headers['Location'])
```
nathanielmanistaatgoogle: ... and here.

jonparrott:
@nathanielmanistaatgoogle all of the instances `self.assertTrue(y in z)` have been changed to `self.assertIn(y, z)`.

tests/contrib/test_flask_util.py
nathanielmanistaatgoogle: Fixed?

theacodes: Somehow got one instance of this but not the other. _Fixed_ fixed.

oauth2client/contrib/flask_util.py
```diff
@@ -443,7 +443,12 @@ def credentials(self):
...
+        if not self.credentials:
+            return False
```
nathanielmanistaatgoogle, line 447: How bad does this look as one boolean expression? Is it just

```
return (
    self.credentials and
    (not self.credentials.access_token_expired or self.credentials.refresh_token))
```

?

If that's too unattractive, please change

```
if <condition>:
  return False
if <other condition>:
  return False
return True
```

to

```
if <condition>:
  return False
elif <other condition>:
  return False
else:
  return True
```

.

theacodes, line 447: Definitely clearer as two statements. I added a comment as well.

oauth2client/contrib/flask_util.py
```diff
@@ -443,7 +443,14 @@ def credentials(self):
...
+        # Is the access token expired? If so, do we have an refresh token?
+        if (self.credentials.access_token_expired
```
nathanielmanistaatgoogle: `elif` on this line.

theacodes: done

nathanielmanistaatgoogle:
Looks good; waiting on fresh test results.

jonparrott:
Thanks. :)

PR #458, opened 2016-03-15
compatibility validate_token with python3

gblcarvalho:
In python3, the encoding is required when create a bytearray from a STRING


gblmiranda:
I signed it!

PR #467, opened 2016-03-17
Added tests for contrib.locked_file.

theacodes:
Towards #212 

This module beastly. It's only user is `multistore_file` and honestly I would be in favor of outright removing this and that from the library altogether.

These tests do not cover the platform-specific opener classes - I'd really like to hear some ideas on how to test those.

### DIFF
--- a/tests/contrib/test_locked_file.py
+++ b/tests/contrib/test_locked_file.py
@@ -1,0 +1,243 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import errno
+import os
+import tempfile
+
+import mock
+from six.moves import reload_module
+import unittest2
+
+from oauth2client.contrib import locked_file
+
+
+class TestOpener(unittest2.TestCase):
+    def _make_one(self):
+        _filehandle, filename = tempfile.mkstemp()
+        os.close(_filehandle)
+        return locked_file._Opener(filename, 'r+', 'r'), filename
+
+    def test_ctor(self):
+        instance, filename = self._make_one()
+        self.assertFalse(instance._locked)
+        self.assertEqual(instance._filename, filename)
+        self.assertEqual(instance._mode, 'r+')
+        self.assertEqual(instance._fallback_mode, 'r')
+        self.assertIsNone(instance._fh)
+        self.assertIsNone(instance._lock_fd)
+
+    def test_is_locked(self):
+        instance, _ = self._make_one()
+        self.assertFalse(instance.is_locked())
+        instance._locked = True
+        self.assertTrue(instance.is_locked())
+
+    def test_file_handle(self):
+        instance, _ = self._make_one()
+        self.assertIsNone(instance.file_handle())
+        fh = mock.Mock()
+        instance._fh = fh
+        self.assertEqual(instance.file_handle(), fh)
+
+    def test_filename(self):
+        instance, filename = self._make_one()
+        self.assertEqual(instance.filename(), filename)
+
+    def test_open_and_lock(self):
+        instance, _ = self._make_one()
+        instance.open_and_lock(1, 1)
+
+    def test_unlock_and_close(self):
+        instance, _ = self._make_one()
+        instance.unlock_and_close()
+
+
+class TestPosixOpener(TestOpener):
+    def _make_one(self):
+        _filehandle, filename = tempfile.mkstemp()
+        os.close(_filehandle)
+        return locked_file._PosixOpener(filename, 'r+', 'r'), filename
+
+    def test_relock_fail(self):
+        instance, _ = self._make_one()
+        instance.open_and_lock(1, 1)
+
+        self.assertTrue(instance.is_locked())
+        self.assertIsNotNone(instance.file_handle())
+        self.assertRaises(
+            locked_file.AlreadyLockedException, instance.open_and_lock, 1, 1)
+
+    @mock.patch('oauth2client.contrib.locked_file.open', create=True)
+    def test_lock_access_error_fallback_mode(self, mock_open):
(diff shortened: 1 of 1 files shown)


jonparrott:
@thobrla it seems that gsutil is a user of `multistore_file` [here](https://github.com/GoogleCloudPlatform/gsutil/blob/4bf87817f6ed95633f157a1c6071c9e0cf982718/gslib/gcs_json_api.py#L234). How would you feel about us trying to remove this? It's a lot of code (and cruft) for a very specific use case. Is there a better storage option we can provide (sqlite?)

thobrla:
I'm fine with removing both locked_file/multistore_file if a suitable alternative is provided.  The requirement is to safely use a single credential across multiple processes and threads.  If we can find a better cross-platform way to do that, I'm happy to port gsutil over (other incompatibilities notwithstanding).

jonparrott:
I'll start a separate issue for that, thanks, @thobrla.

tests/contrib/test_locked_file.py
```diff
@@ -0,0 +1,243 @@
...
+        self.assertIsNone(instance._fh)
+        self.assertIsNone(instance._lock_fd)
+
+    def test_is_locked(self):
+        instance, _ = self._make_one()
+        self.assertFalse(instance.is_locked())
+        instance._locked = True
+        self.assertTrue(instance.is_locked())
+
+    def test_is_locked(self):
```
dhermes: Name duplicate

tests/contrib/test_locked_file.py
```diff
@@ -0,0 +1,243 @@
...
+        instance._fh = fh
+        self.assertEqual(instance.file_handle(), fh)
+
+    def test_filename(self):
+        instance, filename = self._make_one()
+        self.assertEqual(instance.filename(), filename)
+
+    def test_open_and_lock(self):
+        instance, _ = self._make_one()
+        instance.open_and_lock(1, 1)
```
dhermes, line 60: No expected output? Any mock that you could `assert_called_with()` on?

tests/contrib/test_locked_file.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def test_unlock_and_close(self):
+        instance, _ = self._make_one()
+        instance.unlock_and_close()
```
dhermes, line 64: Ditto here

tests/contrib/test_locked_file.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+class TestPosixOpener(TestOpener):
```
dhermes, line 67: Will this test run on non-posix OSes? Do we need a `unittest.skipIf`?

tests/contrib/test_locked_file.py
```diff
@@ -0,0 +1,243 @@
...
+        self.assertTrue(instance.is_locked())
+        self.assertIsNotNone(instance.file_handle())
+        self.assertRaises(
+            locked_file.AlreadyLockedException, instance.open_and_lock, 1, 1)
+
+    @mock.patch('oauth2client.contrib.locked_file.open', create=True)
+    def test_lock_access_error_fallback_mode(self, mock_open):
+        # NOTE: This is a bad case. The behavior here should be that the
+        # error gets re-raised, but the module lets the if statement fall
+        # trhough.
```
dhermes: through

tests/contrib/test_locked_file.py
```diff
@@ -0,0 +1,243 @@
...
+        with mock.patch('os.open') as mock_os_open:
+            mock_os_open.side_effect = [OSError(errno.EPERM, '')]
+            self.assertRaises(OSError, instance.open_and_lock, 1, 1)
+
+    @mock.patch('oauth2client.contrib.locked_file.open', create=True)
+    @mock.patch('oauth2client.contrib.locked_file.logger')
+    @mock.patch('time.time')
+    def test_lock_timeout_error(self, mock_time, mock_logger, mock_open):
+        instance, _ = self._make_one()
+        # Make it seem like 10 seconds have passed between calls.
```
dhermes, line 118: :+1:

tests/contrib/test_locked_file.py
```diff
@@ -0,0 +1,243 @@
...
+            self.assertEqual(instance.file_handle(), fh_mock)
+
+    @mock.patch('oauth2client.contrib.locked_file.open', create=True)
+    @mock.patch('time.time')
+    @mock.patch('time.sleep')
+    def test_lock_retry_success(self, mock_sleep, mock_time, mock_open):
+        instance, _ = self._make_one()
+        # Make it seem like 0.5 second have passed between calls. Extra values
+        # are needed by the logging module.
+        mock_time.side_effect = [0, 1]
```
dhermes, line 154: 0.5s or 1s?

dhermes:
LGTM mostly

tests/contrib/test_locked_file.py
theacodes: Fixed.

tests/contrib/test_locked_file.py
theacodes, line 60: It's an abstract method, really. It should just raise `NotImplementedError`. But I figured it was safer to go ahead and test it before we made that change.

tests/contrib/test_locked_file.py
theacodes, line 67: The posix opener is defined regardless of OS-level dependencies. As far as I can tell, it doesn't actually depend on anything posix-related. The flags it passes to open are available across all platforms. It just does 'posix-like' locking by creating a `.lock` file. 

All the more reason to remove this nonsense.

tests/contrib/test_locked_file.py
theacodes: fixed.

tests/contrib/test_locked_file.py
theacodes, line 154: Fixed.

jonparrott:
Updated, should address most of your comments. 

Thanks for the review. I have no idea what it is about this library but I just end up writing sloppy tests. :|

tests/contrib/test_locked_file.py
dhermes, line 60: Gotcha. I see they just `pass`.

dhermes:
LGTM more mostly

jonparrott:
Any qualms with merging this? Should we get @nathanielmanistaatgoogle? I'm just past caring on this one and would rather push for its removal if we can get gcloud cli and gsutil to play along.

dhermes:
Ha sorry my last comment was meant to be a "go-ahead and merge" LGTM

jonparrott:
Ha. Fair enough. I'm still open to suggestions on testing the platform-specific stuff (mocks?) provided we don't vote it off the island.

PR #468, opened 2016-03-17
Move tests/ dir to oauth2client

### LINKED ISSUE #454: tests/ directory not included in the sdist tarball
Downstream may want to run the unittest when building a package. But the tests/dir is no longer included in the sdist tarball (but it was at least in version 1.5.1) . This was done with https://codereview.appspot.com/6503047/ .

toabctl:
That way, the tests are included when a sdist tarball is created
and downstream can run the unittests in their environment.

Fixes #454


dhermes:
@toabctl This might be easier to review if you include the `bash` / `git` commands you used to move the files

jonparrott:
@toabctl friendly ping.

pferate:
@toabctl There is one more file that was missed, `docs/conf.py`, line 68.  That's why the `docs` test failed.

```diff
diff --git a/docs/conf.py b/docs/conf.py
index f84911c..6398e78 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -65,7 +65,7 @@ exclude_patterns = ['_build']
 # settings module and load it. This assumes django has been installed
 # (but it must be for the docs to build), so if it has not already
 # been installed run `pip install -r docs/requirements.txt`.
-os.environ['DJANGO_SETTINGS_MODULE'] = 'tests.contrib.test_django_settings'
+os.environ['DJANGO_SETTINGS_MODULE'] = 'oauth2client.tests.contrib.test_django_settings'
 import django
 if django.VERSION[1] < 7:
     sys.path.insert(0, '.')
```

@dhermes @jonparrott I started my own branch to replicate/verify this PR and came up with a similar diff using `git mv tests/ oauth2client/tests/`, then updating the files as done here.

Seeing as this PR is out of date now, should I submit my own PR to resolve issue #454?  Unless @toabctl wants to update this PR instead.

nathanielmanistaatgoogle:
@pferate: no preference whether you do this in a separate pull request or whether @toabctl continues with this pull request. Including the tests in the distributed code is always going to be at least a little weird to me, so it's not actually a priority for me that it even happen at all.

PR #473, opened 2016-03-24
Populate token expiry for GCE credentials

bendemaree:
Populates the `token_expiry` property for GCE App Assertion credentials. The token responses from the metadata service have a `expires_in` value in the response that can be leveraged to make things work better on the base `OAuth2Credentials` class, like `access_token_expired`.

I largely followed [the implementation in the base client](https://github.com/google/oauth2client/blob/master/oauth2client/client.py#L904) though less defensively; I don't see any reason why `expires_in` wouldn't be present though perhaps someone knows better. I did note that the existing tests actually assert this value is not set in the mocked token response so perhaps that provides evidence against my assumption.


oauth2client/contrib/gce.py
```diff
@@ -27,6 +28,7 @@
 
 from oauth2client._helpers import _from_bytes
 from oauth2client import util
+from oauth2client.client import _UTCNOW
```
dhermes: Just define a local `_UTCNOW` if you need to stub it out for testing. This is unnecessary coupling of these modules.

bendemaree: Updated; can squash prior to merging. :smile:

tests/contrib/test_gce.py
```diff
@@ -73,6 +78,8 @@ def _refresh_success_helper(self, bytes_response=False):
         self.assertEquals(None, credentials.access_token)
         credentials.refresh(http)
         self.assertEquals(access_token, credentials.access_token)
+        self.assertFalse(credentials.access_token_expired)
+        self.assertTrue(credentials.token_expiry > datetime.utcnow())
```
dhermes, line 82: FYI you could do something like this to control the output of `utcnow`:

``` python
@mock.patch('datetime.datetime.utcnow', return_value=123456789)
def _refresh_success_helper(self, utcnow, bytes_response=False):
    # Make sure the value returned was 12345678
    ...
    utcnow.assert_called_once_with()
```

bendemaree, line 82: Yep, certainly...this is an older habit for me to prevent myself from accidentally returning a mocked value that is not the right type: `utcnow` actually returns a `datetime.datetime` but I had to check the stdlib docs because it seemed possible it would return a Unix time `int`. Happy to update if you'd prefer the mock, though!

oauth2client/contrib/gce.py
```diff
@@ -135,6 +136,8 @@ def _refresh(self, http_request):
                 raise HttpAccessTokenRefreshError(str(e),
                                                   status=response.status)
             self.access_token = token_content['access_token']
+            delta = datetime.timedelta(seconds=int(token_content['expires_in']))
```
elibixby, line 139: This is unnecessary. The metadata server returns an "expires_at" field along with the access token.

elibixby, line 139: See https://github.com/google/oauth2client/compare/master...elibixby:master#diff-f6cbfd778562492d81ffdbcfb39ee128R142

elibixby:
Hi bendemaree@ thanks for the contribution, but I'm actually working on a larger update to GCE credentials that includes this. 

Still writing tests but I'll go ahead and submit a pull request so more work doesn't get duplicated =(

dhermes:
@elibixby smaller PRs are much easier to review FWIW.

elibixby:
See #476 

I rewrote the wrapper function for the metadata server so that it can be used by lots of common calls that will need to be made, so I would like the first PR to contain that. However, if after that we want to rewrite the interface in a bunch of small PRs that's fine with me. Still, I think all these changes are quite related and should be reviewed together. Just my .02

(Also FYI, I AM splitting up the PR with the GCE credential updates, and the PR to add IAM blob signing, as IAM blob signing could potentially be used in other credential types)

oauth2client/contrib/gce.py
bendemaree, line 139: Oh. Is the 0.1 metadata service more recent than v1? Or a different API? I don't have it in front of me but IIRC the v1 server only returned the seconds until expiration.

Edit: Just noticed the commit comment. Still curious, though.

dhermes, line 139: @Galabar001 might be able to answer your question

bendemaree:
Ah, alright, thanks for the review anyway! Looking forward to the update; we're trying out some optimistic credential refresh logic since the OAuth dance takes a bit and could cause a request to hang while the token is implicitly refreshed.

Closing out; feel free to reopen if needed.

bendemaree:
@elibixby Given that #476 was closed out and there aren't followup PRs yet (I may have missed them though), would it be possible to put this back on the table for review?...

elibixby:
I'm going to split #476 into smaller PRs and submit those. It's on my plate for the next couple weeks. 

I think the right thing to do here is write a `metadata` module that wraps the metadata server and caches requests. Which should make follow up PRs easy. If you want to take that on you are welcome to. I can't promise I will get to it this week.

bendemaree:
@elibixby @jonparrott Sorry to bring this back to your attention again, but is there any chance we can merge this in? I don't think it's more than a minor bugfix, and at this point it's been 2 months without the major refactor needed to get this behavioral fix. 😢 I don't have the familiarity or bandwidth to re-wrap the metadata server myself, unfortunately.

jonparrott:
@dhermes are you fine with this being merged?

oauth2client/contrib/gce.py
theacodes, line 139: So which do we prefer? `expires_in` or `expires_at`?

elibixby:
@bendemaree Sorry about taking so long. I will have some time in the next week to work on my PRs, but if you'd like to go ahead and submit this for merge. That's understandable.

dhermes:
https://coveralls.io/builds/5540724

What happened to 100% coverage?

dhermes:
Ahhh I know. This needs to be rebased against `HEAD` in `master`

bendemaree:
@dhermes Squashed and rebased (eventually!).

@elibixby @jonparrott Thanks a lot for circling back on this; much obliged!

oauth2client/contrib/gce.py
```diff
@@ -135,6 +136,8 @@ def _refresh(self, http_request):
...
+            self.token_expiry = delta + datetime.datetime.utcnow()
```
ericbuehl, line 140: should this use oauth2client.client._UTCNOW for consistency?

theacodes, line 140: Please

bendemaree, line 140: That suggestion conflicts with [prior review](https://github.com/google/oauth2client/pull/473#discussion_r57377582).

theacodes, line 140: Gotcha. No worries then. 👍

bendemaree:
Bump. 😁

oauth2client/contrib/gce.py
theacodes, line 139: Historical note: it seems we use `expires_in` in this library consistently, so no need to rock the boat.

jonparrott:
This LGTM.

jonparrott:
@dhermes if you have any post-merge objections, let me know and I'll fix myself.

bendemaree:
@jonparrott Is there an upcoming release planned that will include this?

jonparrott:
Yes, it will be 3.0.0. I'm waiting for a few outstanding PRs to be resolved first.

PR #475, opened 2016-03-25
Add **kwargs to allow "sub" claim in JWT payload for service account type

luisTJ:
By adding optional parameter to following functions:
_from_parsed_json_keyfile
from_json_keyfile_name
from_json_keyfile_dict

_from_p12_keyfile_contents
from_p12_keyfile
from_p12_keyfile_buffer

and pass them to the constructor will allow caller to add "sub" claim field to JWT Payload

### DIFF
--- a/oauth2client/service_account.py
+++ b/oauth2client/service_account.py
@@ -137,7 +137,7 @@ class ServiceAccountCredentials(AssertionCredentials):
             strip, to_serialize=to_serialize)
 
     @classmethod
-    def _from_parsed_json_keyfile(cls, keyfile_dict, scopes):
+    def _from_parsed_json_keyfile(cls, keyfile_dict, scopes, **kwargs):
         """Helper for factory constructors from JSON keyfile.
 
         Args:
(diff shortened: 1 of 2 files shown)


jonparrott:
We already have [create_delegated](http://oauth2client.readthedocs.org/en/latest/source/oauth2client.service_account.html?highlight=create_delegated#oauth2client.service_account.ServiceAccountCredentials.create_delegated) for this. Is this change still necessary?

jonparrott:
& If the change is still necessary, I'd rather it be a `claims` keyword argument rather than just rolling up all of the kwargs.

dhermes:
Ditto. I'm AFK now but seem to remember commenting on one of the `kwargs` bugs (post-2.0) that we should remove it in total and try to get a handle on the claims people will use.

dhermes:
Maybe just an `add_claim` method? Return new instance or mutate current one?

luisTJ:
Calling `create_delegated` does it for me. Thanks!
But having it all done in one shot would be nice in my opinion.

dhermes:
:+1:

dhermes:
@luisTJ Other than `sub`, are there other claims you use?

Also, are we OK to close this PR?

luisTJ:
I only need `sub` for now. And we can close this PR. Thanks!

Issue #471, opened 2016-03-18
Allow GCE credentials to sign blobs


elibixby:
The IAM API which is now in Beta has a [signBlob](https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts/signBlob) method which signs a blob on behalf of a service account for an authorized users. This should allow GCE credentials to act like fully fledged service accounts by signing blobs with this API call after acquiring an access token from the metadata server. This would only work on GCE instances that have the cloud-platform or iam scopes, but it would still be quite nice as it would allow application_default_credentials to almost always have the capacity to sign blobs.

@jonparrott

dhermes:
Woo hoo!

elibixby:
Started work on this https://github.com/google/oauth2client/compare/master...elibixby:master 
Need lots of updates to gce.AppAssertionCredentials first. Details in commit messages.

elibixby:
@dhermes @jonparrott 
So after lots of discussion with an IAM engineer, it looks like there are no plans for Service Accounts to --by default -- have permission to sign blobs on behalf of themselves. 


So there are several options:
1. Drop this entirely, and say users don't need it.

2. Require users add the necessary role `roles/iam.serviceAccountActor` for their service account to their service account in order to sign blobs

3. Require users manually pass in an `IAMSigner` that is authenticated with credentials which have `roles/iam.serviceAccountActor` role on the `gce.AppAssertionCredentials` service account and pass the `IAMSigner` to the `AppAssertionCredentials`

4. Some hybrid of  (2) and (3)

I think (4) is best but I'm not sure how to implement it. To see what I previously had in mind (before I was aware of this awkwardness) check out https://github.com/elibixby/oauth2client/pull/1

JacobHayes:
@elibixby For what it's worth, I'd much prefer this be supported (though understand time/effort is limited). If a choice between 2 vs 3, 3 (as far as I understand) gives developers a bit more testing capability while developing applications (ie: outside GCE), though I think a hybrid would be best as well. We use GCS to store documents (different buckets for each environment), some of which require signed urls to access. Without access to a blob signing API, we are forced to store private keys on developer boxes in order to test out that functionality in our application, which is much less than ideal (and complicates our code, requiring us to implement this ourselves).

jonparrott:
GCS on GCE/GKE is really the key use case here. I'm okay with 3/4 being the approach. Explicit is better than implicit.

elibixby:
[...]

PR #476, opened 2016-03-25
Update GCE AppAssertionCredentials

elibixby:
This provides several much needed updates to GCE App Assertion Credentials
1. Properly populates token expiry
2. Retrieves credentials scopes from the metadata server
3. Allows credentials to use [custom service accounts](https://cloud.google.com/compute/docs/authentication#createcutomserviceaccount) by providing an optional email field
4. Implements serialization/deserialization for this email field
5. Provides a project_id property using metadata server in preparation for a pull request addressing #471

### DIFF
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -183,7 +183,7 @@ class AppAssertionCredentials(AssertionCredentials):
     @property
     def project_id(self):
         if not self._project_id:
-            self._project_id = _get_metadata('project', 'project-id')
+            self._project_id = _get_metadata('project-id')
         return self._project_id
 
     def sign_blob(self, blob):
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -23,7 +23,6 @@ import warnings
 
 import httplib2
 from six.moves import http_client
-from six.moves import urllib
 
 from oauth2client._helpers import _from_bytes
 from oauth2client import util
@@ -36,10 +35,7 @@ __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 logger = logging.getLogger(__name__)
 
 # URI Template for the endpoint that returns access_tokens.
-_METADATA_ROOT = ('http://metadata.google.internal/computeMetadata/v1/'
-                  'instance/service-accounts/default/')
-META = _METADATA_ROOT + 'token'
-_DEFAULT_EMAIL_METADATA = _METADATA_ROOT + 'email'
+_METADATA_ROOT = 'http://metadata.google.internal/0.1/meta-data'
 _SCOPES_WARNING = """\
 You have requested explicit scopes to be used with a GCE service account.
 Using this argument will have no effect on the actual scopes for tokens
@@ -48,28 +44,33 @@ can't be overridden in the request.
 """
 
 
-def _get_service_account_email(http_request=None):
-    """Get the GCE service account email from the current environment.
-
-    Args:
-        http_request: callable, (Optional) a callable that matches the method
-                      signature of httplib2.Http.request, used to make
-                      the request to the metadata service.
-
-    Returns:
-        tuple, A pair where the first entry is an optional response (from a
-        failed request) and the second is service account email found (as
-        a string).
-    """
-    if http_request is None:
+def _get_metadata(http_request=None, *path):
+    if not http_request:
         http_request = httplib2.Http().request
-    response, content = http_request(
-        _DEFAULT_EMAIL_METADATA, headers={'Metadata-Flavor': 'Google'})
+    full_path = "/".join(path.insert(0, _METADATA_ROOT))
+    response, content = http_client.request(
+        full_path,
+        headers={'Metadata-Flavor': 'Google'}
+    )
     if response.status == http_client.OK:
-        content = _from_bytes(content)
-        return None, content
+        return None, json.loads(_from_bytes(content))
     else:
-        return response, content
+        raise AttributeError(
+            (
+                'Failed to retrieve the {} from the Google Compute Engine'
+                'metadata service. Response:\n{}'
+            ).format(full_path, response)
+        )
+
+
+def _get_access_token(http_request, email='default'):
+    token_json = _get_metadata(
+        'service-accounts',
+        email,
+        'acquire',
+        http_request=http_request
+    )
+    return token_json.get('accessToken'), token_json.get('expiresAt')
 
 
 class AppAssertionCredentials(AssertionCredentials):
@@ -86,7 +87,7 @@ class AppAssertionCredentials(AssertionCredentials):
     """
 
     @util.positional(2)
-    def __init__(self, scope='', **kwargs):
+    def __init__(self, scope='', service_account_email='default', **kwargs):
         """Constructor for AppAssertionCredentials
 
         Args:
@@ -97,15 +98,27 @@ class AppAssertionCredentials(AssertionCredentials):
         """
         if scope:
             warnings.warn(_SCOPES_WARNING)
-        # This is just provided for backwards compatibility, but is not
-        # used by this class.
-        self.scope = util.scopes_to_string(scope)
+
+        service_account_json = _get_metadata(
+            'service-accounts',
+            service_account_email
+        )
+
+        self.scope = util.scopes_to_string(
+            service_account_json.get('scopes', [])
+        )
+        self.service_account_email = service_account_json.get(
+            'serviceAccount',
+            None
+        )
+
+        self._project_id = None
+
         self.kwargs = kwargs
 
         # Assertion type is no longer used, but still in the
         # parent class signature.
         super(AppAssertionCredentials, self).__init__(None)
-        self._service_account_email = None
 
     @classmethod
     def from_json(cls, json_data):
@@ -125,21 +138,13 @@ class AppAssertionCredentials(AssertionCredentials):
         Raises:
             HttpAccessTokenRefreshError: When the refresh fails.
         """
-        response, content = http_request(
-            META, headers={'Metadata-Flavor': 'Google'})
-        content = _from_bytes(content)
-        if response.status == http_client.OK:
-            try:
-                token_content = json.loads(content)
-            except Exception as e:
-                raise HttpAccessTokenRefreshError(str(e),
-                                                  status=response.status)
-            self.access_token = token_content['access_token']
-        else:
-            if response.status == http_client.NOT_FOUND:
-                content += (' This can occur if a VM was created'
-                            ' with no service account or scopes.')
-            raise HttpAccessTokenRefreshError(content, status=response.status)
+        try:
+            self.access_token, self.token_expiry = _get_access_token(
+                http_request,
+                email=self.service_account_email
+            )
+        except AttributeError as e:
+            raise HttpAccessTokenRefreshError(str(e))
 
     @property
     def serialization_data(self):
@@ -151,6 +156,12 @@ class AppAssertionCredentials(AssertionCredentials):
 
     def create_scoped(self, scopes):
         return AppAssertionCredentials(scopes, **self.kwargs)
+
+    @property
+    def project_id(self):
+        if not self._project_id:
+            self._project_id = _get_metadata('project-id')
+        return self._project_id
 
     def sign_blob(self, blob):
         """Cryptographically sign a blob (of bytes).
@@ -166,29 +177,5 @@ class AppAssertionCredentials(AssertionCredentials):
             NotImplementedError, always.
         """
         raise NotImplementedError(
-            'Compute Engine service accounts cannot sign blobs')
-
-    @property
-    def service_account_email(self):
-        """Get the email for the current service account.
-
-        Uses the Google Compute Engine metadata service to retrieve the email
-        of the default service account.
-
-        Returns:
-            string, The email associated with the Google Compute Engine
-            service account.
-
-        Raises:
-            AttributeError, if the email can not be retrieved from the Google
-            Compute Engine metadata service.
-        """
-        if self._service_account_email is None:
-            failure, email = _get_service_account_email()
-            if failure is None:
-                self._service_account_email = email
-            else:
-                raise AttributeError('Failed to retrieve the email from the '
-                                     'Google Compute Engine metadata service',
-                                     failure, email)
-        return self._service_account_email
+            'Compute Engine service accounts cannot sign blobs'
+        )
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -45,7 +45,7 @@ can't be overridden in the request.
 """
 
 
-def _get_metadata(http_request=None, path=None):
+def _get_metadata(http_request=None, path=None, recursive=True):
     """Gets a JSON object from the specified path on the Metadata Server
     Args:
         http_request: an httplib2.Http().request object or equivalent
@@ -63,13 +63,18 @@ def _get_metadata(http_request=None, path=None):
     if not http_request:
         http_request = httplib2.Http().request
 
-    full_path = _METADATA_ROOT + '/'.join(path) + '/?recursive=true'
+    r_string = '/?recursive=true' if recursive else ''
+    full_path = _METADATA_ROOT + '/'.join(path) + r_string
     response, content = http_request(
         full_path,
         headers={'Metadata-Flavor': 'Google'}
     )
     if response.status == http_client.OK:
-        return json.loads(_from_bytes(content))
+        decoded = _from_bytes(content)
+        if recursive:
+            return json.loads(decoded)
+        else:
+            return decoded
     else:
         msg = (
             'Failed to retrieve {path} from the Google Compute Engine'
@@ -180,7 +185,10 @@ class AppAssertionCredentials(AssertionCredentials):
     @property
     def project_id(self):
         if not self._project_id:
-            self._project_id = _get_metadata(path=['project', 'project-id'])
+            self._project_id = _get_metadata(
+                path=['project', 'project-id'],
+                recursive=False
+            )
         return self._project_id
 
     @property
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -64,6 +64,7 @@ def _get_metadata(http_request=None, path=None, recursive=True):
             with which to make the call to the metadata server
         path: a list of strings denoting the metadata server request
             path.
+
     Returns:
         A deserialized JSON object representing the data returned
         from the metadata server
@@ -102,6 +103,7 @@ def _get_access_token(http_request, email):
         http_request: an httplib2.Http().request object or equivalent
             with which to make the call to the metadata server
         email: The service account email to request an access token with
+
     Returns:
         A tuple (accessToken, token expiry)
     """
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -71,7 +71,7 @@ def _get_metadata(http_request=None, path=None, recursive=True):
     )
     if response.status == http_client.OK:
         decoded = _from_bytes(content)
-        if recursive:
+        if response['content-type'] == 'application/json':
             return json.loads(decoded)
         else:
             return decoded
@@ -99,7 +99,8 @@ def _get_access_token(http_request, email):
             'service-accounts',
             email,
             'token'
-        ]
+        ],
+        recursive=False
     )
     token_expiry = datetime.datetime.now() + datetime.timedelta(
         seconds=token_json.get('expires_in')
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -16,19 +16,18 @@
 
 Utilities for making it easier to use OAuth 2.0 on Google Compute Engine.
 """
-
+import datetime
 import json
 import logging
 import warnings
-import datetime
 
 import httplib2
 from six.moves import http_client
 
-from oauth2client._helpers import _from_bytes
 from oauth2client.client import HttpAccessTokenRefreshError
 from oauth2client.client import AssertionCredentials
-
+from oauth2client._helpers import _from_bytes
+from oauth2client.contrib.iam_signer import IAMSigner
 
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -137,9 +137,6 @@ class AppAssertionCredentials(AssertionCredentials):
             service_account_info:
                 Deserialized JSON object, returned by self.service_account_info
         """
-        if scope:
-            warnings.warn(_SCOPES_WARNING)
-
         self._service_account_info = service_account_info or {
             'email': service_account_email
         }
@@ -152,6 +149,9 @@ class AppAssertionCredentials(AssertionCredentials):
         # parent class signature.
         super(AppAssertionCredentials, self).__init__(None)
 
+        if scope:
+            self.scopes = scope
+
     @property
     def service_account_info(self):
         return self._get_service_account_info()
@@ -162,8 +162,14 @@ class AppAssertionCredentials(AssertionCredentials):
 
     @scopes.setter
     def scopes(self, value):
-        """ Scopes on metadata service account are immutable """
-        pass
+        """ Scopes on metadata service account are immutable.
+        We should fail harder in the case that a user sets scopes
+        which are not available on the instance
+        """
+        if self.has_scopes(value):
+            warnings.warn(_SCOPES_WARNING)
+        else:
+            raise ValueError(_SCOPES_WARNING)
 
     @property
     def service_account_email(self):
@@ -228,11 +234,12 @@ class AppAssertionCredentials(AssertionCredentials):
         except (AttributeError, ValueError) as e:
             raise HttpAccessTokenRefreshError(str(e))
 
-    def create_scoped_required(self):
-        return False
-
     def create_scoped(self, scopes):
-        return AppAssertionCredentials(scopes, **self.kwargs)
+        new = AppAssertionCredentials(
+            service_account_info=self.service_account_info
+        )
+        new.scopes = scopes
+        return new
 
     def sign_blob(self, blob):
         """Cryptographically sign a blob (of bytes).
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -22,6 +22,7 @@ import logging
 import warnings
 
 import httplib2
+import time
 from six.moves import http_client
 
 from oauth2client._helpers import _from_bytes
@@ -35,7 +36,7 @@ __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 logger = logging.getLogger(__name__)
 
 # URI Template for the endpoint that returns access_tokens.
-_METADATA_ROOT = 'http://metadata.google.internal/0.1/meta-data'
+_METADATA_ROOT = 'http://metadata.google.internal/v1/computeMetadata'
 _SCOPES_WARNING = """\
 You have requested explicit scopes to be used with a GCE service account.
 Using this argument will have no effect on the actual scopes for tokens
@@ -45,9 +46,21 @@ can't be overridden in the request.
 
 
 def _get_metadata(http_request=None, *path):
+    """
+    Args:
+        http_request: an httplib2.Http().request object or equivalent
+            with which to make the call to the metadata server
+        *path: a list of strings denoting the metadata server request
+            path.
+    Returns:
+        A deserialized JSON object representing the data returned
+        from the metadata server
+    """
+
     if not http_request:
         http_request = httplib2.Http().request
-    full_path = "/".join(path.insert(0, _METADATA_ROOT))
+
+    full_path = '/'.join(path.insert(0, _METADATA_ROOT)) + '/?recursive=true'
     response, content = http_request(
         full_path,
         headers={'Metadata-Flavor': 'Google'}
@@ -55,26 +68,50 @@ def _get_metadata(http_request=None, *path):
     if response.status == http_client.OK:
         return json.loads(_from_bytes(content))
     else:
-        raise AttributeError(
-            (
-                'Failed to retrieve {} from the Google Compute Engine'
-                'metadata service. Response:\n{}'
-            ).format(full_path, response)
-        )
+        msg = (
+            'Failed to retrieve {} from the Google Compute Engine'
+            'metadata service. Response:\n{}'
+        ).format(full_path, response)
+        raise AttributeError(msg)
 
 
 def _get_access_token(http_request, email):
+    """
+    Args:
+        http_request: an httplib2.Http().request object or equivalent
+            with which to make the call to the metadata server
+        email: The service account email to request an access token with
+    Returns:
+        A tuple (accessToken, token expiry)
+    """
     token_json = _get_metadata(
+        'instance',
         'service-accounts',
         email,
         'acquire',
         http_request=http_request
     )
-    return token_json.get('accessToken'), int(token_json.get('expiresAt'))
+    token_expiry = int(token_json.get('expires_in')) + time.time()
+    return token_json.get('access_token'), token_expiry
 
 
 def _get_service_account_info(email, http_request=None):
+    """
+    Args:
+        http_request: an httplib2.Http().request object or equivalent
+            with which to make the call to the metadata server
+        email: The service account email to request an access token with
+    Returns:
+        A deserialized JSON service account object of the form:
+            {
+               'aliases': [],
+               'scopes': [],
+               'email': "a@example.com"
+            }
+    """
+
     return _get_metadata(
+        'instance',
         'service-accounts',
         email,
         http_request=http_request
@@ -113,8 +150,7 @@ class AppAssertionCredentials(AssertionCredentials):
         if scope:
             warnings.warn(_SCOPES_WARNING)
 
-        self._service_account_email = service_account_email
-        self._service_account_info = None
+        self._service_account_info = {'email': service_account_email}
         self._project_id = None
 
         self.kwargs = kwargs
@@ -124,28 +160,35 @@ class AppAssertionCredentials(AssertionCredentials):
         super(AppAssertionCredentials, self).__init__(None)
 
     @property
+    def service_account_info(self):
+        if self._service_account_info.get('email', 'default') == 'default':
+            self._service_account_info = _get_service_account_info(
+                self._service_account_info.get('email', 'default')
+            )
+        return self._service_account_info
+
+    @property
     def scopes(self):
         return self._retrieve_scopes(httplib2.Http().request)
 
     @property
     def service_account_email(self):
-        if not self._service_account_info:
-            self._service_account_info = _get_service_account_info(
-                self._service_account_email
-            )
-        return self._service_account_info['serviceAccount']
+        return self.service_account_info['email']
+
+    @property
+    def project_id(self):
+        if not self._project_id:
+            self._project_id = _get_metadata('project', 'project-id')
+        return self._project_id
+
+    @property
+    def serialization_data(self):
+        return {
+            'email': self.service_account_email
+        }
 
     def _retrieve_scopes(self, http_request):
-        if not self._service_account_info:
-            self._service_account_info = _get_service_account_info(
-                self._service_account_email,
-                http_request=http_request
-            )
-        return self._service_account_info['scopes']
-
-    @classmethod
-    def from_json(cls, json_data):
-        return AppAssertionCredentials(email=json_data['email'])
+        return self.service_account_info['scopes']
 
     def _refresh(self, http_request):
         """Refreshes the access_token.
@@ -168,23 +211,11 @@ class AppAssertionCredentials(AssertionCredentials):
         except Exception as e:
             raise HttpAccessTokenRefreshError(str(e))
 
-    @property
-    def serialization_data(self):
-        return {
-            'email': self.service_account_email
-        }
-
     def create_scoped_required(self):
         return False
 
     def create_scoped(self, scopes):
         return AppAssertionCredentials(scopes, **self.kwargs)
-
-    @property
-    def project_id(self):
-        if not self._project_id:
-            self._project_id = _get_metadata('project-id')
-        return self._project_id
 
     def sign_blob(self, blob):
         """Cryptographically sign a blob (of bytes).
@@ -201,4 +232,10 @@ class AppAssertionCredentials(AssertionCredentials):
         """
         raise NotImplementedError(
             'Compute Engine service accounts cannot sign blobs'
+        )
+
+    @classmethod
+    def from_json(cls, json_data):
+        return AppAssertionCredentials(
+            service_account_email=json_data['email']
         )
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -163,9 +163,10 @@ class AppAssertionCredentials(AssertionCredentials):
         super(AppAssertionCredentials, self).__init__(
             None, scopes=scope, **kwargs)
 
-        self._service_account_info = {}
-        if service_account_email:
-            self._service_account_info['email'] = service_account_email
+        if service_account_email is None:
+            self._service_account_info = {}
+        else:
+            self._service_account_info = {'email': service_account_email}
 
         self._project_id = None
         self._partial = True
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -36,7 +36,7 @@ __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 logger = logging.getLogger(__name__)
 
 # URI Template for the endpoint that returns access_tokens.
-_METADATA_ROOT = 'http://metadata.google.internal/v1/computeMetadata/'
+_METADATA_ROOT = 'http://metadata.google.internal/computeMetadata/v1/'
 _SCOPES_WARNING = """\
 You have requested explicit scopes to be used with a GCE service account.
 Using this argument will have no effect on the actual scopes for tokens
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -20,9 +20,9 @@ Utilities for making it easier to use OAuth 2.0 on Google Compute Engine.
 import json
 import logging
 import warnings
+import datetime
 
 import httplib2
-import datetime
 from six.moves import http_client
 
 from oauth2client._helpers import _from_bytes
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -74,7 +74,7 @@ def _get_metadata(http_request=None,
     )
     if response.status == http_client.OK:
         decoded = _from_bytes(content)
-        if recursive:
+        if returns_json:
             return json.loads(decoded)
         else:
             return decoded
@@ -102,7 +102,8 @@ def _get_access_token(http_request, email):
             'service-accounts',
             email,
             'token'
-        ]
+        ],
+        recursive=False
     )
     token_expiry = datetime.datetime.now() + datetime.timedelta(
         seconds=token_json.get('expires_in')
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -36,7 +36,7 @@ logger = logging.getLogger(__name__)
 # URI Template for the endpoint that returns access_tokens.
 _METADATA_ROOT = 'http://metadata.google.internal/computeMetadata/v1/'
 # Backwards Compat
-META = _METADATA_ROOT + '/instance/service-accounts/default/token'
+META = _METADATA_ROOT + 'instance/service-accounts/default/token'
 _SCOPES_WARNING = """\
 You have specified explicit scopes to be used with a GCE service account
 credentials, and these scopes *are* present on the credentials.
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -45,7 +45,10 @@ can't be overridden in the request.
 """
 
 
-def _get_metadata(http_request=None, path=None, recursive=True):
+def _get_metadata(http_request=None,
+                  path=None,
+                  recursive=True,
+                  returns_json=True):
     """Gets a JSON object from the specified path on the Metadata Server
     Args:
         http_request: an httplib2.Http().request object or equivalent
@@ -187,7 +190,8 @@ class AppAssertionCredentials(AssertionCredentials):
         if not self._project_id:
             self._project_id = _get_metadata(
                 path=['project', 'project-id'],
-                recursive=False
+                recursive=False,
+                returns_json=False
             )
         return self._project_id
 
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -36,7 +36,7 @@ __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 logger = logging.getLogger(__name__)
 
 # URI Template for the endpoint that returns access_tokens.
-_METADATA_ROOT = 'http://metadata.google.internal/v1/computeMetadata'
+_METADATA_ROOT = 'http://metadata.google.internal/v1/computeMetadata/'
 _SCOPES_WARNING = """\
 You have requested explicit scopes to be used with a GCE service account.
 Using this argument will have no effect on the actual scopes for tokens
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -72,9 +72,9 @@ def _get_metadata(http_request=None, path=None):
         return json.loads(_from_bytes(content))
     else:
         msg = (
-            'Failed to retrieve {} from the Google Compute Engine'
-            'metadata service. Response:\n{}'
-        ).format(full_path, response)
+            'Failed to retrieve {path} from the Google Compute Engine'
+            'metadata service. Response:\n{error}'
+        ).format(path=full_path, error=response)
         raise AttributeError(msg)
 
 
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -74,7 +74,7 @@ def _get_metadata(http_request=None,
     )
     if response.status == http_client.OK:
         decoded = _from_bytes(content)
-        if returns_json:
+        if recursive:
             return json.loads(decoded)
         else:
             return decoded
@@ -102,8 +102,7 @@ def _get_access_token(http_request, email):
             'service-accounts',
             email,
             'token'
-        ],
-        recursive=False
+        ]
     )
     token_expiry = datetime.datetime.now() + datetime.timedelta(
         seconds=token_json.get('expires_in')
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -35,8 +35,10 @@ logger = logging.getLogger(__name__)
 
 # URI Template for the endpoint that returns access_tokens.
 _METADATA_ROOT = 'http://metadata.google.internal/computeMetadata/v1/'
+# Backwards Compat
+META = _METADATA_ROOT + '/instance/service-accounts/default/token'
 _SCOPES_WARNING = """\
-You have requested explicit scopes to be used with a GCE service account
+You have specified explicit scopes to be used with a GCE service account
 credentials, and these scopes *are* present on the credentials.
 However, setting scopes on the GCE service account credentials has no effect
 on the actual scopes for tokens requested. The credentials scopes
@@ -44,7 +46,7 @@ are set at VM instance creation time and can't be overridden in the request.
 To learn more go to https://cloud.google.com/compute/docs/authentication .
 """
 _SCOPES_ERROR = """\
-You have requested explicit scopes to be used with a GCE service account
+You have specified explicit scopes to be used with a GCE service account
 which are not available on the credentials. The scopes are set at VM instance
 creation time and can't be overridden in the request.
 To learn more go to https://cloud.google.com/compute/docs/authentication .
@@ -175,6 +177,7 @@ class AppAssertionCredentials(AssertionCredentials):
 
         # This function call must be made because AssertionCredentials
         # will not pass the scopes kwarg to parent class
+        # In a breaking version change, remove the scope kwarg
         self._check_scopes_and_notify(scope)
 
     @property
@@ -263,8 +266,7 @@ class AppAssertionCredentials(AssertionCredentials):
             self._service_account_info.get('email', 'default'))
 
     def create_scoped(self, scopes):
-        # Trigger warning or error based on scopes
-        self.scopes = scopes
+        self._check_scopes_and_notify(scopes)
         # No need for new object creation
         return self
 
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -112,20 +112,16 @@ def _get_access_token(http_request, email):
     try:
         token_json = _get_metadata(
             http_request=http_request,
-            path=[
-                'instance',
-                'service-accounts',
-                email,
-                'token'
-            ],
-            recursive=False
-        )
+            path=['instance',
+                  'service-accounts',
+                  email,
+                  'token'],
+            recursive=False)
     except MetadataServerHttpError as failed_fetch:
         raise HttpAccessTokenRefreshError(str(failed_fetch))
 
     token_expiry = _NOW() + datetime.timedelta(
-        seconds=token_json['expires_in']
-    )
+        seconds=token_json['expires_in'])
     return token_json['access_token'], token_expiry
 
 
@@ -144,9 +140,8 @@ class AppAssertionCredentials(AssertionCredentials):
 
     def __init__(self,
                  scope=None,
-                 service_account_email='default',
-                 service_account_info=None,
-                 **unused_kwargs):
+                 service_account_email=None,
+                 **kwargs):
         """Constructor for AppAssertionCredentials
 
         Args:
@@ -160,31 +155,20 @@ class AppAssertionCredentials(AssertionCredentials):
                 service accounts, or left blank to use the default service
                 account for the instance. The default service account is
                 usually the shared Compute Engine service account.
-            service_account_info:
-                (Optional) Dictionary containing information about a
-                service account (typically deserialized from JSON).
-                If passed in, will be used as ``service_account_info``
-                property. Otherwise, the Compute Engine metadata server
-                will be used to determine this value. Additionally,
-                if present the 'email' field in this argument must be
-                provided and will override the ``service_account_email``
-                argument.
         """
         # Assertion type is no longer used, but still in the
         # parent class signature.
         super(AppAssertionCredentials, self).__init__(
-            None,
-            scopes=scope,
-            **unused_kwargs
-        )
+            None, scopes=scope, **kwargs)
 
-        self._service_account_info = service_account_info or {
-            'email': service_account_email
-        }
+        self._service_account_info = {}
+        if service_account_email:
+            self._service_account_info['email'] = service_account_email
+
         self._project_id = None
-        self._partial = service_account_info is None
+        self._partial = True
 
-        self.kwargs = unused_kwargs
+        self.kwargs = kwargs
 
         # This function call must be made because AssertionCredentials
         # will not pass the scopes kwarg to parent class
@@ -216,9 +200,7 @@ class AppAssertionCredentials(AssertionCredentials):
     def project_id(self):
         if not self._project_id:
             self._project_id = _get_metadata(
-                path=['project', 'project-id'],
-                recursive=False
-            )
+                path=['project', 'project-id'], recursive=False)
         return self._project_id
 
     @property
@@ -249,13 +231,10 @@ class AppAssertionCredentials(AssertionCredentials):
         """
         if self._partial:
             self._service_account_info = _get_metadata(
-                path=[
-                    'instance',
-                    'service-accounts',
-                    self._service_account_info['email'],
-                ],
-                http_request=http_request
-            )
+                path=['instance',
+                      'service-accounts',
+                      self._service_account_info.get('email', 'default')],
+                http_request=http_request)
             self._partial = False
         return self._service_account_info
 
@@ -277,7 +256,8 @@ class AppAssertionCredentials(AssertionCredentials):
             HttpAccessTokenRefreshError: When the refresh fails.
         """
         self.access_token, self.token_expiry = _get_access_token(
-            http_request, self._service_account_info['email'])
+            http_request,
+            self._service_account_info.get('email', 'default'))
 
     def create_scoped(self, scopes):
         # Trigger warning or error based on scopes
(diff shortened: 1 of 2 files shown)


oauth2client/contrib/gce.py
```diff
@@ -86,31 +88,59 @@ class AppAssertionCredentials(AssertionCredentials):
...
+                service accounts, or left blank to use the default service account
+                for the instance. Usually the compute engine service account.
         """
         if scope:
             warnings.warn(_SCOPES_WARNING)
-        # This is just provided for backwards compatibility, but is not
-        # used by this class.
-        self.scope = util.scopes_to_string(scope)
+
+        self.scopes = None
```
elibixby: I'm concerned that the parent classes have this property exposed as a public member, 

Should this be an `@property` , or is implementing `_retrieve_scopes` enough?

oauth2client/contrib/gce.py
```diff
@@ -125,26 +155,19 @@ def _refresh(self, http_request):
...
+        try:
+            self.access_token, self.token_expiry = _get_access_token(
+                http_request,
+                email=self.service_account_email
+            )
+        except Error as e:
+            raise HttpAccessTokenRefreshError(str(e))
 
     @property
     def serialization_data(self):
```
elibixby: I'm not entirely sure how the various serialization members interact here. Clarity would be appreciated. Since calls to the metadata server are low cost I'm only serializing email here, but I suspect there are other things I need to implement?

elibixby:
I am still in the process of adding tests for this PR. Do Not Merge.

oauth2client/contrib/gce.py
```diff
@@ -36,10 +35,7 @@
 logger = logging.getLogger(__name__)
 
 # URI Template for the endpoint that returns access_tokens.
-_METADATA_ROOT = ('http://metadata.google.internal/computeMetadata/v1/'
-                  'instance/service-accounts/default/')
-META = _METADATA_ROOT + 'token'
-_DEFAULT_EMAIL_METADATA = _METADATA_ROOT + 'email'
+_METADATA_ROOT = 'http://metadata.google.internal/0.1/meta-data'
```
dhermes: The 0.1 endpoint was replaced in #387, this would be a reversion

oauth2client/contrib/gce.py
```diff
@@ -48,28 +44,41 @@
...
-                      signature of httplib2.Http.request, used to make
-                      the request to the metadata service.
-
-    Returns:
-        tuple, A pair where the first entry is an optional response (from a
-        failed request) and the second is service account email found (as
-        a string).
-    """
-    if http_request is None:
+def _get_metadata(http_request=None, *path):
```
dhermes: Surprised to find out this is valid Python! TIL

oauth2client/contrib/gce.py
```diff
@@ -48,28 +44,41 @@
...
-    Args:
-        http_request: callable, (Optional) a callable that matches the method
-                      signature of httplib2.Http.request, used to make
-                      the request to the metadata service.
-
-    Returns:
-        tuple, A pair where the first entry is an optional response (from a
-        failed request) and the second is service account email found (as
-        a string).
-    """
```
dhermes, line 63: Replacing a function with a docstring with one that doesn't have one makes me sad. Please add?

oauth2client/contrib/gce.py
```diff
@@ -48,28 +44,41 @@
...
-    if http_request is None:
+def _get_metadata(http_request=None, *path):
+    if not http_request:
         http_request = httplib2.Http().request
+    full_path = "/".join(path.insert(0, _METADATA_ROOT))
```
dhermes: Mixing `"` and `'`. This file uses `'`

oauth2client/contrib/gce.py
```diff
@@ -48,28 +44,41 @@
...
-        content = _from_bytes(content)
-        return None, content
+        return json.loads(_from_bytes(content))
     else:
-        return response, content
+        raise AttributeError(
+            (
+                'Failed to retrieve {} from the Google Compute Engine'
+                'metadata service. Response:\n{}'
+            ).format(full_path, response)
```
dhermes: A lot going on here. Create the error message first and then just `raise AttributeError(msg)`

oauth2client/contrib/gce.py
```diff
@@ -86,31 +95,57 @@ class AppAssertionCredentials(AssertionCredentials):
...
+                self._service_account_email,
+                http_request=http_request
+            )
+        return self._service_account_info['scopes']
 
     @classmethod
     def from_json(cls, json_data):
-        data = json.loads(_from_bytes(json_data))
-        return AppAssertionCredentials(data['scope'])
+        return AppAssertionCredentials(email=json_data['email'])
```
dhermes: Intended to be `service_account_email`? Also, The old method also used `scope` so it shouldn't be removed?

oauth2client/contrib/gce.py
```diff
@@ -125,26 +160,19 @@ def _refresh(self, http_request):
...
-            if response.status == http_client.NOT_FOUND:
-                content += (' This can occur if a VM was created'
-                            ' with no service account or scopes.')
-            raise HttpAccessTokenRefreshError(content, status=response.status)
+        try:
+            self.access_token, self.token_expiry = _get_access_token(
+                http_request,
+                email=self.service_account_email
+            )
+        except Exception as e:
```
dhermes: Please use a more focused exception (I just realized that was the old exception, still `ValueError` should suffice for `json.loads` failures)

oauth2client/contrib/gce.py
```diff
@@ -86,31 +95,57 @@ class AppAssertionCredentials(AssertionCredentials):
...
         # parent class signature.
         super(AppAssertionCredentials, self).__init__(None)
-        self._service_account_email = None
+
+    @property
+    def scopes(self):
+        return self._retrieve_scopes(httplib2.Http().request)
+
+    @property
+    def service_account_email(self):
```
dhermes, line 207: It's a bit confusing that this can be different than `self._service_account_email`

oauth2client/contrib/gce.py
elibixby: Yeah, I was using it because I didn't think the new endpoint supported some of the info I needed, but it turns out there was a magic query arg to make it give me scopes etc. Will update.

oauth2client/contrib/gce.py
elibixby, line 207: So, a potential problem
1. Someone creates a `AppAssertionCredentials` object on instance `A` with default service account `b@example.com`,
2. Serializes it
3. Deserializes it on instance `C` with default service account `d@example.com` and no access to `b@example.com`

Should
a. `ValueError` raised the first time access token is requested (`get_access_token` requests access token for an inaccessible email, because the deserialized email is `b@example.com` which `C` doesn't have access to)
b. The credentials work fine, provided `d@example.com` has the authorization

I'm leaning towards (a), so objects should have effectively immutable email address, in which case we could dump `self._service_account_email` and just use a partial `self._service_account_info` object to represent the "default" service account.

oauth2client/contrib/gce.py
elibixby: A scope argument doesn't actually do anything except throw a warning, so I thought I'd leave it out to avoid repeatedly throwing warnings on serialization/deserialization, thoughts?

oauth2client/contrib/gce.py
elibixby: Also, it unfortunately looks like the _removed_ support for `expires_at` going from 0.1 to v1

> _<

oauth2client/contrib/gce.py
elibixby, line 207: I went ahead an implemented behavior for (a) PTAL

oauth2client/contrib/gce.py
```diff
@@ -86,31 +132,69 @@ class AppAssertionCredentials(AssertionCredentials):
...
+
+    @property
+    def project_id(self):
+        if not self._project_id:
+            self._project_id = _get_metadata('project', 'project-id')
+        return self._project_id
+
+    @property
+    def serialization_data(self):
+        return self.service_account_info
```
elibixby: @dhermes @jonparrott 
Should this:
(a) serialize using the external property, meaning if a metadata server request for info hasn't been made, it will be made at serialization time, or 
(b) serialize the internal property, potentially meaning that it could be serialized with the email as 'default' which ambiguously maps to actual service accounts. 

I went with (a), but am open to other opinions

elibixby:
@dhermes @jonparrott 

This is ready for closer inspection and potentially a merge

oauth2client/contrib/gce.py
```diff
@@ -48,28 +45,61 @@
 """
 
 
-def _get_service_account_email(http_request=None):
-    """Get the GCE service account email from the current environment.
-
+def _get_metadata(http_request=None, path=None):
+    """
```
theacodes: I'd love a docstring that says what this does. :)

oauth2client/contrib/gce.py
```diff
@@ -48,28 +45,61 @@
...
-        content = _from_bytes(content)
-        return None, content
+        return json.loads(_from_bytes(content))
     else:
-        return response, content
+        msg = (
+            'Failed to retrieve {path} from the Google Compute Engine'
+            'metadata service. Response:\n{error}'
+        ).format(path=full_path, error=response)
+        raise AttributeError(msg)
```
theacodes: `AttributeError`?

oauth2client/contrib/gce.py
```diff
@@ -48,28 +45,61 @@
...
+
+
+def _get_access_token(http_request, email):
+    """
```
theacodes: Short docstring here would be nice as well.

oauth2client/contrib/gce.py
```diff
@@ -48,28 +45,61 @@
...
     """
-    if http_request is None:
+
+    if not path:
+        path = []
+
+    if not http_request:
         http_request = httplib2.Http().request
+
+    full_path = _METADATA_ROOT + '/'.join(path) + '/?recursive=true'
```
theacodes: `urljoin`? Why not just specify the path as a path instead of a list?

oauth2client/contrib/gce.py
```diff
@@ -48,28 +45,61 @@
...
-        a string).
+        A deserialized JSON object representing the data returned
+        from the metadata server
     """
-    if http_request is None:
+
+    if not path:
+        path = []
+
+    if not http_request:
```
theacodes, line 79: thanks for doing this, it'll make it much easier for me to rip httplib2 out later.

oauth2client/contrib/gce.py
```diff
@@ -85,32 +115,100 @@ class AppAssertionCredentials(AssertionCredentials):
...
-    def from_json(cls, json_data):
-        data = json.loads(_from_bytes(json_data))
-        return AppAssertionCredentials(data['scope'])
+    @property
+    def service_account_info(self):
+        return self._get_service_account_info()
+
+    @property
+    def scopes(self):
+        return self.service_account_info['scopes']
```
theacodes, line 200: If there a chance this key doesn't exist? Would be be better to do `self.service_account_info.get('scopes', None)`?

oauth2client/contrib/gce.py
```diff
@@ -85,32 +115,100 @@ class AppAssertionCredentials(AssertionCredentials):
...
+
+    @scopes.setter
```
theacodes, line 202: You can leave this out.

oauth2client/contrib/gce.py
```diff
@@ -166,29 +251,19 @@ def sign_blob(self, blob):
...
-        Uses the Google Compute Engine metadata service to retrieve the email
-        of the default service account.
+            'Compute Engine service accounts cannot sign blobs'
+        )
 
-        Returns:
-            string, The email associated with the Google Compute Engine
-            service account.
+    def to_json(self):
+        # Why is this not default -_-
```
theacodes: I agree, but please leave this comment out. :)

oauth2client/contrib/gce.py
```diff
@@ -166,29 +251,19 @@ def sign_blob(self, blob):
...
-            'Compute Engine service accounts cannot sign blobs')
-
-    @property
-    def service_account_email(self):
-        """Get the email for the current service account.
-
-        Uses the Google Compute Engine metadata service to retrieve the email
-        of the default service account.
+            'Compute Engine service accounts cannot sign blobs'
+        )
```
theacodes: nit: You can put this at the end of the line above.

oauth2client/contrib/gce.py
elibixby, line 202: Unfortunately the parent class tries to set scope in the constructor, so I need to provide a method for it, and make it do nothing.

oauth2client/contrib/gce.py
elibixby, line 200: service_account_info will retrieve the info from the metadata server if it's partial, which will always include scopes, so I think this is fine? I'll add a comment to that effect.

oauth2client/contrib/gce.py
elibixby: I didn't want all the other calls to have to do their own join. idk, seemed a little cleaner. Also I don't think urljoin would be any cleaner, because the metadata root isn't just a host, and the usage of the query string is nonstandard (the trailing slash on the path is required)

oauth2client/contrib/gce.py
elibixby: `ValueError`?

theacodes: That would be more appropriate, but it might make sense to make a new exception.

oauth2client/contrib/gce.py
theacodes, line 200: That's fair, just making sure.

oauth2client/contrib/gce.py
theacodes, line 202: ∑(ﾟДﾟ)

elibixby, line 202: When I tried to make scopes error on setting and it broke 17 tests I was like:
(╯°□°）╯︵ ┻━┻

oauth2client/contrib/gce.py
elibixby: I'll just do `ValueError` because that's what's thrown by the `json` module and it's going to be caught and wrapped up as an `HttpAccessTokenError` most of the time anyway

elibixby:
FYI I am preparing the follow up PR here: https://github.com/elibixby/oauth2client/compare/master...elibixby:iamsigner

elibixby:
@dhermes Follow up PR with IAM Blob Signing for `gce.AppAssertionCredentials` is ready when you are.

Thanks for reviewing!

nathanielmanistaatgoogle:
I intend to review this also after @dhermes and @jonparrott are satisfied.

oauth2client/contrib/gce.py
```diff
@@ -48,28 +45,67 @@
...
+def _get_metadata(http_request=None, path=None, recursive=True):
+    """Gets a JSON object from the specified path on the Metadata Server
     Args:
-        http_request: callable, (Optional) a callable that matches the method
-                      signature of httplib2.Http.request, used to make
-                      the request to the metadata service.
-
+        http_request: an httplib2.Http().request object or equivalent
+            with which to make the call to the metadata server
+        *path: a list of strings denoting the metadata server request
```
dhermes: s/`*path`/`path`

oauth2client/contrib/gce.py
```diff
@@ -48,28 +45,67 @@
...
+        msg = (
+            'Failed to retrieve {path} from the Google Compute Engine'
+            'metadata service. Response:\n{error}'
+        ).format(path=full_path, error=response)
+        raise ValueError(msg)
+
+
+def _get_access_token(http_request, email):
+    """Get an access token for the specified email from the Metadata Server.
+    Args:
```
dhermes, line 105: Empty line between docstring one-liner and args

oauth2client/contrib/gce.py
```diff
@@ -48,28 +45,67 @@
...
+        http_request: an httplib2.Http().request object or equivalent
+            with which to make the call to the metadata server
+        email: The service account email to request an access token with
+    Returns:
```
dhermes, line 110: Ditto here on empty line before `Returns`

oauth2client/contrib/gce.py
```diff
@@ -48,28 +45,67 @@
...
+        path=[
+            'instance',
+            'service-accounts',
+            email,
+            'token'
+        ],
+        recursive=False
+    )
+    token_expiry = datetime.datetime.now() + datetime.timedelta(
+        seconds=token_json.get('expires_in')
```
dhermes: Not sure what the purpose of `token_json.get` is. The `KeyError` seems more useful than the ensuing `TypeError` if `expires_in` is absent.

``` python
>>> import datetime
>>> datetime.timedelta(seconds=None)
TypeError: unsupported type for timedelta seconds component: NoneType
```

oauth2client/contrib/gce.py
```diff
@@ -48,28 +45,67 @@
...
+    )
+    return token_json.get('access_token'), token_expiry
```
dhermes: Ditto here on the `KeyError`. If `access_token` isn't there, then this method has failed big time?

oauth2client/contrib/gce.py
```diff
@@ -85,32 +121,110 @@ class AppAssertionCredentials(AssertionCredentials):
     information to generate and refresh its own access tokens.
     """
 
-    @util.positional(2)
-    def __init__(self, scope='', **kwargs):
+    @util.positional(3)
```
dhermes: 1. @nathanielmanistaatgoogle Can confirm but I don't think we are supposed to change the number in `positional` on a non-breaking release.
2. If this is going to change, we should delete it.

oauth2client/contrib/gce.py
```diff
@@ -85,32 +121,110 @@ class AppAssertionCredentials(AssertionCredentials):
...
+            scope:
+                string or iterable of strings, scope(s) of the credentials
+                being requested. Using this argument will have no effect on
+                the actual scopes for tokens requested. These scopes are
+                set at VM instance creation time and won't change.
+            service_account_email:
+                the email for these credentials. This can be used with custom
+                service accounts, or left blank to use the default service
+                account for the instance. Usually the compute engine service
+                account.
```
dhermes: What does "Usually the compute engine service account" mean?

oauth2client/contrib/gce.py
```diff
@@ -85,32 +121,110 @@ class AppAssertionCredentials(AssertionCredentials):
...
+            service_account_info:
+                Deserialized JSON object, returned by self.service_account_info
```
dhermes: Doesn't saying `returned by...` jump the gun a bit?

oauth2client/contrib/gce.py
```diff
@@ -85,32 +121,110 @@ class AppAssertionCredentials(AssertionCredentials):
...
+
         if scope:
-            warnings.warn(_SCOPES_WARNING)
-        # This is just provided for backwards compatibility, but is not
-        # used by this class.
-        self.scope = util.scopes_to_string(scope)
+            if self.has_scopes(scope):
+                warnings.warn(_SCOPES_WARNING)
+            else:
+                raise ValueError(_SCOPES_WARNING)
```
dhermes: Can you explain this?

oauth2client/contrib/gce.py
```diff
@@ -85,32 +121,110 @@ class AppAssertionCredentials(AssertionCredentials):
...
-        self._service_account_email = None
 
-    @classmethod
-    def from_json(cls, json_data):
-        data = json.loads(_from_bytes(json_data))
-        return AppAssertionCredentials(data['scope'])
+    @property
+    def service_account_info(self):
+        """Info about this service account
+        By using _get_service_account_info this property is
```
dhermes: Empty line after one-liner.

**UPDATE**: Referring to a non-public helper in a public property doesn't seem like a good idea.

oauth2client/contrib/gce.py
dhermes, line 202: This is no fun. This setter should return an `AttributeError`. What value does the parent class provide?

oauth2client/contrib/gce.py
```diff
@@ -85,32 +121,110 @@ class AppAssertionCredentials(AssertionCredentials):
...
+            )
+        return self._project_id
+
+    @property
+    def serialization_data(self):
+        return {'service_account_info': self.service_account_info}
+
+    def _get_service_account_info(self, http_request=None):
+        """Retrieves the full info for a service account and caches it.
+        Args:
```
dhermes, line 224: Empty line before `Args`

oauth2client/contrib/gce.py
```diff
@@ -85,32 +121,110 @@ class AppAssertionCredentials(AssertionCredentials):
...
+            http_request: an httplib2.Http().request object or equivalent
+                with which to make the call to the metadata server
+        Returns:
```
dhermes, line 228: Empty line before `Returns`

oauth2client/contrib/gce.py
```diff
@@ -85,32 +121,110 @@ class AppAssertionCredentials(AssertionCredentials):
...
+            A deserialized JSON service account object of the form:
+                {
+                   'aliases': [],
+                   'scopes': [],
```
dhermes: This implies that `scopes` and `aliases` will be empty lists. Is that always the case?

oauth2client/contrib/gce.py
```diff
@@ -48,28 +45,67 @@
 """
 
 
-def _get_service_account_email(http_request=None):
-    """Get the GCE service account email from the current environment.
-
+def _get_metadata(http_request=None, path=None, recursive=True):
```
dhermes, line 61: There are quite a lot of calls to `_get_metadata` below, it seems this will slow down apps unnecessarily. Why not cache results or do more work in the constructor?

oauth2client/contrib/gce.py
```diff
@@ -125,32 +239,19 @@ def _refresh(self, http_request):
...
-            'Cannot serialize credentials for GCE service accounts.')
-
-    def create_scoped_required(self):
-        return False
+        try:
+            self.access_token, self.token_expiry = _get_access_token(
+                http_request,
+                self._service_account_info['email']
+            )
+        except ValueError as e:
```
dhermes: Please use a longer variable name than `e`, try `exc` or `err`. Is the `ValueError` caused by `json.loads`? If yes, why doesn't `_get_access_token` handle it?

oauth2client/contrib/gce.py
```diff
@@ -125,32 +239,19 @@ def _refresh(self, http_request):
...
+            raise HttpAccessTokenRefreshError(str(e))
 
     def create_scoped(self, scopes):
-        return AppAssertionCredentials(scopes, **self.kwargs)
+        return AppAssertionCredentials(
+            scope=scopes,
+            service_account_info=self.service_account_info
```
dhermes: Why are you dropping `kwargs` on the floor?

tests/contrib/test_gce.py
```diff
@@ -14,102 +14,195 @@
...
-from oauth2client._helpers import _to_bytes
-from oauth2client.client import AccessTokenRefreshError
-from oauth2client.client import Credentials
-from oauth2client.client import save_to_well_known_file
-from oauth2client.contrib.gce import _DEFAULT_EMAIL_METADATA
-from oauth2client.contrib.gce import _get_service_account_email
-from oauth2client.contrib.gce import _SCOPES_WARNING
-from oauth2client.contrib.gce import AppAssertionCredentials
-
+from oauth2client.client import HttpAccessTokenRefreshError,\
```
dhermes: Trailing backslashes never please. From the [Google Python style guide](https://google.github.io/styleguide/pyguide.html?showone=Line_length#Line_length):

> Do not use backslash line continuation.

---

Also, no comma imports (again from [style guide](https://google.github.io/styleguide/pyguide.html?showone=Imports_formatting#Imports_formatting)):

> Imports should be on separate lines.
> 
> Yes:
> 
> ``` python
> import os
> import sys
> ```
> 
> No:
> 
> ``` python
> import os, sys
> ```

tests/contrib/test_gce.py
```diff
@@ -14,102 +14,195 @@
...
 """Unit tests for oauth2client.contrib.gce."""
 
-import json
 from six.moves import http_client
-from six.moves import urllib
 import unittest2
-
+import time
 import mock
+import json
```
dhermes, line 18: 1. Please move `time` and `json` above these imports (they are in the stdlib, none of the rest are)
2. Please put imports in alphabetical order.

tests/contrib/test_gce.py
```diff
@@ -14,102 +14,195 @@
...
+}
+
+GET_ACCESS_TOKEN = 'oauth2client.contrib.gce._get_access_token'
+ONE_SECOND_TOKEN = {'access_token': '12345abcde', 'expires_in': 1}
+FOREVER_TOKEN = {'access_token': '12345abcde', 'expires_in': 999999999}
+
 
 class AppAssertionCredentialsTests(unittest2.TestCase):
 
+    # BASIC TESTS
```
dhermes: Comment not needed.

tests/contrib/test_gce.py
```diff
@@ -14,102 +14,195 @@
...
-        http = mock.MagicMock()
-        http.request = mock.MagicMock(
-            return_value=(mock.Mock(status=http_client.OK), return_val))
+    @mock.patch(METADATA_SERVER, return_value=A_SERVICE_ACCOUNT)
+    def test_construtor_with_missing_scopes(self, get_metadata):
+        scope = 'http://wwww.nobodyhasthisscope.com'
+        error = None
+        try:
+            AppAssertionCredentials(scope=scope)
+        except ValueError as e:
```
dhermes: Ditto on a longer variable name:

> Names to Avoid: ...single character names except for counters or iterators

tests/contrib/test_gce.py
```diff
@@ -14,102 +14,195 @@
...
-        scope = 'http://example.com/a http://example.com/b'
-        scopes = scope.split()
-        credentials = AppAssertionCredentials(scope=scopes, foo='bar')
-        self.assertEqual(credentials.scope, scope)
-        self.assertEqual(credentials.kwargs, {'foo': 'bar'})
-        self.assertEqual(credentials.assertion_type, None)
+    @mock.patch(METADATA_SERVER, return_value=A_SERVICE_ACCOUNT)
+    def test_constructor_with_matching_scopes(self, get_metadata, warn_mock):
+        scope = 'https://www.googleapis.com/auth/cloud-platform'
+        AppAssertionCredentials(scope=scope)
```
dhermes: You should actually make assertions about the object created as with done in the deleted `test_constructor_with_scopes`.

tests/contrib/test_gce.py
```diff
@@ -14,102 +14,195 @@
...
-            return_value=(mock.Mock(status=http_client.OK), return_val))
+    @mock.patch(METADATA_SERVER, return_value=A_SERVICE_ACCOUNT)
+    def test_construtor_with_missing_scopes(self, get_metadata):
+        scope = 'http://wwww.nobodyhasthisscope.com'
+        error = None
+        try:
+            AppAssertionCredentials(scope=scope)
+        except ValueError as e:
+            error = e
+        self.assertIsNotNone(error)
```
dhermes: Use `self.assertRaises`:

``` python
with self.assertRaises(ValueError):
    AppAssertionCredentials(scope=scope)
```

tests/contrib/test_gce.py
dhermes: NVM This is moot since you wanted to use `assertRaises`

tests/contrib/test_gce.py
```diff
@@ -14,102 +14,195 @@
...
 
+    @mock.patch(METADATA_SERVER, return_value=A_SERVICE_ACCOUNT)
+    def test_default_service_account_info(self, get_metadata):
+        credentials = AppAssertionCredentials()
+        self.assertEquals(
+            credentials._service_account_info,
+            DEFAULT_SERVICE_ACCOUNT
+        )
+        self.assertEquals(credentials.scopes, A_SERVICE_ACCOUNT['scopes'])
```
dhermes: Might be clearer to say `get_metadata.return_value['scopes']` instead of `A_SERVICE_ACCOUNT['scopes']`.

tests/contrib/test_gce.py
```diff
@@ -14,102 +14,195 @@
...
+        try:
+            AppAssertionCredentials(scope=scope)
+        except ValueError as e:
+            error = e
+        self.assertIsNotNone(error)
 
+    @mock.patch(METADATA_SERVER, return_value=A_SERVICE_ACCOUNT)
+    def test_default_service_account_info(self, get_metadata):
+        credentials = AppAssertionCredentials()
+        self.assertEquals(
```
dhermes: Just noticed the `assertEquals` instead of `assertEqual`. The plural is moving towards deprecation, so prefer `assertEqual`:

> The plurals are undocumented. Keep them that way to discourage use.

tests/contrib/test_gce.py
```diff
@@ -14,102 +14,195 @@
...
 
-    def test_refresh_success(self):
-        self._refresh_success_helper(bytes_response=False)
+    @mock.patch(METADATA_SERVER, return_value=A_SERVICE_ACCOUNT)
+    def test_serialize_deserialize(self, get_metadata):
+        credentials = AppAssertionCredentials()
+        credentials_from_json = Credentials.new_from_json(
+            credentials.to_json()
+        )
+        self.assertTrue(
```
dhermes: `assertEqual`?

tests/contrib/test_gce.py
```diff
@@ -14,102 +14,195 @@
...
-        self.assertEquals(access_token, credentials.access_token)
 
-        base_metadata_uri = (
-            'http://metadata.google.internal/computeMetadata/v1/instance/'
-            'service-accounts/default/token')
-        http.request.assert_called_once_with(
-            base_metadata_uri, headers={'Metadata-Flavor': 'Google'})
+        with mock.patch(METADATA_SERVER, return_value=ONE_SECOND_TOKEN):
+            credentials.get_access_token()
+            time.sleep(2)
```
dhermes: No no never `time.sleep`. Just `mock` out the calls to `datetime.datetime.utcnow()`.

dhermes: This will also make it so that you don't need a "one second" token. You can use any value you like, just make it match your mock.

tests/contrib/test_gce.py
```diff
@@ -14,102 +14,195 @@
...
+
+    @mock.patch('warnings.warn')
+    @mock.patch(METADATA_SERVER, return_value=A_SERVICE_ACCOUNT)
+    def test_create_scoped_valid(self, get_metadata, warn_mock):
+        scope = 'https://www.googleapis.com/auth/cloud-platform'
+        credentials = AppAssertionCredentials()
+        credentials.create_scoped(scope)
+        warn_mock.assert_called_once_with(_SCOPES_WARNING)
+
+    # ERROR TESTS
```
dhermes: Axe this comment. If you think the tests should be split up, just make a new `TestCase`.

tests/contrib/test_gce.py
```diff
@@ -121,132 +214,55 @@ def test_refresh_failure_404(self):
...
-        import os
-        ORIGINAL_ISDIR = os.path.isdir
+        error = None
         try:
-            os.path.isdir = lambda path: True
+            credentials._retrieve_scopes(http.request)
+        except ValueError as e:
+            error = e
+
+        self.assertIsNotNone(error)
```
dhermes: Ditto here on `assertRaises`

tests/contrib/test_gce.py
```diff
@@ -121,132 +214,55 @@ def test_refresh_failure_404(self):
...
+            credentials._service_account_info['email'],
+            'default'
+        )
+        self.assertFalse(
+            'scopes' in credentials._service_account_info
+        )
+
+    @mock.patch(METADATA_SERVER, return_value=A_SERVICE_ACCOUNT)
+    def test_save_to_well_known_file(self, get_metadata):
+        with tempfile.NamedTemporaryFile() as f:
```
dhermes: Please use a better name than `f`.

tests/contrib/test_gce.py
```diff
@@ -121,132 +214,55 @@ def test_refresh_failure_404(self):
...
             credentials = AppAssertionCredentials()
-            self.assertRaises(NotImplementedError, save_to_well_known_file,
-                              credentials)
-        finally:
-            os.path.isdir = ORIGINAL_ISDIR
+            save_to_well_known_file(credentials, well_known_file=f.name)
```
dhermes: There are no assertions in this test.

tests/contrib/test_gce.py
```diff
@@ -254,11 +270,27 @@ def test_failure(self):
...
         http_request.return_value = (response, content)
-        result = _get_service_account_email(http_request)
+        error = None
+        try:
+            _get_metadata(http_request)
+        except ValueError as e:
+            error = e
 
-        self.assertEqual(result, (response, content))
+        self.assertIsNotNone(error)
```
dhermes: `self.assertRaises`

oauth2client/contrib/gce.py
elibixby: Unless configured otherwise, the default service account for a compute engine instance will be the "compute engine service account" which is shared by all GCE instances within a project. I can remove that or clarify if you think it's confusing.

oauth2client/contrib/gce.py
elibixby: I can change it back, I was just being extra strict.

oauth2client/contrib/gce.py
elibixby: Sorry, what do you mean by "jump the gun"? 

This was meant to be an allusion to how serialization/deserialization is meant to work. Can remove if it's confusing.

oauth2client/contrib/gce.py
elibixby: Yeah, I'd like to fail faster when a user instantiates these credentials with scopes it cannot possibly have. e.g. a user calls `credentials.create_scoped(['a','b'])` and these credentials only have scope `'c'`. Calls which require scopes `'a'` and `'b'` will fail _when they are made_ but we know that these credentials don't match user expectations when they are instantiated so we should fail then. 

In the case that the user tries to create these credentials with scopes the _do_ have we can just warn them that they shouldn't be specifying scopes, but the credentials will work as expected.

oauth2client/contrib/gce.py
elibixby: I added this comment to reassure reviewers that `service_account_email` and `scopes` would never result in `KeyError`s. 

I can remove it from the docstring and put it in a code comment though.

oauth2client/contrib/gce.py
elibixby: Nope, I'll add `...`

oauth2client/contrib/gce.py
elibixby, line 61: So this is a bigger design question: Lazily populate values, or populate them in the constructor. 

I went with lazy population because that is what the original author did, and also, there is a chance that a user will never use the some properties, resulting in fewer overall calls. 

I have cached the values, as well as consolidating calls (e.g. scopes and email are fetched together)  as much as possible. If you see other opportunities to cache them please let me know.

oauth2client/contrib/gce.py
elibixby, line 202: [https://github.com/google/oauth2client/blob/master/oauth2client/client.py#L555](https://github.com/google/oauth2client/blob/3ca2ca7f39b7ac60d449ca1ce6062e46b0dd899d/oauth2client/client.py#L555) Looks like here:

Part of it was it didn't feel appropriate to expect a warning on literally every test that makes a credential. Maybe I'll warn only on attempts to set with a non-empty value? That seems like a good solution I'll try it.

oauth2client/contrib/gce.py
elibixby: fair enough.

oauth2client/contrib/gce.py
dhermes: Yeah it's a sentence fragment and hard to parse. Just update it to be clearer.

oauth2client/contrib/gce.py
dhermes: Definitely change it back, but maybe we can just axe it all together?

oauth2client/contrib/gce.py
dhermes: Sorry for my vague comment. I'd say just update to this value to something more descriptive. How about:

```
            service_account_info:
                (Optional) Dictionary containing information about a
                service account (typically deserialized from JSON).
                If passed in, will be used as ``service_account_info``
                property. Otherwise, the Compute Engine metadata server
                will be used to determine this value.
```

oauth2client/contrib/gce.py
elibixby: Value error is returned on a bad status from the `httplib2.Response` object.

elibixby: Actually, going to move this all up to the _get_access_token method, so it doesn't cover up deserialization errors.

oauth2client/contrib/gce.py
dhermes: Yeah I like ~~to~~ the goal. But you should get a message other than `_SCOPES_WARNING`?

oauth2client/contrib/gce.py
dhermes: Just start with `This property is...` and drop the reference to the non-public method. Also don't forget about the empty line.

oauth2client/contrib/gce.py
dhermes, line 202: I'm unclear on what you're trying (or what I've suggested?)

oauth2client/contrib/gce.py
dhermes, line 61: I'd guess the original author haphazardly threw this together, so basing on their impl. may not be useful. Also there was only one thing retrieved (the access token) and now there are many.

Can you throw together a list of all metadata URLs retrieved and all the values used from each?

oauth2client/contrib/gce.py
elibixby, line 202: Ok, lemme try to lay this out more clearly:

My desired behavior:
Any time a `scope` is specified (either via `create_scoped_required`, or with a `scope` kwarg, or by assigning to the `scope` property), I want `AppAssertionCredentials` to fail fast, by checking these scopes against the immutable scopes of the credentials, and warning if they are a subset, and erroring if they have are not. This way users don't have to wait to make a call to know they won't work (and we can provide a better error message).

There are lots of weird inconsistencies in how scopes are handled in this library:
- The constructor for `Credentials` takes a `scopes` argument which is assigned to the property `self.scopes`
- Neither `GoogleCredentials` nor `AssertionCredentials` pass unused kwargs up to their parent's constructor. So there is no way for me to get at this arg. 
- This `gce.AppAssertionCredentials` defines (apparently for legacy reasons) a `scope` argument. it's not clear to me why this is different from the `scopes` argument of `Credentials`, or why it needs to be provided (given that the parent class uses `scopes`)

Overall this class hierarchy has me quite confused, and I'm not sure how to implement the behavior I want, and what exactly I'm expected to support...

oauth2client/contrib/gce.py
elibixby: I'd be fine with that.

oauth2client/contrib/gce.py
elibixby: SGTM, I also added a line about it overriding the `service_account_email`

oauth2client/contrib/gce.py
elibixby, line 61: So there are just 3:
- `_METADATA_ROOT + 'instance/service-accounts/<email>/?recursive=true'`
  Which returns basically a `service_account_info` object of which we use `scopes` and `email`
- `_METADATA_ROOT + 'project/project-id'` which returns the project-id as a string and will only be used for blob signing (or if someone is curious for some reason)
- `_METADATA_ROOT + 'instance/service-accounts/<email>/token'` which returns the token, which is only the two fields we use.

All the recursive nonsense is basically just to bundle the `scopes` and `email` call together...

dhermes, line 61: Why is recursive nonsense? The project ID as numeric or the project name? What do you mean by "which is only the two fields we use"?

oauth2client/contrib/gce.py
dhermes, line 202: The class hierarchy is what you get when a large disparate group of people implement just the thing they need with no real owner guiding the contributions. We're hoping to slowly mitigate this and bring the library to being higher quality.

I'm just saying it'd be nice to avoid having this setter do nothing. Maybe you should warn if `value` isn't empty string or empty list?

tests/contrib/test_gce.py
elibixby: That doesn't work, because `datetime` is immutable. Mocking `datetime.datetime.now()` appears to be non-trivial, there are some libraries that do it well. Thoughts on making one of them a test dependency?

dhermes: We mock `now()` in `oauth2client`. Do a `git grep` for `_UTCNOW` to see some examples.

elibixby: I find that reduction in readability for the sake of tests without `sleep()` _very_ distasteful. But w/e

oauth2client/contrib/gce.py
elibixby, line 61: Just because it's inconsistent and it requires the trailing `/` and gave me some frustrating bugs. No actual qualms.
`access_token` and `expires_in` are the only fields in the response from the `.../token` endpoint, and we use both of them.

tests/contrib/test_gce.py
dhermes: The philosophy is that unit tests that take a long time to run make new contributions to a project difficult (and make iterating on the codebase difficult). So a 2s sleep is a large penalty to bear when the entire test suite should take less than 0.5s to run.

oauth2client/contrib/gce.py
elibixby, line 202: I added a `_check_scopes_and_notify` function that will be called both from the setter and the constructor. ¯_(ツ)_/¯

elibixby:
@dhermes I think all of your comments have been addressed. Lemme know if there's any other changes that need to be made.

oauth2client/contrib/gce.py
```diff
@@ -22,11 +22,10 @@
 import warnings
 
 import httplib2
+import datetime
```
dhermes: Import order should be change here. Imports should be ordered first in sections (std lib then 3rd party then local imports) and then alphabetical within sections. (~~i.e.~~ e.g. `datetime` is in the standard library but `httplib2` is 3rd party)

elibixby: Fixed (I think)

oauth2client/contrib/gce.py
```diff
@@ -20,13 +20,12 @@
 import json
 import logging
 import warnings
+import datetime
```
dhermes: Missed the alphabetical order part.

elibixby: aaaaaaarg, sorry about that.

elibixby:
@dhermes @nathanielmanistaatgoogle 
Ready for another look I think?

oauth2client/contrib/gce.py
```diff
@@ -16,60 +16,113 @@
...
+_METADATA_ROOT = 'http://metadata.google.internal/computeMetadata/v1/'
 _SCOPES_WARNING = """\
 You have requested explicit scopes to be used with a GCE service account.
-Using this argument will have no effect on the actual scopes for tokens
-requested. These scopes are set at VM instance creation time and
-can't be overridden in the request.
+These scopes *are* present on the credentials; but this argument will have
+no effect on the actual scopes for tokens requested. The credentials scopes
+are set at VM instance creation time and can't be overridden in the request.
+To learn more go to: https://cloud.google.com/compute/docs/authentication
```
nathanielmanistaatgoogle: Drop the colon and add a period at the end.

oauth2client/contrib/gce.py
```diff
@@ -16,60 +16,113 @@
...
-                  'instance/service-accounts/default/')
-META = _METADATA_ROOT + 'token'
-_DEFAULT_EMAIL_METADATA = _METADATA_ROOT + 'email'
+_METADATA_ROOT = 'http://metadata.google.internal/computeMetadata/v1/'
 _SCOPES_WARNING = """\
 You have requested explicit scopes to be used with a GCE service account.
-Using this argument will have no effect on the actual scopes for tokens
-requested. These scopes are set at VM instance creation time and
-can't be overridden in the request.
+These scopes *are* present on the credentials; but this argument will have
```
nathanielmanistaatgoogle: "These scopes" is a little ambiguous because the previous sentence talks about scopes being requested (so they don't yet exist) rather than being things that already exist.

Also "the credentials" is ambiguous; do you mean "a service account credentials" in the previous sentence?

oauth2client/contrib/gce.py
```diff
@@ -16,60 +16,113 @@
...
+_NOW = datetime.datetime.now
+
 
+class MetadataServerHttpError(Exception):
+    """Error for Http failures originating from the Metadata Server"""
 
-def _get_service_account_email(http_request=None):
-    """Get the GCE service account email from the current environment.
+
+def _get_metadata(http_request=None, path=None, recursive=True):
```
nathanielmanistaatgoogle, line 61: Why doesn't `recursive` appear in the `Args:` section below?

oauth2client/contrib/gce.py
```diff
@@ -16,60 +16,113 @@
...
     Returns:
-        tuple, A pair where the first entry is an optional response (from a
-        failed request) and the second is service account email found (as
-        a string).
+        A deserialized JSON object representing the data returned
+        from the metadata server
     """
-    if http_request is None:
+
+    if not path:
```
nathanielmanistaatgoogle: Do you actually mean to trigger on any falsey value, or should this be an explicit `is None`?

oauth2client/contrib/gce.py
```diff
@@ -16,60 +16,113 @@
...
+def _get_access_token(http_request, email):
+    """Get an access token for the specified email from the Metadata Server.
+
+    Args:
+        http_request: an httplib2.Http().request object or equivalent
+            with which to make the call to the metadata server
+        email: The service account email to request an access token with
+
+    Returns:
+        A tuple (accessToken, token expiry)
```
nathanielmanistaatgoogle: Fully describe the type of the return value. What is the type of the returned access token? What is the type of the returned token expiry? Why is one written thisWay and the other written this way? Also add trailing punctuation.

oauth2client/contrib/gce.py
```diff
@@ -85,32 +138,128 @@ class AppAssertionCredentials(AssertionCredentials):
     information to generate and refresh its own access tokens.
     """
 
-    @util.positional(2)
-    def __init__(self, scope='', **kwargs):
+    def __init__(self,
+                 scope=None,
+                 service_account_email='default',
```
nathanielmanistaatgoogle: Keyword parameter default values that are not `None` scare me. Why is this string, which doesn't even look like an email address, the right default value for this parameter?

oauth2client/contrib/gce.py
```diff
@@ -85,32 +138,128 @@ class AppAssertionCredentials(AssertionCredentials):
...
+                provided and will override the ``service_account_email``
+                argument.
         """
-        if scope:
-            warnings.warn(_SCOPES_WARNING)
-        # This is just provided for backwards compatibility, but is not
-        # used by this class.
-        self.scope = util.scopes_to_string(scope)
-        self.kwargs = kwargs
+
```
nathanielmanistaatgoogle, line 177: Maintain the codebase-wide no-blank-line-between-behavior-doc-string-and-behavior-body convention, please.

oauth2client/contrib/gce.py
```diff
@@ -85,32 +138,128 @@ class AppAssertionCredentials(AssertionCredentials):
...
+        self._service_account_info = service_account_info or {
```
nathanielmanistaatgoogle: Never use `<A> or <B>` to mean `<A> if <A> else <B>`. The default value of the `service_account_info` parameter is `None`, so if the caller passes an empty dictionary they must certainly have some specific intention for it. Do you mean to override their input in such cases?

oauth2client/contrib/gce.py
```diff
@@ -85,32 +138,128 @@ class AppAssertionCredentials(AssertionCredentials):
...
+
+        # This function call must be made because AssertionCredentials
+        # will not pass the scopes kwarg to parent class
+        self._check_scopes_and_notify(scope)
 
         # Assertion type is no longer used, but still in the
         # parent class signature.
-        super(AppAssertionCredentials, self).__init__(None)
-        self._service_account_email = None
+        super(AppAssertionCredentials, self).__init__(
```
nathanielmanistaatgoogle: In a constructor containing a `super` call, the `super` call should always be the first statement of the constructor unless there's a clearly documented reason for it having to be delayed until after something else happens.

oauth2client/contrib/gce.py
```diff
@@ -85,32 +138,128 @@ class AppAssertionCredentials(AssertionCredentials):
...
+                ],
+                http_request=http_request
+            )
+            self._partial = False
+        return self._service_account_info
+
+    def _retrieve_scopes(self, http_request):
+        return self._get_service_account_info(
+            http_request=http_request
+        )['scopes']
```
nathanielmanistaatgoogle: I don't understand the line-wrap here.

Can the whole expression fit on one line?

oauth2client/contrib/gce.py
```diff
@@ -125,32 +274,16 @@ def _refresh(self, http_request):
...
-    @property
-    def serialization_data(self):
-        raise NotImplementedError(
-            'Cannot serialize credentials for GCE service accounts.')
-
-    def create_scoped_required(self):
-        return False
+        self.access_token, self.token_expiry = _get_access_token(
+            http_request,
+            self._service_account_info['email']
```
nathanielmanistaatgoogle: Condense these three lines; this should be formatted as

```
self.access_token, self.token_expiry = _get_access_token(
    http_request, self._service_account_info['email'])
```

.

tests/contrib/test_gce.py
```diff
@@ -13,240 +13,252 @@
 # limitations under the License.
 
 """Unit tests for oauth2client.contrib.gce."""
```
nathanielmanistaatgoogle, line 15: One blank line between module doc strings and import statements, please.

tests/contrib/test_gce.py
```diff
@@ -13,240 +13,252 @@
...
-
+from datetime import datetime
```
nathanielmanistaatgoogle: [Import only modules, not classes, functions, or other code elements of modules.](https://google.github.io/styleguide/pyguide.html?showone=Imports#Imports)

nathanielmanistaatgoogle:
Be prepared at the end of the review to squash commits.

oauth2client/contrib/gce.py
elibixby: The metadata server provides a "default" service account so that users don't have to know the email of the default service account in order to fetch access tokens. To use the default service account, you supply the string 'default' as the `service_account_email` to the metadata server. For backwards compatibility, (where specifying a specific service account is not possible, and this class always uses the default service account), I need to be able to leave this out. 

I could do

```
def __init__(..., service_account_email=None):
    if service_account_email is None:
        service_account_email='default'
```

But I don't see any advantage to that. Do you feel the doc string needs to be clearer? Do you have a suggestion for prose?

oauth2client/contrib/gce.py
elibixby: The argument `service_account_info` is only provided for serialization and deserialization, and not intended for end user use. I could remove it from the constructor and do,

```
def from_json(cls, json_data):
    creds = cls()
    creds._service_account_info = json_data['service_account_info']
```

If you prefer

oauth2client/contrib/gce.py
elibixby: Another option is leave `service_account_email` as `None` and `service_account_info` as `{}`, and change `_refresh` to look like

```
def _refresh(self, http_request):
    self.access_token, self.token_expiry = _get_access_token(
        http_request,
        sef._service_account_info.get('email', 'default')
)
```

Which maybe I like better? Can't decide.

elibixby: Also would need to change `_get_service_account_info` to use `.get('email', 'default')`

elibixby:
https://github.com/blog/2141-squash-your-commits 

Is this option available to you now? Or do I still have to manually squash?

oauth2client/contrib/gce.py
nathanielmanistaatgoogle: Yes. A public class that is itself "intended for end user use" should have no construction parameters that are "not intended for end user use".

nathanielmanistaatgoogle:
A repository setting to require all pull requests to be one commit each probably is available to us, but I'm not sure that we'd want to enable it for all changes. Some are still appropriate to be done in multiple commits.

oauth2client/contrib/gce.py
```diff
@@ -16,60 +16,117 @@
 
 Utilities for making it easier to use OAuth 2.0 on Google Compute Engine.
 """
-
```
nathanielmanistaatgoogle, line 19: Maintain one-blank-line-between-module-doc-string-and-import-statements, please.

oauth2client/contrib/gce.py
```diff
@@ -16,60 +16,117 @@
...
-META = _METADATA_ROOT + 'token'
-_DEFAULT_EMAIL_METADATA = _METADATA_ROOT + 'email'
+_METADATA_ROOT = 'http://metadata.google.internal/computeMetadata/v1/'
 _SCOPES_WARNING = """\
-You have requested explicit scopes to be used with a GCE service account.
-Using this argument will have no effect on the actual scopes for tokens
-requested. These scopes are set at VM instance creation time and
-can't be overridden in the request.
+You have requested explicit scopes to be used with a GCE service account, and
+these scopes *are* present on the credentials. However, this request will have
```
nathanielmanistaatgoogle: "However, your request"? This text reads awkwardly in the way it uses the word "request" to mean both "thing the user asked to happen" and "a software object representing a yet-to-be-transmitted HTTP request".

Also what is "the credentials"? Should the previous line read "a GCE service account credentials" rather than "a GCE service account"?

oauth2client/contrib/gce.py
```diff
@@ -16,60 +16,117 @@
...
+                'service-accounts',
+                email,
+                'token'
+            ],
+            recursive=False
+        )
+    except MetadataServerHttpError as failed_fetch:
+        raise HttpAccessTokenRefreshError(str(failed_fetch))
+
+    token_expiry = _NOW() + datetime.timedelta(
```
nathanielmanistaatgoogle, line 127: This line wrapping looks really, really weird. Does it really not fit all on one line?

oauth2client/contrib/gce.py
nathanielmanistaatgoogle: Definitely avoid having a non-`None` keyword argument default value, even if it does make the class' internals slightly more complicated.

oauth2client/contrib/gce.py
```diff
@@ -85,32 +142,126 @@ class AppAssertionCredentials(AssertionCredentials):
...
-        self.scope = util.scopes_to_string(scope)
-        self.kwargs = kwargs
-
         # Assertion type is no longer used, but still in the
         # parent class signature.
-        super(AppAssertionCredentials, self).__init__(None)
-        self._service_account_email = None
+        super(AppAssertionCredentials, self).__init__(
+            None,
+            scopes=scope,
```
nathanielmanistaatgoogle: This should be laid out as

```
super(AppAssertionCredentials, self).__init__(
    None, scopes=scope, **unused_kwargs)
```

. Maybe other codebases do a one-parameter-per-line thing (I have tender memories of Java, for example), but mostly not Python and not at all this codebase.

Now, two other questions:
(1) The `scopes=scope` part suggests that one of those two things is misnamed.
(2) The use of `unused_kwargs` suggests that it is misnamed.

oauth2client/contrib/gce.py
```diff
@@ -85,32 +142,126 @@ class AppAssertionCredentials(AssertionCredentials):
...
+                 **unused_kwargs):
         """Constructor for AppAssertionCredentials
 
         Args:
-            scope: string or iterable of strings, scope(s) of the credentials
-                   being requested. Using this argument will have no effect on
-                   the actual scopes for tokens requested. These scopes are
-                   set at VM instance creation time and won't change.
+            scope:
+                string or iterable of strings, scope(s) of the credentials
```
nathanielmanistaatgoogle, line 158: Can we avoid this? One of Python's greatest mistakes is the conflation of strings and iterables of strings; is there any reason we should perpetuate it here?

oauth2client/contrib/gce.py
```diff
@@ -85,32 +142,126 @@ class AppAssertionCredentials(AssertionCredentials):
...
+    @property
+    def service_account_email(self):
+        return self.service_account_info['email']
+
+    @property
+    def project_id(self):
+        if not self._project_id:
+            self._project_id = _get_metadata(
+                path=['project', 'project-id'],
+                recursive=False
```
nathanielmanistaatgoogle: More weird function call arguments formatting.

oauth2client/contrib/gce.py
elibixby, line 158: I'm trying not to break things. The old argument accepted a string or iterable. Given that scope is a totally non-useful arg, I'd be happiest to remove it completely, but it could unnecessarily break user code. Granted it's entire existence I think there has been a warning that fires if you specify a scope, so maybe we can call that sufficient warning to remove it.

oauth2client/contrib/gce.py
elibixby: `scope` is definitely misnamed. I'd prefer to remove it all together. See below.

It's my habit to call `unused_kwargs` all those kwargs which are passed to the parent constructor. 

However. `AssertionCredentials` doesn't do that. Which is actually a big mistake because it prevents me from overriding the scopes usage in the parent `GoogleCredentials` class. So these are actually even more unused than my normal usage of `unused_kwargs`. I can change back to  `kwargs` if it bothers you though.

oauth2client/contrib/gce.py
nathanielmanistaatgoogle, line 158: Oh right...

My apologies.

oauth2client/contrib/gce.py
nathanielmanistaatgoogle: Yep. The `unused_` prefix is reserved for fields that are _syntactically_ unused, not just _semantically_ unused.

elibixby:
@nathanielmanistaatgoogle Fixes pushed

oauth2client/contrib/gce.py
```diff
@@ -17,59 +17,114 @@
...
+from oauth2client._helpers import _from_bytes
 
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 
 logger = logging.getLogger(__name__)
 
 # URI Template for the endpoint that returns access_tokens.
-_METADATA_ROOT = ('http://metadata.google.internal/computeMetadata/v1/'
-                  'instance/service-accounts/default/')
-META = _METADATA_ROOT + 'token'
```
nathanielmanistaatgoogle, line 41: Since you're aspiring to make a backwards-compatible change, this public constant should not be removed.

Right?

oauth2client/contrib/gce.py
```diff
@@ -17,59 +17,114 @@
...
-can't be overridden in the request.
+You have requested explicit scopes to be used with a GCE service account
+credentials, and these scopes *are* present on the credentials.
+However, setting scopes on the GCE service account credentials has no effect
+on the actual scopes for tokens requested. The credentials scopes
+are set at VM instance creation time and can't be overridden in the request.
+To learn more go to https://cloud.google.com/compute/docs/authentication .
+"""
+_SCOPES_ERROR = """\
+You have requested explicit scopes to be used with a GCE service account
```
nathanielmanistaatgoogle: Same concerns that shaped the warning string above should shape this error string, right? "request" having two meanings and so on?

oauth2client/contrib/gce.py
```diff
@@ -125,32 +257,15 @@ def _refresh(self, http_request):
...
-    def create_scoped_required(self):
-        return False
+        self.access_token, self.token_expiry = _get_access_token(
+            http_request,
+            self._service_account_info.get('email', 'default'))
 
     def create_scoped(self, scopes):
-        return AppAssertionCredentials(scopes, **self.kwargs)
+        # Trigger warning or error based on scopes
+        self.scopes = scopes
```
nathanielmanistaatgoogle: Avoid calling one part of the public API of an object from within the implementation of another part of the public API of the same object.

Why is `self.scopes = scopes` better here than `self._check_scopes_and_notify(scopes)`? If you chose the latter the purpose would be clear and you could drop the implementation comment, right?

oauth2client/contrib/gce.py
```diff
@@ -85,32 +140,109 @@ class AppAssertionCredentials(AssertionCredentials):
...
+            self._service_account_info['email'] = service_account_email
+
+        self._project_id = None
+        self._partial = True
+
+        self.kwargs = kwargs
+
+        # This function call must be made because AssertionCredentials
+        # will not pass the scopes kwarg to parent class
+        self._check_scopes_and_notify(scope)
```
nathanielmanistaatgoogle, line 186: Calling an instance method from the constructor of the same instance is inviting trouble. In this case the instance method being called itself calls a public instance method (`has_scopes`) inherited by this class, so we're adding "and guest" to the invitation.

How avoidable are these issues?

oauth2client/contrib/gce.py
```diff
@@ -85,32 +140,109 @@ class AppAssertionCredentials(AssertionCredentials):
...
-        super(AppAssertionCredentials, self).__init__(None)
-        self._service_account_email = None
+        super(AppAssertionCredentials, self).__init__(
+            None, scopes=scope, **kwargs)
 
-    @classmethod
-    def from_json(cls, json_data):
-        data = json.loads(_from_bytes(json_data))
-        return AppAssertionCredentials(data['scope'])
+        self._service_account_info = {}
```
nathanielmanistaatgoogle: Always replace

```
<name> = <value>
if <condition>:
  <name> = <other value>
```

with

```
if <condition>:
  <name> = <other value>
else:
  <name> = <value>
```

. Now obviously this case escapes the strict letter of that rule because instead of `<name> = <other value>` you have `<mutation of value assigned to <name>>`, but the principle still applies. This should be

```
if service_account_email is None:
  self._service_account_info = {}
else:
  self._service_account_info = {'email': service_account_email}
```

.

oauth2client/contrib/gce.py
elibixby, line 186: I do think it's important to let users know when their credentials aren't behaving as they would expect from credentials with explicit scopes. Given that, I think the choice here is between code duplication, and calling object code from within the constructor. Pick your poison.

elibixby, line 186: Another things to note is that when we release the next breaking version we can simply remove the `scope` argument to the constructor, and then remove this awkward code. So it's not a key implementation detail so much as a hack for verbose backwards compat.

oauth2client/contrib/gce.py
elibixby: Fair point, `requested` -> `specified` in the first usage. Hopefully that's better prose.

elibixby:
@nathanielmanistaatgoogle More fixes.

elibixby:
@nathanielmanistaatgoogle @dhermes Bump.

elibixby:
After discussion with @jonparrott I'm going to break this into 3 smaller PRs:

* add a `metadata` module that takes care of caching paths

* Move `gce.AppAssertionCredentials` to use the metadata module

* Add `credentials.scopes` and support for custom emails

Issue #393, opened 2016-02-02
Unable to switch abstract classes to use of abc module


nathanielmanistaatgoogle:
Since 2.6 the right way to write abstract classes in Python is with use of the `abc` standard library module. I took a stab today at converting `client.Credentials`, `client.Storage`, and `client.AssertionCredentials` to using `abc` and a bunch of tests failed with abstract-class-instantiation errors.

Which means that currently our tests instantiate nonsense partially-abstract objects and are totally cool with that. That is, in fact, not actually cool, and should be fixed.

theacodes:
Thank you for creating this issue, however, this project is deprecatedand we will only be addressing critical security issues. You can read moreabout this deprecation [here](http://google-auth.readthedocs.io/en/latest/oauth2client-deprecation.html).

If you need support or help using this library, we recommend that you ask yourquestion on StackOverflow.

If you still think this issue is relevant and should be addressed, pleasecomment and let us know!

PR #477, opened 2016-04-06
Adding test coverage for client.py module.

dhermes:
- Replacing `_abstract()` with directly raising `NotImplementedError`
- Testing all abstract code paths
- Adding no-cover for a failed import
- Completely testing the base `Credentials` class

---

This is the beginning of a sprint to get to 100% code coverage (#212). I plan to focus on `client.py` and send small-ish PRs (to make review easier) and then just use each PR as a diffbase.

If too many diffbases makes reviews hard, I can just make the changes from `HEAD` and then deal with merge conflicts on my own.

/cc @waprin @nathanielmanistaatgoogle

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -55,7 +55,7 @@ try:
     HAS_CRYPTO = True
     if crypt.OpenSSLVerifier is not None:
         HAS_OPENSSL = True
-except ImportError:
+except ImportError:  # pragma: NO COVER
     pass
 
 
@@ -179,10 +179,6 @@ class CryptoUnavailableError(Error, NotImplementedError):
     """Raised when a crypto library is required, but none is available."""
 
 
-def _abstract():
-    raise NotImplementedError('You need to override this function')
-
-
 class MemoryCache(object):
     """httplib2 Cache implementation which only caches locally."""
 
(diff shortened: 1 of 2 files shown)


oauth2client/client.py
```diff
@@ -229,7 +225,7 @@ def authorize(self, http):
             http: httplib2.Http, an http object to be used to make the refresh
                   request.
         """
-        _abstract()
+        raise NotImplementedError
```
nathanielmanistaatgoogle, line 228: I prefer `()` at the end of `raise NotImplementedError`, but if you feel strongly it's also fine to leave these as they are.

nathanielmanistaatgoogle:
Looks great; merge if you like?

oauth2client/client.py
theacodes, line 228: Same.

jonparrott:
LGTM as well.

oauth2client/client.py
dhermes, line 228: I prefer this way (I always feel special because I know Python will create the instance for me).

nathanielmanistaatgoogle:
Also curious about this is that it is completely orthogonal to [issue 393](https://github.com/google/oauth2client/issues/393).

PR #478, opened 2016-04-06
Move platform-specific openers to their own modules.

theacodes:


oauth2client/contrib/locked_file.py
```diff
@@ -347,10 +189,15 @@ def __init__(self, filename, mode, fallback_mode, use_native_locking=True):
         """
         opener = None
         if not opener and use_native_locking:
-            if _Win32Opener:
+            try:
+                from oauth2client.contrib._win32_opener import _Win32Opener
```
dhermes, line 193: Please move all imports to the top-level and then just do the conditional checks here based on import success or failure.

dhermes:
@jonparrott I noticed you made some test changes. Did you make any changes (other than de-dents) to the code moved out into the other modules?

oauth2client/contrib/locked_file.py
theacodes, line 193: So this leads to fun circular import problems. Any recommendations?

jonparrott:
> I noticed you made some test changes.

Just fixing the mock patching.

oauth2client/contrib/locked_file.py
dhermes, line 193: Can you enumerate the cycles?

theacodes, line 193: Basically the opener classes depend on two exceptions and one helper method defined in `locked_file.py`.

dhermes, line 193: What does the helper do? How detailed is the implementation of the exceptions?

theacodes, line 193: The exceptions are trivial. The helper just makes sure that the file isn't a link. Both opener classes also subclass `_Opener` defined in `locked_file`.

dhermes, line 193: Ahhh that is a problem. Thoughts on just defining mix-ins in the platform specific modules?

theacodes, line 193: Not sure what you mean.

dhermes, line 193: Sorry for being vague. I propose this. Instead of defining

``` python
class _FcntlOpener(_Opener):
    ...
```

in the `_fcntl_opener.py`, just define

``` python
class _FcntlMixin(object):
    ...
```

and then in `locked_file.py` you could define

``` python
class _FcntlOpener(_Opener, _FcntlMixin):
    ...
```

to get both feature sets in.

dhermes, line 193: And if the import failed you could just set `_FcntlMixin = object`, which won't do much but will allow inheritance.

dhermes, line 193: I just took a look at the implementations, we could just make them ~~methods~~ standalone functions?

theacodes, line 193: Still not seeing how that fixes the circular imports for the exceptions.

dhermes, line 193: It doesn't, though do-nothing exception classes should probably be excised.

Let's punt on all of this for this PR but definitely totally promise to follow up and fix it after the fact?

theacodes, line 193: Sure, I'm cool with that. Hopefully we can just remove this altogether. Merge if you're okay with it?

dhermes, line 193: Let's have a plan in place first (i.e. an issue with an owner that will "definitely totally promise to follow up")

theacodes, line 193: Sure: how about I file a bug to resolve circular dependencies or follow through with removing this altogether?

dhermes, line 193: SGTM. Ping me once it's filed

theacodes, line 193: #487 filed.

PR #479, opened 2016-04-06
Improving test branch coverage in client.py.

dhermes:
Testing un-traversed branches in
- `OAuth2Credentials._expires_in()`
- `OAuth2Credentials.get_access_token()`
- `OAuth2Credentials._do_refresh_request()`
- `HAS_OPENSSL` branch miss

In particular, this reveals that the `self.store.locked_put(self)` in `_do_refresh_request` is in a strange place. Someone want to handle filing a bug?

Continuing from #477

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -25,7 +25,9 @@ import json
 import os
 import socket
 import sys
+import tempfile
 
+import httplib2
 import mock
 import six
 from six.moves import http_client
@@ -296,7 +298,6 @@ class GoogleCredentialsTests(unittest2.TestCase):
 
     def setUp(self):
         self.os_name = os.name
-        from oauth2client import client
         client.SETTINGS.env_name = None
 
     def tearDown(self):
@@ -607,7 +608,6 @@ class GoogleCredentialsTests(unittest2.TestCase):
         in_gae.assert_called_once_with()
 
     def test_env_name(self):
-        from oauth2client import client
         self.assertEqual(None, client.SETTINGS.env_name)
         self.test_get_adc_from_environment_variable_service_account()
         self.assertEqual(DEFAULT_ENV_NAME, client.SETTINGS.env_name)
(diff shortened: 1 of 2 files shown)


jonparrott:
LGTM.

What's strange about the locked_put? Is it better placed in `_refresh`?

dhermes:
It's strange which branch it occurs in. It should be unconditional on a failure, right?

jonparrott:
Good question. It seems it also updates it in the else branch. So... idk?

tests/test_client.py
```diff
@@ -1074,6 +1103,131 @@ def test_get_access_token(self, utcnow):
...
+
+        call1 = mock.call('Refreshing access_token')
+        failure_template = 'Failed to retrieve access token: %s'
+        call2 = mock.call(failure_template, content)
+        self.assertEqual(logger.info.mock_calls, [call1, call2])
+        if store is not None:
+            store.locked_put.assert_called_once_with(credentials)
+
+    def test__do_refresh_request_non_json_failure(self):
+        import httplib2
```
nathanielmanistaatgoogle: Huh? This is a unit test module, surely it must be able to have `httplib2` imported at module top-level?

dhermes: Nope. But I can add it top-level if you like. I intentionally left it here to limit the scope of this test case.

dhermes: And to restrict any import failures to occur at runtime (instead of import time). This is a philosophy adopted from @tseaver

nathanielmanistaatgoogle: Yeah, please put it at top-level. It's what the style guide requires and local imports are rare enough to invite the reader to believe that they are necessary.

In a test, a local import almost always indicates "testing importability and import-time-behavior of the system under test". But `httplib2` isn't part of the system under test for this test code, so it being locally imported is just a distraction.

dhermes: @nathanielmanistaatgoogle Moving it now, though we should discuss @tseaver testing philosophy sometime. It's got a lot of great insights.

nathanielmanistaatgoogle:
Change content looks great; merge when code review conversation resolves?

dhermes:
Filed #480. That was the point of the discussion any how (code changes not blocking for the tests added).

Issue #470, opened 2016-03-17
Remove multistore file/locked file


jonparrott:
Multistore file and locked file are designed to fit a very specific use case:

> Safely use a credential across multiple processes.

They unfortunately also represent a great deal of cruft and technical debt in this repository. I would greatly prefer to do one of the following:

1. Remove `locked_file` and use an existing interprocess lock library such as [fasteners](http://fasteners.readthedocs.org/en/latest/examples.html#interprocess-locks). Keep `multistore_file`.
2. Remove both.

If we go route two, downstream users of this file have two options:

1. Pull `multistore_file` and `locked_file` into their own codebase. Considering that these files are now completely separated from core and `oauth2client` has always intended users to write their own storage classes if needed, this doesn't seem like a bad option.
2. We could provide a version of `multistore_file` based on sqlite which has concurrency built-in. However, downstream users would have to take on an additional dependency with a native component. If that is untenable, they could still go with route one.

/cc @thobrla @dhermes @nathanielmanistaatgoogle

thobrla:
Ideally whatever changes we make here should be transparent to users.

Option 1 sounds fine.
Option 2.1 effectively removes support for this use case, so I strongly dislike it.  It's going to be hard for users to maintain forks of those files given how frequently the underlying oauth2client implementation changes.
Option 2.2. sounds fine assuming there are no issues with that component.

jonparrott:
> Option 2.1 effectively removes support for this use case, so I strongly dislike it. It's going to be hard for users to maintain forks of those files given how frequently the underlying oauth2client implementation changes.

Storage doesn't seem likely to change in any major way. I tried and failed to do so. From what I can tell, the only users of this are first-party tools.

dhermes:
@thobrla 

> how frequently the underlying `oauth2client` implementation changes.

This will certainly slow down (hopefully at some point in 2016). The reason for changes lately is because the codebase is in such bad shape, people like @jonparrott, @nathanielmanistaatgoogle and I have been trying to rescue it.

jonparrott:
@thobrla @nathanielmanistaatgoogle

The more I think about it and the more I interact with this code, I'm extremely inclined to remove it. Cloud SDK and gsutil are the biggest public users, but a cursory internal search shows that there are some internal clients that still use it. @nathanielmanistaatgoogle is technically able to add a separate library to provide this functionality for internal users, but external users would still have to absorb the code into their repository.

[...]

PR #484, opened 2016-04-07
Clarify control flow in LockedFile constructor

nathanielmanistaatgoogle:
Eliminate the unnecessary local field "opener" and ensure that
self._opener is assigned exactly once along every trajectory through
the constructor body.

I don't believe the `not opener` on line 349 of the old implementation could ever have been false, and if `_Win32Opener` and `_FcntlOpener` were both non-None then for calls with `use_native_locking` true the old implementation would have constructed an object just to throw it away. Right?

### DIFF
--- a/oauth2client/contrib/locked_file.py
+++ b/oauth2client/contrib/locked_file.py
@@ -345,17 +345,15 @@ class LockedFile(object):
             use_native_locking: bool, Whether or not fcntl/win32 locking is
                                 used.
         """
-        opener = None
-        if not opener and use_native_locking:
-            if _Win32Opener:
-                opener = _Win32Opener(filename, mode, fallback_mode)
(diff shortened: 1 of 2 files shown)


dhermes:
@nathanielmanistaatgoogle Careful not to step on #478 toes. Assigned to @jonparrott since he is in the process of sorting this out.

jonparrott:
This slipped under my radar and I apologize for that. @nathanielmanistaatgoogle if you want to resolve the conflicts then this LGTM. Otherwise, we can just hold off on this in favor of #504's eventual outcome.

nathanielmanistaatgoogle:
Yeah, I'm perfectly happy to just sit on this for a while.

jonparrott:
Closing because #470's outcome will be the removal of this module.

PR #485, opened 2016-04-07
Adding test coverage for client.py.

dhermes:
- `DeviceFlowInfo` (and using datetime now patch).
- Corner cases for `step1_get_authorize_url` and
  `step2_exchange`
- Failure cases for `flow_from_clientsecrets`
- `verify_id_token` when the cached HTTP is used (i.e.
  the default argument)

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -120,6 +120,7 @@ _DESIRED_METADATA_FLAVOR = 'Google'
 # Expose utcnow() at module level to allow for
 # easier testing (by replacing with a stub).
 _UTCNOW = datetime.datetime.utcnow
+_NOW = datetime.datetime.now
 
 
 class SETTINGS(object):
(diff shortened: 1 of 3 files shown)


oauth2client/client.py
```diff
@@ -1862,7 +1863,7 @@ def FromResponse(cls, response):
         })
         if 'expires_in' in response:
             kwargs['user_code_expiry'] = (
-                datetime.datetime.now() +
+                _NOW() +
```
dhermes, line 1866: @nathanielmanistaatgoogle The device flow code chunks seem to be very haphazard so I'd guess `utcnow` was intended here instead of `now`. What are the implications of us righting this wrong?

dhermes:
As of this PR only a few spots left in `client.py`. Very exciting!

```
2001-2039
2070
2085
2193
2069->2070
2084->2085
2190->2192
2192->2193
```

jonparrott:
LGTM once @nathanielmanistaatgoogle answers your question.

oauth2client/client.py
nathanielmanistaatgoogle, line 1866: We would break [this code](http://furyu.hatenablog.com/entry/20150820/1440064896). But that's the only code I could find.

I agree that the value should be UTC. Maybe make the change in its own pull request, so it at least shows up prominently in release notes?

nathanielmanistaatgoogle:
Look good; merge at your leisure.

PR #488, opened 2016-04-07
Replacing now() with utcnow() in device info code.

dhermes:
This change was discussed in #485.

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -120,7 +120,6 @@ _DESIRED_METADATA_FLAVOR = 'Google'
 # Expose utcnow() at module level to allow for
 # easier testing (by replacing with a stub).
 _UTCNOW = datetime.datetime.utcnow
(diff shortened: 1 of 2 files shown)


nathanielmanistaatgoogle:
Commit looks good; merge when test results land.

dhermes:
So we aren't worried about breakage?

nathanielmanistaatgoogle:
I'm a little worried, but not as much as I would be had the library ever documented the attribute as being in local time.

Do we feel that the value is worth the cost?

Should this be put on ice and rolled into whatever it is that motivates a more dramatic breaking change?

dhermes:
I don't mind if it gets put on ice but have no desire to be in charge of 

- Keeping track of when it is appropriate to merge this
- Keeping this PR up to date with the library as it gets updated

jonparrott:
I'm in favor of merging this to release with 2.1.0

PR #492, opened 2016-04-08
100% coverage for oauth2client.contrib.multistore_file

theacodes:


tests/contrib/test_multistore_file.py
```diff
@@ -133,6 +147,33 @@ def test_read_only_file_fail_lock(self):
...
+    def test_read_only_file_fail_lock_no_warning(self):
+        filehandle, filename = tempfile.mkstemp('oauth2client_test.data')
+        os.chmod(filename, 0o400)
+
+        multistore = multistore_file._MultiStore(filename)
+
+        with mock.patch.object(multistore_file.logger, 'warn') as mock_warn:
+            multistore._warn_on_readonly = False
+            multistore._lock()
+            self.assertFalse(mock_warn.called)
```
dhermes, line 159: Officially I am supposed to recommended that you clean up the file after the test

dhermes:
LGTM other than nit about cleaning up the `os` file resources

dhermes:
We have a context manager for making temp files: https://github.com/GoogleCloudPlatform/gcloud-python/blob/master/gcloud/_testing.py

Can't remember why we need it, something Windows related I think. But could be useful here

jonparrott:
I just re-used their top-level file which has cleanup logic in setup and teardown. I'll consider the context managers when I re-write.

This makes me miss py.test, which has proper fixtures.

dhermes:
Very tricky!

dhermes:
Alright @waprin let's cross that finish line!

PR #490, opened 2016-04-08
Omit platform-specific openers from coverage.

theacodes:

PR #493, opened 2016-04-08
Get django_orm to 100% coverage

waprin:
This is a first-pass at "shortest path to 100% coverage". It aggressively mocks out Django model methods and doesn't assert all too much. I am hesitant to invest too much in asserting more since I want to refactor this whole file anyway and merge it into django_util. 

If we want to mock out Django models less, we would need to figure out how to rig up the Django test database for the tests which I think is not what people want (although it would probably be significantly easier with something like Jon's nox).

### DIFF
--- a/tests/contrib/test_django_orm.py
+++ b/tests/contrib/test_django_orm.py
@@ -38,6 +38,8 @@ import django
 django.setup()
 from django.apps import AppConfig
 
+import mock
+
 
 class DjangoOrmTestApp(AppConfig):
     """App Config for Django Helper."""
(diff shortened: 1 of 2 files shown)


waprin:
Screwed up the last test hold on

tests/contrib/test_django_orm.py
```diff
@@ -138,27 +141,111 @@ def setUp(self):
...
 
         self.assertEqual(storage.model_class, FakeCredentialsModel)
-        self.assertEqual(storage.key_name, key_name)
-        self.assertEqual(storage.key_value, key_value)
-        self.assertEqual(storage.property_name, property_name)
+        self.assertEqual(storage.key_name, self.key_name)
+        self.assertEqual(storage.key_value, self.key_value)
+        self.assertEqual(storage.property_name, self.property_name)
+
+    @mock.patch("django.db.models")
```
dhermes: Single quotes for consistency

tests/contrib/test_django_orm.py
```diff
@@ -38,6 +38,9 @@
 django.setup()
 from django.apps import AppConfig
 
+import mock
+from mock import MagicMock
```
dhermes: Why not just use `mock.MagicMock` below? Also, as @jonparrott taught me, only `mock.Mock` is needed in most situations, unless you need magic methods also mocked.

tests/contrib/test_django_orm.py
```diff
@@ -138,27 +141,111 @@ def setUp(self):
...
+    def __init__(self):
+        self.model = CredentialWithSetStore
+
+    def set_store(self, storage):
+        pass
 
 
 class FakeCredentialsModel(models.Model):
+    """ FakeCredentials class that uses actual Django models methods.
+    """
```
dhermes: Why `"""` on its own line?

tests/contrib/test_django_orm.py
```diff
@@ -138,27 +141,111 @@ def setUp(self):
...
+    def __init__(self):
+        self.deleted = False
+
     credentials = CredentialsField()
 
 
+class FakeCredentialsModelMock(object):
+    """ FakeCredentials class for when mocking out Django model methods is
+    necessary.
+    """
```
dhermes: This docstring is longer than 80 chars, but since in a test, no need for a docstring, just delete it so we don't have to fight it?

tests/contrib/test_django_orm.py
```diff
@@ -138,27 +141,119 @@ def setUp(self):
...
+    def __init__(self, *args, **kwargs):
+        self.model = FakeCredentialsModelMock
+        self.saved = False
+        self.deleted = False
+
+    def save(self):
+        self.saved = True
+
+    credentials = CredentialWithSetStore()
+
```
dhermes, line 247: Two newlines before `if __name__ ...`

waprin:
Done to all your comments, also fixed up last test.

dhermes:
LGTM @jonparrott can you give it a final pass?

dhermes:
@waprin What's your plan for `oauth2client/contrib/django_util/apps.py`?

dhermes:
>  coverage/coveralls — Coverage increased (+0.2%) to 99.974%

w00t

jonparrott:
LGTM

dhermes:
OK @waprin how do we get to the finish line? What needs to happen for`apps.py` to get executed? We can also just ignore it if it shouldn't be executed.

dhermes:
Also some branch misses in `django_orm.py`:

```
oauth2client/contrib/django_orm.py                  75      0     22      4    96%
37->39, 74->76, 143->147, 145->147
```

waprin:
@dhermes apps.py basically does nothing, it's not actually used. Django lets you introspect on your [applications](https://docs.djangoproject.com/en/1.9/ref/applications/) so to be consistent we should have it. Writing actual tests for it will be tricky without the whole Django test runner and is of questionable value. 

I vote to just ignore it. 

I can fix the branch coverage, forgot to run it with that.

dhermes:
OK, can do as in #490 to ignore it? Just toss the ignore into your PR to cover those branches for `django_orm`?

waprin:
Sound good.

Issue #128, opened 2015-02-04
Enable BYO http library


dhermes:
If this is a duplicate I apologize.

ISTM this requires
- [x] Move all uses of `httplib2` into a module (call it `transport` probably) (#554, #559, #561, #577)
- [ ] Determine what assumed interface of `httplib2.Http` is needed
- [ ] Optional. Try to reduce the interface needed
- [x] Factor out the hard dependencies on `httplib2` and just use these interfaces
- [ ] Document the interface and allow credentials constructors to optionally take a custom `Http` object from any old library, e.g. `twisted` or `requests`.
- [ ] Revisit all changes and ensure consistency
- [ ] Revisit query string update hackery in #622

dhermes:
@nathanielmanistaatgoogle I think as a first start, using a first-class `Http` object (owned by `oauth2client`) as the output of `Credentials.authorize`. Right now, the wrapped `http` object just gets [monkey patched][1] and has a tenuous relationship with the original `Http` object and the original `Credentials` object.

[1]: https://github.com/google/oauth2client/blob/6e3c0c58f5a3ba18dc69bce9c49a975738c46c74/oauth2client/client.py#L609-L610

dhermes:
Here are all places an `httplib2.Http` object is created:

```bash
$ git log --pretty=%H -1  # current HEAD
6e3c0c58f5a3ba18dc69bce9c49a975738c46c74
$ git grep 'Http('
oauth2client/appengine.py:        return self.credentials.authorize(httplib2.Http(*args, **kwargs))
oauth2client/client.py:            h = httplib2.Http()
oauth2client/client.py:                http = httplib2.Http()
oauth2client/client.py:        http = httplib2.Http()
oauth2client/client.py:_cached_http = httplib2.Http(MemoryCache())
oauth2client/client.py:            http = httplib2.Http()
oauth2client/client.py:            http = httplib2.Http()
oauth2client/flask_util.py:        return self.credentials.authorize(httplib2.Http(*args, **kwargs))
samples/oauth2_for_devices.py:youtube = build("youtube", "v3", http=credentials.authorize(httplib2.Http()))
scripts/run_system_tests.py:    http = credentials.authorize(httplib2.Http())
tests/test_appengine.py:        http = httplib2.Http()
tests/test_appengine.py:        http = httplib2.Http()
tests/test_appengine.py:        http = httplib2.Http()
tests/test_appengine.py:            http = httplib2.Http()
```

[...]

PR #503, opened 2016-04-22
Added self signed jwt service account credentials

kpayson64:
Initial self signed jwt credential support #252.  Here is my thinking with this initial API:

-SelfSignedJWTCredentials class should not be on the OAuth2Client inheritance hierarchy, there are many oauth2 specific methods (like refresh) that don't apply to self signed credentials.
-SelfSignedJWTCredentials can be placed in the service_account module,  SelfSignedJWTCredentials are highly coupled with service account credentials, self signed jwts can only be signed using service account credentials
-get_access_token() can be the external interface for generating tokens, it is consistent with the GoogleCredentials API, and there is nothing OAuth2 specific about the naming.
-The fields contained in the JWT are Google specific as far as I can tell, so this should be considered a Google specific class

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -495,6 +495,26 @@ def _update_query_params(uri, params):
     return urllib.parse.urlunparse(new_parts)
 
 
+def _initialize_headers(headers):
+    """Creates a copy of the headers."""
+    if headers is None:
+        headers = {}
+    else:
+        headers = dict(headers)
+    return headers
+
+
+def _apply_user_agent(headers, user_agent):
+    """Adds a user-agent to the headers."""
+    if user_agent is not None:
+        if 'user-agent' in headers:
+            headers['user-agent'] = (user_agent + ' ' + headers['user-agent'])
+        else:
+            headers['user-agent'] = user_agent
+
+    return headers
+
+
 class OAuth2Credentials(Credentials):
     """Credentials object for OAuth 2.0.
 
@@ -598,18 +618,9 @@ class OAuth2Credentials(Credentials):
 
             # Clone and modify the request headers to add the appropriate
             # Authorization header.
-            if headers is None:
-                headers = {}
-            else:
-                headers = dict(headers)
+            headers = _initialize_headers(headers)
             self.apply(headers)
-
-            if self.user_agent is not None:
-                if 'user-agent' in headers:
-                    headers['user-agent'] = (self.user_agent + ' ' +
-                                             headers['user-agent'])
-                else:
-                    headers['user-agent'] = self.user_agent
+            _apply_user_agent(headers, self.user_agent)
 
             body_stream_position = None
             if all(getattr(body, stream_prop, None) for stream_prop in
@@ -1237,6 +1248,7 @@ class GoogleCredentials(OAuth2Credentials):
         # TODO(issue 388): eliminate the circularity that is the reason for
         #                  this non-top-level import.
         from oauth2client.service_account import ServiceAccountCredentials
+        from oauth2client.service_account import _JWTAccessCredentials
         data = json.loads(_from_bytes(json_data))
 
         # We handle service_account.ServiceAccountCredentials since it is a
@@ -1244,6 +1256,10 @@ class GoogleCredentials(OAuth2Credentials):
         if (data['_module'] == 'oauth2client.service_account' and
             data['_class'] == 'ServiceAccountCredentials'):
             return ServiceAccountCredentials.from_json(data)
+        elif (data['_module'] == 'oauth2client.service_account' and
+              data['_class'] == '_JWTAccessCredentials'):
+            return _JWTAccessCredentials.from_json(data)
+        
 
         token_expiry = _parse_expiry(data.get('token_expiry'))
         google_credentials = cls(
@@ -1523,8 +1539,8 @@ def _get_application_default_credential_from_file(filename):
             token_uri=GOOGLE_TOKEN_URI,
             user_agent='Python client library')
     else:  # client_credentials['type'] == SERVICE_ACCOUNT
-        from oauth2client.service_account import ServiceAccountCredentials
-        return ServiceAccountCredentials.from_json_keyfile_dict(
+        from oauth2client.service_account import _JWTAccessCredentials
+        return _JWTAccessCredentials.from_json_keyfile_dict(
             client_credentials)
 
 
--- a/oauth2client/service_account.py
+++ b/oauth2client/service_account.py
@@ -17,6 +17,7 @@
 import base64
 import copy
 import datetime
+import httplib2
 import json
 import time
 
@@ -26,9 +27,16 @@ from oauth2client._helpers import _json_encode
 from oauth2client._helpers import _from_bytes
 from oauth2client._helpers import _urlsafe_b64encode
 from oauth2client import util
+from oauth2client.client import _apply_user_agent
+from oauth2client.client import _initialize_headers
+from oauth2client.client import AccessTokenInfo
 from oauth2client.client import AssertionCredentials
+from oauth2client.client import clean_headers
 from oauth2client.client import EXPIRY_FORMAT
+from oauth2client.client import GoogleCredentials
 from oauth2client.client import SERVICE_ACCOUNT
+from oauth2client.client import TokenRevokeError
+from oauth2client.client import _UTCNOW
 from oauth2client import crypt
 
 
@@ -426,6 +434,32 @@ class ServiceAccountCredentials(AssertionCredentials):
         result._private_key_pkcs12 = self._private_key_pkcs12
         result._private_key_password = self._private_key_password
         return result
+ 
+    def create_with_claims(self, claims):
+        """Create credentials that specify additional claims.
+
+        Args:
+            claims: dict, key-value pairs for claims.
+
+        Returns:
+            ServiceAccountCredentials, a copy of the current service account
+            credentials with updated claims to use when obtaining access tokens.
+        """
+        new_kwargs = dict(self._kwargs)
+        new_kwargs.update(claims)
+        result = self.__class__(self._service_account_email,
+                                self._signer,
+                                scopes=self._scopes,
+                                private_key_id=self._private_key_id,
+                                client_id=self.client_id,
+                                user_agent=self._user_agent,
+                                **new_kwargs)
+        result.token_uri = self.token_uri
+        result.revoke_uri = self.revoke_uri
+        result._private_key_pkcs8_pem = self._private_key_pkcs8_pem
+        result._private_key_pkcs12 = self._private_key_pkcs12
+        result._private_key_password = self._private_key_password
+        return result
 
     def create_delegated(self, sub):
         """Create credentials that act as domain-wide delegation of authority.
@@ -446,18 +480,158 @@ class ServiceAccountCredentials(AssertionCredentials):
             ServiceAccountCredentials, a copy of the current service account
             updated to act on behalf of ``sub``.
         """
-        new_kwargs = dict(self._kwargs)
-        new_kwargs['sub'] = sub
-        result = self.__class__(self._service_account_email,
-                                self._signer,
-                                scopes=self._scopes,
-                                private_key_id=self._private_key_id,
-                                client_id=self.client_id,
-                                user_agent=self._user_agent,
-                                **new_kwargs)
+        return self.create_with_claims({'sub': sub})
+
+
+def _datetime_to_secs(utc_time):
+   # TODO(issue 298): use time_delta.total_seconds()
+   # time_delta.total_seconds() not supported in Python 2.6
+   epoch = datetime.datetime(1970, 1, 1)
+   time_delta = utc_time - epoch
+   return time_delta.days * 86400 + time_delta.seconds
+
+
+class _JWTAccessCredentials(ServiceAccountCredentials):
+    """Self signed JWT credentials.
+
+    Makes an assertion to server using a self signed JWT from service account
+    credentials.  These credentials do NOT use OAuth 2.0 and instead
+    authenticate directly.
+    """
+    _MAX_TOKEN_LIFETIME_SECS = 3600
+    """Max lifetime of the token (one hour, in seconds)."""
+
+    def __init__(self,
+                 service_account_email,
+                 signer,
+                 scopes=None,
+                 private_key_id=None,
+                 client_id=None,
+                 user_agent=None,
+                 additional_claims=None):
+        if additional_claims is None:
+            additional_claims = {}
+        super(_JWTAccessCredentials, self).__init__(
+            service_account_email,
+            signer,
+            private_key_id=private_key_id,
+            client_id=client_id,
+            user_agent=user_agent,
+            **additional_claims)
+
+    def authorize(self, http):
+        """Authorize an httplib2.Http instance with a JWT assertion.
+
+        Unless specified, the 'aud' of the assertion will be the base
+        uri of the request.
+
+        Args:
+            http: An instance of ``httplib2.Http`` or something that acts
+                  like it.
+        Returns:
+            A modified instance of http that was passed in.
+        Example::
+            h = httplib2.Http()
+            h = credentials.authorize(h)
+        """
+        request_orig = http.request
+        request_auth = super(_JWTAccessCredentials, self).authorize(http).request
+
+        # The closure that will replace 'httplib2.Http.request'.
+        def new_request(uri, method='GET', body=None, headers=None,
+                        redirections=httplib2.DEFAULT_MAX_REDIRECTS,
+                        connection_type=None):
+            if 'aud' in self._kwargs:
+                # Preemptively refresh token, this is not done for OAuth2
+                if self.access_token is None or self.access_token_expired:
+                    self.refresh(None)
+                return request_auth(uri, method, body,
+                                    headers, redirections,
+                                    connection_type)
+            else:
+                # If we don't have an 'aud' (audience) claim,
+                # create a 1-time token with the uri root as the audience
+                headers = _initialize_headers(headers)
+                _apply_user_agent(headers, self.user_agent)
+                uri_root = uri.split('?', 1)[0]
+                token, unused_expiry = self._create_token({'aud': uri_root})
+
+                headers['Authorization'] = 'Bearer ' + token
+                return request_orig(uri, method, body,
+                                    clean_headers(headers),
+                                    redirections, connection_type)
+
+        # Replace the request method with our own closure.
+        http.request = new_request
+
+        return http
+
+    def get_access_token(self, http=None, additional_claims=None):
+        """Create a signed jwt.
+
+        Args:
+            http: unused
+            additional_claims: dict, additional claims to add to
+                the payload of the JWT.
+        Returns:
+            An AccessTokenInfo with the signed jwt
+        """
+        if additional_claims is None:
+            if self.access_token is None or self.access_token_expired:
+                self.refresh(None)
+            return AccessTokenInfo(access_token=self.access_token,
+                                   expires_in=self._expires_in())
+        else:
+            # Create a 1 time token
+            token, unused_expiry = self._create_token(additional_claims)
+            return AccessTokenInfo(access_token=token,
+                                   expires_in=self._MAX_TOKEN_LIFETIME_SECS)
+
+    def revoke(self, http):
+        """Cannot revoke JWTAccessCredentials tokens."""
+        pass
+
+    def create_scoped_required(self):
+        # JWTAccessCredentials are unscoped by definition
+        return True
+
+    def create_scoped(self, scopes):
+        # Returns an OAuth2 credentials with the given scope
+        result = ServiceAccountCredentials(self._service_account_email,
+                                           self._signer,
+                                           scopes=scopes,
+                                           private_key_id=self._private_key_id,
+                                           client_id=self.client_id,
+                                           user_agent=self._user_agent,
+                                           **self._kwargs)
         result.token_uri = self.token_uri
         result.revoke_uri = self.revoke_uri
-        result._private_key_pkcs8_pem = self._private_key_pkcs8_pem
-        result._private_key_pkcs12 = self._private_key_pkcs12
-        result._private_key_password = self._private_key_password
+        if self._private_key_pkcs8_pem is not None:
+            result._private_key_pkcs8_pem = self._private_key_pkcs8_pem
+        if self._private_key_pkcs12 is not None:
+            result._private_key_pkcs12 = self._private_key_pkcs12
+        if self._private_key_password is not None:
+            result._private_key_password = self._private_key_password
         return result
+
+    def refresh(self, http):
+        self._refresh(None)
+
+    def _refresh(self, http_request):
+        self.access_token, self.token_expiry = self._create_token()
+
+    def _create_token(self, additional_claims=None):
+        now = _UTCNOW()
+        expiry = now + datetime.timedelta(seconds=self._MAX_TOKEN_LIFETIME_SECS)
+        payload = {
+            'iat': _datetime_to_secs(now),
+            'exp': _datetime_to_secs(expiry),
+            'iss': self._service_account_email,
+            'sub': self._service_account_email
+        }
+        payload.update(self._kwargs)
+        if additional_claims is not None:
+            payload.update(additional_claims)
+        jwt = crypt.make_signed_jwt(self._signer, payload,
+                                    key_id=self._private_key_id)
+        return jwt.decode('ascii'), expiry
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -840,6 +840,20 @@ class GoogleCredentialsTests(unittest2.TestCase):
         creds2_vals.pop('_signer')
         self.assertEqual(creds1_vals, creds2_vals)
 
+    def test_to_from_json_service_account_scoped(self):
+        credentials_file = datafile(
+            os.path.join('gcloud', _WELL_KNOWN_CREDENTIALS_FILE))
+        creds1 = GoogleCredentials.from_stream(credentials_file)
+        creds1 = creds1.create_scoped(['dummy_scope'])
+        # Convert to and then back from json.
+        creds2 = GoogleCredentials.from_json(creds1.to_json())
+
+        creds1_vals = creds1.__dict__
+        creds1_vals.pop('_signer')
+        creds2_vals = creds2.__dict__
+        creds2_vals.pop('_signer')
+        self.assertEqual(creds1_vals, creds2_vals)
+
     def test_parse_expiry(self):
         dt = datetime.datetime(2016, 1, 1)
         parsed_expiry = client._parse_expiry(dt)
--- a/tests/test_service_account.py
+++ b/tests/test_service_account.py
@@ -23,11 +23,13 @@ import os
 import rsa
 import tempfile
 
+import httplib2
 import mock
 import unittest2
 
 from .http_mock import HttpMockSequence
 from oauth2client import crypt
+from oauth2client.service_account import _JWTAccessCredentials
 from oauth2client.service_account import ServiceAccountCredentials
 from oauth2client.service_account import SERVICE_ACCOUNT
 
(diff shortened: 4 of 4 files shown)


oauth2client/service_account.py
```diff
@@ -461,3 +462,56 @@ def create_delegated(self, sub):
...
+        }
+        payload.update(self._creds._kwargs)
+        access_token = crypt.make_signed_jwt(self._creds._signer, payload,
+                                      key_id=self._creds._private_key_id)
+    
+        return AccessTokenInfo(access_token=access_token,
+                              expires_in=self._time_limit)
+
+    @classmethod
+    def from_service_account_credentials(cls, creds, audience,
```
dhermes: This is identical to the constructor.

oauth2client/service_account.py
```diff
@@ -461,3 +462,56 @@ def create_delegated(self, sub):
...
+    """
+    MAX_TOKEN_LIFETIME_SECS = 3600
+    """Default lifetime of the token (one hour, in seconds)."""
+
+    def __init__(self, creds, audience, time_limit):
+        self._creds = creds
+        self._audience = audience
+        self._time_limit = time_limit
+    
+    def get_access_token(self):
```
theacodes: `access_token` is OAuth2-specific terminology and doesn't apply to jwts. I recommend following pyjwt and using `encode`.

Additionally, this method should allow you to specify additional claims to go in the payload.

oauth2client/service_account.py
```diff
@@ -461,3 +462,56 @@ def create_delegated(self, sub):
...
+            Returns:
+                An AccessTokenInfo with the signed jwt
+        """
+        now = int(time.time())
+        payload = {
+            'aud': self._audience,
+            'iat': now,
+            'exp': now + self._time_limit,
+            'iss': self._creds._service_account_email,
+            'sub': self._creds._service_account_email
```
theacodes: subject is not always the service account email. The user of the API should be able to specify their own subject.

oauth2client/service_account.py
```diff
@@ -461,3 +462,56 @@ def create_delegated(self, sub):
...
+    Makes an assertion to server using a self signed JWT from service account
+    credentials.  These credentials do NOT use OAuth 2.0 and instead authenticate
+    directly.
+    """
+    MAX_TOKEN_LIFETIME_SECS = 3600
+    """Default lifetime of the token (one hour, in seconds)."""
+
+    def __init__(self, creds, audience, time_limit):
+        self._creds = creds
+        self._audience = audience
```
theacodes: I don't see why you couldn't use the same JWT Credentials instance to encode jwts for multiple audiences. I feel like this should be an argument on `encode`.

oauth2client/service_account.py
kpayson64: @jonparrott 
I agree with making `encode` function more flexible.  Would `encode(self, audience, kwargs**)` be a good approach in your opinion?

theacodes: honestly just doing `encode(self, claims)` would be preferable for me.

In the function, you could just do `payload.update(claims)`.

jonparrott:
@kpayson64 Thanks for contributing this! :) 👍

kpayson64:
@jtattermusch

jonparrott:
@anthmgoogle @dhermes @nathanielmanistaatgoogle  Care to chime in on API and naming conventions here?

I'd personally prefer this API (which closely matches pyjwt's):

```python
service_account = oauth2client.service_account.ServiceAccount.from_json_keyfile_name('service-account.json')
jwt = oauth2client.jwt.encode(service_account)
```

Additional claims:

```python
jwt = oauth2client.jwt.encode(service_account, {'sub': 'user@example.com'})
```

@nathanielmanistaatgoogle would probably be happy to see less classes. :D

oauth2client/service_account.py
kpayson64: Because this is a google specific JWT, I think having audience as an explicit parameter is a good idea.  Self-signed JWTs will always require an audience, so I think that contract should be baked into the API.

theacodes: It's required in the spec, I'm not suggesting leaving it out, just letting the caller override it.

jonparrott:
We could also expose this as `ServiceAccount.encode_jwt`, since a lot of this logic is already in `ServiceAccount._generate_assertion`. ¯\_(ツ)_/¯

dhermes:
@jonparrott That misses the mark. These credentials need to be able to be used as a swap-in replacement for credentials.

jonparrott:
@dhermes interesting. Can you expand more on that? JWTs are *not* credentials, they are signed claims that can be used for authorization. The service account private key is the credential.

jonparrott:
Is the desire to do something like this?

```python
credentials = JwtCredentials(service_account_credentials)
credentials.authorize(http)
# http now sends the JWT in the Authorization header
```

If so, I don't see how this is distinct from just making the existing `ServiceAccount` accept a `claims` arg in its constructor.

dhermes:
For APIs like Cloud Bigtable, the JWT can be used **as** the access token (so-called scopless credentials). Hence, we want a class to allow this, and maybe have some built-in retry behavior which regenerates the local token (this PR doesn't quite get there either).

dhermes:
@jonparrott Sorry we posted at essentially the same moment.

The difference is that `ServiceAccountCredentials` uses the JWT to exchange for a bearer token (which requires a round trip).

jonparrott:
Ah, I see, I should get more sleep before discussing auth.

So, I think I'd prefer "both". We should separate the logic of generating a JWT vs using the jwt.

I suggest we:

* Add `oauth2client.jwt.encode` that accepts a signer and set of claims and returns a JWT.
* Make this new class use it to generate its access token.
* Make `ServiceAccount` use it to generate its assertion.

This covers all use-cases well and prevents duplicate code.

kpayson64:
@jonparrott 
I think the oauth2client.jwt.encode you are suggesting is essentially  crypt.make_signed_jwt.

@dhermes 
Do we want the new class to implement client.Credentials?  Could you be more specific with the functionality that is missing?

oauth2client/service_account.py
```diff
@@ -461,3 +462,56 @@ def create_delegated(self, sub):
         result._private_key_pkcs12 = self._private_key_pkcs12
         result._private_key_password = self._private_key_password
         return result
+
+
+class SelfSignedJWTCredentials(object):
```
jboeuf: JWTs are always self-signed whether you use them directly (like here) or exchange them for an oauth2 access token. Here are some suggestions to rename them:
- JWTAccessCredentials (good)
- JWTScopelessCredentials (not as good).

oauth2client/service_account.py
jboeuf: There should be only one audience for these credentials which is the service url that these tokens are going to be sent to.

oauth2client/service_account.py
dhermes: How many users know what a JWT is? In my experience with service accounts, the previous name `SignedJWTCredentials` was confusing to users (and to contributors / Googlers, some of which accidentally created a new service account class because they didn't realize `SignedJWTCredentials` was for service accounts).

oauth2client/service_account.py
jboeuf: Fair enough. note that the these are called JWTAccessTokens as opposed to Oauth2AccessTokens though.

oauth2client/service_account.py
jboeuf: @jonparrott in which case is the subject not the service account email?

oauth2client/service_account.py
dhermes: @jboeuf are there some user facing docs for scopeless credentials? (**UPDATE**: I'm curious about terminology / the surface that a user sees.)

oauth2client/service_account.py
jboeuf: Good point. In C++ we call them `ServiceAccountJWTAccessCredentials`. Would that answer your concern?

oauth2client/service_account.py
jboeuf: I've been pushing for that internally but no response so far. Let me ping the thread again.

oauth2client/service_account.py
theacodes: Delegated credentials or using the envelope pattern for asserting a user across requests, for example:

``` python
    payload = {
        'iat': now,
        'exp': now + credentials.MAX_TOKEN_LIFETIME_SECS,
        # aud must match 'audience' in the security configuration in your
        # open api spec. It can be any string.
        'aud': 'echo.esp.sample.cloud.google.com',
        # iss must match 'issuer' in the security configuration in your
        # open api spec. it is typically the service account email, but
        # may be any string that uniquely identifies the issuer.
        'iss': service_account_email,
        # sub and email are mapped to the user id and email respectively.
        'sub': '12345678',
        'email': 'user@example.com'
    }
```

theacodes: N.B., the use case for that is for an upcoming product release. We can speak internally about it if needed.

oauth2client/service_account.py
dhermes: Probably. Just want to make sure this library uses the same language as the "official" docs do.

oauth2client/service_account.py
jboeuf: Thanks. Do you have a link to this delegation pattern?

oauth2client/service_account.py
theacodes: That's a mouthful. I think `JWTAccessCredentials` are pretty accurate without being verbose.

oauth2client/service_account.py
theacodes: One audience at a time, but do you really want to make users construct multiple instances just to talk to a different service?

oauth2client/service_account.py
theacodes: Fair enough, now that I better understand how these are used `access_token` is fine.

oauth2client/service_account.py
theacodes: https://developers.google.com/identity/protocols/OAuth2ServiceAccount#delegatingauthority

This case is used to acquire a bearer token, but the same principle applies here too.

oauth2client/service_account.py
jboeuf: Ah thanks. Yes, I agree then, It's probably better (if not needed) to put the audience in the `encode`/`get_access_token` method, whatever we call it.

theacodes: I take this back, due to how this class will be used everything should go into the constructor and `access_token` should take no arguments.

jonparrott:
Okay, after chatting with @jboeuf the API should shake out like this:

```
class JWTAccessCredentials(object):
    def __init__(self, service_account, additional_claims=None)
         pass
 
    def get_access_token(self, additional_claims=None)
          pass

    def create_with_claims(claims):
          pass
```

This allows for it to satisfy the `oauth2client.client.Credentials` interface as well as work for gRPC by allowing gRPC to explicitly call `get_access_token(additional_claims={'aud': service_url})` without having to create tons of instances.

kpayson64:
This latest commit has the new API.  Also, I have made JWTAccessCredentials inherit GoogleCredentials  .  I'm holding off on unit tests/documentation until I have an API nailed down.

oauth2client/service_account.py
```diff
@@ -461,3 +465,100 @@ def create_delegated(self, sub):
...
+    Makes an assertion to server using a self signed JWT from service account
+    credentials.  These credentials do NOT use OAuth 2.0 and instead authenticate
+    directly.
+    """
+    MAX_TOKEN_LIFETIME_SECS = 3600
+    """Max lifetime of the token (one hour, in seconds)."""
+
+    def __init__(self, service_account_creds, additional_claims=None, access_token=None,
+                 token_expiry=None):
+        super(JWTAccessCredentials, self).__init__(access_token, None, None, None, token_expiry,
```
theacodes: Why `GoogleCredentials` as the superclass? Why not just `oauth2client.client.Credentials`?

oauth2client/service_account.py
```diff
@@ -461,3 +465,100 @@ def create_delegated(self, sub):
...
+                                                   None, None)
+        self._creds = service_account_creds
+        self._claims = {}
+        self._claims.update(service_account_creds._kwargs)
```
theacodes: Hm.. I'm unsure if we want to do this. The JWT's claims may be quite different from the claims used to grab an access token. I would prefer us to create our own set of claims here.

oauth2client/service_account.py
```diff
@@ -461,3 +465,100 @@ def create_delegated(self, sub):
...
+          # Create a 1 time token
+          return self._create_token(additional_claims)
+        else:
+          if self.access_token is None or self.access_token_expired:
+            self._refresh(None)
+
+        return AccessTokenInfo(access_token=self.access_token,
+                              expires_in=self._expires_in())
+    def revoke(self, http):
+        # TODO(kpayson): Is there any way to actually invalidate a signed JWT?
```
theacodes: Please don't put todos in source files. And no, there is not way to invalidate a JWT. I think this should just pass, I don't think we need to raise an error here.

theacodes: @nathanielmanistaatgoogle may disagree, though.

oauth2client/service_account.py
```diff
@@ -461,3 +465,100 @@ def create_delegated(self, sub):
...
+        raise TokenRevokeError('Cannot revoke JWTAccessCredentials')
+
+    @classmethod
+    def from_json(cls, json_data):
+      if not isinstance(json_data, dict):
```
theacodes: Is this check present in other credentials?

oauth2client/service_account.py
```diff
@@ -461,3 +465,100 @@ def create_delegated(self, sub):
...
+      if to_serialize is None:
+            to_serialize = copy.copy(self.__dict__)
+      creds_val = to_serialize.get('_creds')
+      if creds_val is not None:
+        to_serialize['_creds'] = json.loads(self._creds.to_json())
+      return super(JWTAccessCredentials, self)._to_json(
+        strip, to_serialize=to_serialize)
+
+    def _refresh(self, http):
+        self.token_expiry = _UTCNOW() + datetime.timedelta(seconds=self.MAX_TOKEN_LIFETIME_SECS)
```
theacodes: This doesn't seem to be the exact same calculation used below. Is it possible for this to hurt you? Why not just have _create_token update the expiry?

oauth2client/service_account.py
```diff
@@ -461,3 +465,100 @@ def create_delegated(self, sub):
...
+            service_account_creds: ServiceAccountCredentials, the
+              credentials used to sign the JWTs
+            claims: dict, the claims to place in the JWT payload.
+              The 'aud' claim must be specified to authenticate with Google. 
+              Ex.  {'aud' : {https://googleapis/MyService'}
+
+        Returns:
+            SelfSignedJWTCredentials for the service account
+        """
+        return cls(service_account_creds, claims)
```
theacodes: This should merge the current instance's claims with the new claims and return an instance.

jonparrott:
API surface looks good, I completed a round of reviews. Please ping when you're ready for another review.

Side note: your file seems to have mixed indentation, please use 4 spaces.

oauth2client/service_account.py
kpayson64: I want to reuse most of the authorization functionality in OAuth2Credentials.  The only difference is how the tokens are refreshed.

Ideally we could pull the core authorize() functionality out of OAuth2Credentials, but I'm not sure if we would want to move that to Credentials.

oauth2client/service_account.py
kpayson64: ServiceAccountCredentials does, but it seems like for the most part json_data is assumed to be a string.  I'm removing the check.

oauth2client/service_account.py
kpayson64: `_create_token()` might be used to create a 1 time token if we specify additional claims.  I've updated _create_token to also return an expiry, this way we can ensure that the actual expiry matches the saved expiry.

oauth2client/service_account.py
theacodes: `GoogleCredentials` seems to pull in a lot of stuff that's not needed, my question was why that base class instead of a simpler one such as `Credentials` or `OAuth2Credentials`.

oauth2client/service_account.py
theacodes: Perfect, thanks.

oauth2client/service_account.py
kpayson64: Makes sense.

oauth2client/service_account.py
kpayson64: Removed.

kpayson64:
@jonparrott 
I've updated the pr from the feedback.

oauth2client/service_account.py
```diff
@@ -461,3 +465,96 @@ def create_delegated(self, sub):
         result._private_key_pkcs12 = self._private_key_pkcs12
         result._private_key_password = self._private_key_password
         return result
+
+
+def _datetime_to_secs(utc_time):
+   # time_delta.total_seconds() not supported in Python 2.6
```
theacodes, line 488: Sigh.

oauth2client/service_account.py
```diff
@@ -461,3 +465,96 @@ def create_delegated(self, sub):
...
+
+    @classmethod
+    def from_json(cls, json_data):
+        json_data = json.loads(_from_bytes(json_data))
+
+        service_account_creds = ServiceAccountCredentials.from_json(json_data['_creds'])
+        token_expiry = datetime.datetime.strptime(json_data['token_expiry'], EXPIRY_FORMAT)
+        retval = cls(service_account_creds, 
+                     additional_claims=json_data['_claims'],
+                     access_token=json_data['access_token'],
```
theacodes: Since the access token is an assertion, is it necessary to load this from json? Can't it just be re-generated?

jonparrott:
This mostly looks good to me, although I would really like to see the docstrings fleshed out.

Once my comments are addressed, feel free to ping @nathanielmanistaatgoogle to do the final review.

jonparrott:
Let me caveat that: don't ping @nathanielmanistaatgoogle until you've written tests as well, because that'll be the first thing he asks for. ;)

oauth2client/service_account.py
kpayson64: Agreed, this simplifies the from_json and constructor logic.

kpayson64:
I had a discussion with @jtattermusch, and another goal of this change is "Usable Auth".  The idea is that  when GoogleCredentials.get_application_default() is called, self signed jwt credentials will be used if no scope is specified.  This behavior is consistent with other google auth libraries.

I've attempted to do this is a fashion that minimally pollutes the API, and part of that includes having JWTAccessCredentials extend GoogleCredentials again.

kpayson64:
Before I write doc strings/unit tests, could @nathanielmanistaatgoogle and @jtattermusch take a look over the API?

oauth2client/service_account.py
nathanielmanistaatgoogle, line 488: Please make this a "TODO(issue 298): use time_delta.total_seconds()."?

dhermes, line 488: I am big time :-1: on `TODO` comments in code. There is a world-class issue tracker right here, put the `TODO` there (and you can even link to the PR / line of code with commit hash and all).

oauth2client/service_account.py
```diff
@@ -461,3 +465,94 @@ def create_delegated(self, sub):
...
+
+
+def _datetime_to_secs(utc_time):
+   # time_delta.total_seconds() not supported in Python 2.6
+   epoch = datetime.datetime(1970, 1, 1)
+   time_delta = utc_time - epoch
+   return time_delta.days * 86400 + time_delta.seconds
+
+
+class JWTAccessCredentials(GoogleCredentials):
```
nathanielmanistaatgoogle: You've chosen to use inheritance, so welcome to the inheritance questionnaire! :-)
- Why should instances of this class be-a `GoogleCredentials` rather than have-a `GoogleCredentials`?
- Does this class conform to the [Liskov Substitution Principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle)?
  - Are there places where a `GoogleCredentials` is requested and at those places is a `JWTAccessCredentials` acceptable?
  - Are there places where a `OAuth2Credentials` is requested and at those places is a `JWTAccessCredentials` acceptable?
  - Are there places where a `Credentials` is requested and at those places is a `JWTAccessCredentials` acceptable?
- Within the body of this class, which code elements override inherited concrete elements?
- Within the body of this class, which code elements implement inherited abstract elements?
- This class has seven public instance methods and one public class method. How many of those are intended to be used by users of the `oauth2client` library?

oauth2client/service_account.py
```diff
@@ -461,3 +465,94 @@ def create_delegated(self, sub):
...
+        """
+        if additional_claims is not None:
+            # Create a 1 time token
+            token, expiry = self._create_token(additional_claims)
+            return token
+        else:
+            if self.access_token is None or self.access_token_expired:
+                self._refresh(None)
+
+        return AccessTokenInfo(access_token=self.access_token,
```
nathanielmanistaatgoogle: Move this return into the `else` branch so that it's more clear that each branch has its own `return`.

oauth2client/service_account.py
```diff
@@ -461,3 +465,94 @@ def create_delegated(self, sub):
...
+    
+    def get_access_token(self, http=None, additional_claims=None):
+        """Create a signed jwt
+        Agrs:
+            additional_claims: dict, additional claims to add 
+                the payload of the JWT. 
+            Returns:
+                An AccessTokenInfo with the signed jwt
+        """
+        if additional_claims is not None:
```
nathanielmanistaatgoogle: Drop the `not` by flipping the branches.

oauth2client/service_account.py
```diff
@@ -461,3 +465,94 @@ def create_delegated(self, sub):
...
+            # Create a 1 time token
+            token, expiry = self._create_token(additional_claims)
```
nathanielmanistaatgoogle: `unused_expiry` since the name is not later used (right?).

oauth2client/service_account.py
```diff
@@ -461,3 +465,94 @@ def create_delegated(self, sub):
...
+    def create_scoped(self, scopes):
+        # Returns an OAuth2 credentials with the given scope
+        return self._creds.create_scoped(scopes)
+
+    def create_with_claims(self, claims):
+        additional_claims = self._claims.copy()
+        additional_claims.update(claims)
+        return JWTAccessCredentials(self._creds, additional_claims=additional_claims)
+
+    @classmethod
```
nathanielmanistaatgoogle: Please please please no more `classmethod`s in this codebase please. What's stopping this from being a module-scope function?

oauth2client/service_account.py
```diff
@@ -461,3 +465,94 @@ def create_delegated(self, sub):
...
+            'exp': _datetime_to_secs(expiry),
+            'iss': self._creds._service_account_email,
+            'sub': self._creds._service_account_email
+        }
+        payload.update(self._claims)
+        if additional_claims is not None:
+            payload.update(additional_claims)
+        jwt = crypt.make_signed_jwt(self._creds._signer, payload,
+                                    key_id=self._creds._private_key_id)
+        return (jwt, expiry)
```
nathanielmanistaatgoogle: Omit these unnecessary parentheses.

oauth2client/service_account.py
```diff
@@ -461,3 +465,94 @@ def create_delegated(self, sub):
...
+
+    @classmethod
+    def from_json(cls, json_data):
+        json_data = json.loads(_from_bytes(json_data))
+
+        service_account_creds = ServiceAccountCredentials.from_json(json_data['creds'])
+        return cls(service_account_creds, additional_claims=json_data['claims'])
+       
+    def to_json(self):
+      json_data = {}
```
nathanielmanistaatgoogle: Favor construct-whole-value over construct-empty-value-and-then-mutate. In this case:

```
json_data = {
    'creds': json.loads(<expression>),
    'clams': self._clams,
}
```

Mmm, clams... :fork_and_knife:

oauth2client/service_account.py
nathanielmanistaatgoogle, line 488: I know you are, and the exception that I'm trying to carve out is for references to issues.

I agree that bare `TODO`s are bad. I agree that `TODO(username)` is bad. I hope I can persuade you to love `TODO(issue reference)` as just a little frosting on the cake of the world-class issue tracker.

Imagine someone coming to this code and just seeing something that looks sub-optimal. They might go to the issue tracker and file an issue about it. They might draft a pull request. They might go to the issue tracker and search and find our issue about it. But by cross-referencing our issue here in the code with a `TODO(issue <number>)`, we get to tell them that we already know about the problem and have a work item filed (and _filed in the right place_) for it.

Perhaps another part of the trick to liking these is that they're just pointers and they don't take away from the issue tracker being the primary store of work to be done. A bare `TODO` or a `TODO(<username>)` is a problem because it's the only record of work to be done, and it's not indexed or landed on anyone's product roadmap or anything; it's more or less lost until it's rediscovered.

Is that any help?

oauth2client/service_account.py
dhermes: or just `token, _`

nathanielmanistaatgoogle: `_` is allowed; I counsel you to use `unused_<helpfully descriptive name>` as a kindness to the poor soul in the future learning the codebase.

oauth2client/service_account.py
kpayson64: - These credentials need to be a GoogleCredentials because one of the "Usable Auth" requirements is that `GoogleCredentials.get_application_default()`  will return a GoogleCredentials object that uses JWT authentication by default if service account credentials are found on the host.
- Not really.  In order to get GoogleCredentials that actually work, we need to pass in an 'aud' claim.  This will mean having an extra optional parameter `additional_claims` in the `get_access_token()` (and possibly `authorize()` method).  One of the suggestions was to add that parameter to all GoogleCredentials, but I'm hesitant to do that.
- Concrete: (get_access_token, create_scoped, create_scoped_required, revoke, to_json, _refresh)
- Abstract: None (All abstract methods are made concrete higher up)
- All of them.

oauth2client/service_account.py
```diff
@@ -461,3 +465,94 @@ def create_delegated(self, sub):
...
+
+
+class JWTAccessCredentials(GoogleCredentials):
+    """Self signed JWT credentials.
+
+    Makes an assertion to server using a self signed JWT from service account
+    credentials.  These credentials do NOT use OAuth 2.0 and instead authenticate
+    directly.
+    """
+    MAX_TOKEN_LIFETIME_SECS = 3600
```
nathanielmanistaatgoogle: Does this constant have to be at class-scope? Can it not be at module-scope? Does it have to be at public visibility? Is it intended for use by users outside the library?

oauth2client/service_account.py
nathanielmanistaatgoogle: Does this class have to be at public visibility? Would it be enough just to have a module-scope `jwt_access_credentials(service_account_creds, additional_claims=None)` function that returned an instance of this class?

kpayson64: I'm fine with hiding this class.

nathanielmanistaatgoogle: Great; please do so. :+1:

oauth2client/service_account.py
kpayson64: from_json is inherited.  If we don't override it, `JWTAccessCredential.from_json(jwt_.to_json())` would fail.  We can remove this if we are hiding the class though.

oauth2client/service_account.py
dhermes: @nathanielmanistaatgoogle Why hide the definition of a class you intend people to have instances of? I could see if it was going to be an underlying property of some other object, but this is a top-level

``` python
creds = JWTAccessCreds(...)
# Use creds for literally my entire program
```

nathanielmanistaatgoogle: Well, that's exactly what I'm waffling on and trying to understand. Is it part of the required developer experience of the feature being implemented that code and developers outside the `oauth2client` library be aware of a JWT access credentials being a kind of thing that is different from a `GoogleCredentials`? Do we support users being able to look at them and examine them and discover "oh, this first one is a `GoogleCredentials` and I can do some things with it but this second one is a `JWTAccessCredentials` and I can do everything with it that I could with the first plus more"?

Or is the situation more like "I need a `GoogleCredentials`, I have a `ServiceAccountCredentials` and some `additional_claims`; ooh look here's a function that I can call that will give me what I need from what I have!"?

To directly answer the question "Why hide the definition of a class [of which] [one] intend[s] people to have instances?": it's appropriate when one is (1) writing a library for a large user base to be supported for a long period of time, and a new public class would increase the support load going forward and (2) the class being proposed doesn't extend the public API of the type according to which its users would use it.

Is that the case here? Can users do anything with a `JWTAccessCredentials` that they couldn't do with a `GoogleCredentials`? Does `JWTAccessCredentials` afford any more methods than `GoogleCredentials`?

kpayson64: From an API perspective, the new functionality they have with `JWTAccessCredentials` is `create_with_claims()`.  One argument for hiding this class for now is that there is no public documentation on using JWTs to access GCE resources.  I'm not sure if JWTAccessCredentials will provide access to everything OAuth2 ServiceAccountCredentials provide access to.

oauth2client/service_account.py
kpayson64: I've made it private for now.  The user shouldn't have to worry about this value.

oauth2client/service_account.py
theacodes: I'm with @dhermes, I'd prefer this not be private. If I understand correctly, grpc will likely want to check if they have an instance of this class?

oauth2client/service_account.py
```diff
@@ -461,3 +465,94 @@ def create_delegated(self, sub):
...
+
+    def create_scoped_required(self):
+        return True
+
+    def create_scoped(self, scopes):
+        # Returns an OAuth2 credentials with the given scope
+        return self._creds.create_scoped(scopes)
+
+    def create_with_claims(self, claims):
+        additional_claims = self._claims.copy()
```
nathanielmanistaatgoogle: Prefer `dict(<dictionary expression>)` over `<dictionary expression>.copy()`.

oauth2client/service_account.py
nathanielmanistaatgoogle: @jonparrott: Expound on that? When would gRPC want to do that, and what action would it take based on the outcome of the check?

@kpayson64: Does `create_with_claims(claims)` have to be an instance-scope method of a new public class? Would it be possible to write a function `create_with_claims(google_credentials, claims)` that affords the behavior? (If the respective answers are "yes" and "no" that goes a long way toward justifying the new class' existence.)

oauth2client/service_account.py
soltanmm: Just a sidenote: AFAICT the stigma towards class-scope attributes stem from them being overridable (or surprisingly not) in subclasses/instances-in-a-way, making things confusing in the long run by paradoxically breaking encapsulation. Or that they represent a 'second' way of doing things. Or... other style reasons.

Kind of a silly thought given how we've all been taught, "Encapsulation! Scopes! Classes!" since day one, but... ¯\_(ツ)_/¯

oauth2client/service_account.py
theacodes: From what I understand, gRPC would call `get_access_token()` directly and provide the `claims` argument.

kpayson64: `create_with_claims()` has no meaning for most google credentials, so I would like to avoid a factory method `create_with_claims(google_credentials, claims)`.  However, ServiceAccountCredentials actually use claims when obtaining an oauth2 access token, so I think it would be reasonable to have a  `create_with_claims(service_account_credentials, claims)``, and have JWTAccessCredentials extend ServiceAccountCredentials.

kpayson64:
I've implemented the feedback and added unit tests.

One of the main changes is that ```JWTAccessCredentials``` now inherits ```ServiceAccountCredentials```.  It is a natural extension, and makes it more reasonable to hide this class if we so choose.

kpayson64:
@jboeuf 
I'm looking to bootstrap an OpenAPI JSON/REST API that supports JWT authentication on GCE to have as a proof of concept for the httplib2 use case of this library.  Is there any guides/examples you are aware of?

jonparrott:
@kpayson64 as an aside, you should probably avoid writing new stuff with httplib2 unless absolutely necessary.

kpayson64:
@jonparrott 
Is the following use case of this library deprecated?

```
http = httplib2.Http()
creds.authorize(http)
```

If so, is there any expected use case of JWT tokens outside of gRPC I should consider?

jonparrott:
@kpayson64 no, but in general we want to move away from httplib2 (or rather, support urllib3) as soon as possible. If you're writing a net-new project and want thread-safety and proper https, urllib3 is a much more sane way to go. see #128 and gcloud-python#1364.

nathanielmanistaatgoogle:
I think I really like the API change in the current draft - we're adding one new method to one existing class and that's it for changes to the public surface area of the entire library?

Is this the shape this change should have going forward or are there still design pressures pushing this in the direction of "no, a new public class is the right thing for this feature"?

kpayson64:
@jonparrott 
Sounds good, I can try to produce a POC using urllib3.

A side note, AFAICT credentials are not thread safe, and I assumed this was because the underlying httplib2 is not thread safe.  Should the new code be thread safe?

jonparrott:
Credentials are thread-safe as long as their storage is thread-safe.

kpayson64:
It seems like we have converged on an API.  I think the one unresolved issue is if JWTAccessCredentials should be public.  It should be a trivial change to make JWTAccessCredentials public in the future, so I think we should leave it private for now. 

If we are comfortable with this API, could @jonparrott @nathanielmanistaatgoogle do a final review?

oauth2client/client.py
```diff
@@ -598,19 +598,9 @@ def new_request(uri, method='GET', body=None, headers=None,
 
             # Clone and modify the request headers to add the appropriate
             # Authorization header.
-            if headers is None:
-                headers = {}
-            else:
-                headers = dict(headers)
+            headers = self._initialize_headers(headers)
```
nathanielmanistaatgoogle: The semantics here are slightly changed, aren't they? In the code on the left side of the diff, the adjusted user-agent value is not included in the headers passed to `self.apply`, but in the code on the right it is. Is this a concern?

oauth2client/client.py
```diff
@@ -844,6 +834,19 @@ def _generate_refresh_request_headers(self):
 
         return headers
 
+    def _initialize_headers(self, headers):
```
nathanielmanistaatgoogle: Please place this definition earlier in the file ("above") any of the places where it is used.

nathanielmanistaatgoogle: Does this behavior have to be an instance-scope method that plays a role in the inheritance chain? It barely uses its `self` parameter. Can it be a library-private helper function?

oauth2client/service_account.py
```diff
@@ -426,6 +432,32 @@ def create_scoped(self, scopes):
...
+        Args:
+            claims: dict, key-value pairs for claims.
+
+        Returns:
+            ServiceAccountCredentials, a copy of the current service account
+            credentials with updated claims to use when obtaining access tokens.
+        """
+        new_kwargs = dict(self._kwargs)
+        new_kwargs.update(claims)
+        result = self.__class__(self._service_account_email,
```
nathanielmanistaatgoogle, line 450: `self.__class__` is a little surprising here - I guess I thought this would have thought this would always return instances of one specific class. Am I just missing the larger picture?

kpayson64, line 450: For ServiceAccountCredentials(Oauth2), `create_with_claims()` should provide additional claims when attempting to get an access token.  For JWT service account credentials, `create_with_claims()` should directly add additional claims to the AccessJWT.

In essence, a ServiceAccountCredentials and JWTAccessCredentials are both allowed to specify additional claims, but the behavior is different.

oauth2client/client.py
kpayson64: Good catch.  I doubt this is an issue, but we should avoid potentially breaking changes when possible.

kpayson64:
I have addressed the feedback.

oauth2client/client.py
```diff
@@ -1523,8 +1539,9 @@ def _get_application_default_credential_from_file(filename):
             token_uri=GOOGLE_TOKEN_URI,
             user_agent='Python client library')
     else:  # client_credentials['type'] == SERVICE_ACCOUNT
-        from oauth2client.service_account import ServiceAccountCredentials
-        return ServiceAccountCredentials.from_json_keyfile_dict(
+        from oauth2client.service_account import _JWTAccessCredentials
+
```
nathanielmanistaatgoogle: Drop this blank line.

nathanielmanistaatgoogle:
Change content looks good!

Please squash commits.

Please [compose a commit message](http://chris.beams.io/posts/git-commit/#seven-rules) with at least a paragraph explaining what's going on. Is this a Google-proprietary feature? Is this a Google-specific feature? Is there a publicly available open and unencumbered standard being implemented in this change that you can cite and to which you can link?

kpayson64:
I've squashed the commits and updated the commit message.

nathanielmanistaatgoogle:
I'm satisfied; @dhermes, @jonparrott, and @jboeuf: please either also make a final review or explicitly decline to do so?

dhermes:
I decline :grinning: I trust you

jonparrott:
I'm satisfied.

jboeuf:
LGTM. Thanks! Please CC me on the PR that will do the integration with grpc.

nathanielmanistaatgoogle:
Woohoo!

kpayson64:
@jonparrott 
What is the process for releasing a new version?

jonparrott:
I probably won't have time to cut a release this week. @nathanielmanistaatgoogle may be able to.

jonparrott:
I'd like to see #506 merged before we release.

ensonic:
Can you make _JWTAccessCredentials public? Or how else would/show I do something like below. Please also not the hack to set 'aud'.

```python
from oauth2client.service_account import _JWTAccessCredentials as JWTAccessCredentials
import httplib2

httplib2.debuglevel = 1

credentials = JWTAccessCredentials.from_json_keyfile_name(
    '/path/to/service-account-credentials.json', scopes=['https://www.googleapis.com/auth/datastore'])
# the class method does not pass through he kwargs to __init__()
# this needs to match the service name in the api-config:
credentials._kwargs['aud'] = 'myapi.endpoints.myproject.cloud.goog'

http_auth = credentials.authorize(httplib2.Http(disable_ssl_certificate_validation=True))
resp, content = http_auth.request('https://example.com/v1/myapi', 'GET')
```

Issue #494, opened 2016-04-08
Refactor Django ORM


waprin:
I was hoping to refactor some of django_orm into django_util. This also ties into #319. There are currently 3 classes:

1) CredentialsField -  It makes sense to me for this to exist if you want to store the oauth credential as part of some sort of UserModel. You sometimes interact with the credential directly, and  you might want to use it to do an oauth call in a background process outside of the session context, for example, in which case you want a credential in the database rather then only in a session.

2) Storage - This is a storage class that uses the CredentialField as the storage object. This will be one of the Storage options for our helpers to use.

The one that I think we should maybe get rid of is 

3) FlowField - Back when people had to roll their own auth flow, this would be one way to store the Flow between the user being redirected to the authorization server and the callback being called with the code to obtain the credential. But now we are providing the code to do do that for you, and using the session seems pretty reasonable given the entire flow should happen within the context of a single session. The only counter-example I can think of is if you are trying to use cookie-based sessions and can't fit the credential into the session (this is currently biting us for our Flask example app, but I think the solution is just to use server-side sessions). Even if we want to keep it, we should probably just make it a configurable option for our code to use the ORM to store the flow rather then the session. 

My inclination is just to remove FlowField but I'm not sure if people use it. I guess it comes down to whether people are still rolling their own flows and if we need to still support those use cases, even though we're now providing the view logic to do it for them.

What do people think?

dhermes:
I am all for removing code. That entire file seems ill-conceived. I had a hard time figuring out many of the design decisions when I tried to test it a while back.

jonparrott:
@waprin can we close this in favor of the new plan for Django stuff?

waprin:
This is basically the tracking issue for that work, although the scope of changes is somewhat bigger then just killing that field.

nathanielmanistaatgoogle:
Retitle this issue to better summarize the intended improvements?

waprin:
Retitled, basic plan is:
- [ Create usage docs to explain the two main use cases (oauth2 (Google) based user system vs adding oauth2 credentials into existing Django user system. ] 
- [Kill FlowField, Move CredentialFiield into django_util/models.py]
- [Keep existing decorator usage exactly as is, add a new config option that lets you specify a Model with a Django UserFIeld and CredentialField, and make the decorator store credentials in that field (where request.user matches the model's user) ]
- [Figure out best place to add complete canonical samples for both main use cases, either directly in this repo or more likely linked to]

[...]

PR #506, opened 2016-05-18
django 1.10 fix

theromis:
```python
/usr/local/lib/python2.7/dist-packages/six.py:808: RemovedInDjango110Warning: SubfieldBase has been deprecated. Use Field.from_db_value instead.
  return meta(name, bases, d)

/usr/local/lib/python2.7/dist-packages/six.py:808: RemovedInDjango110Warning: SubfieldBase has been deprecated. Use Field.from_db_value instead.
  return meta(name, bases, d)
```

### DIFF
--- a/oauth2client/contrib/django_orm.py
+++ b/oauth2client/contrib/django_orm.py
@@ -31,7 +31,7 @@ from oauth2client.client import Storage as BaseStorage
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 
 
-class CredentialsField(six.with_metaclass(models.SubfieldBase, models.Field)):
+class CredentialsField(models.Field):
 
     def __init__(self, *args, **kwargs):
         if 'null' not in kwargs:
@@ -40,6 +40,11 @@ class CredentialsField(six.with_metaclass(models.SubfieldBase, models.Field)):
 
     def get_internal_type(self):
         return 'TextField'
+
+    def from_db_value(self, value, expression, connection, context):
+        if value is None:
+            return None
+        return pickle.loads(base64.b64decode(smart_bytes(value)))
 
     def to_python(self, value):
         if value is None:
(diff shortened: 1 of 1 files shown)


theromis:
I signed it!

nathanielmanistaatgoogle:
I don't know Django well enough to review this. @dhermes or @jonparrott, might either or both of you look it over?

dhermes:
@waprin Has been our go-to Django person

nathanielmanistaatgoogle:
Of course! @waprin, I apologize.

waprin:
LGTM. This PR drops support for 1.7 in favor of the development version, which seems reasonable given that 1.7 is EOL. Created #508 to make sure we explicitly document which versions of Django we support (1.8+ after this). 

@theromis I am curious if you are actually using FlowField or if the errors from it were just breaking things, because based on #494 I was planning on deleting it.

theromis:
@waprin checked for FlowField, don't see it in my code, only CredentialsField

jonparrott:
@waprin do you feel like any tests need to be updated for this?

waprin:
Yes, good point, Since we're still heavily mocking they wont' be the most useful tests but might as well keep it at 100%. @theromis let me know if you can write them, otherwise I will.

theromis:
@waprin I can write them, just give me template, from scratch is hard to do.

waprin:
https://github.com/google/oauth2client/blob/master/tests/contrib/test_django_orm.py is the current tests, just add methods to TestCredentialsField and TestFlowField that calls `from_db_value` and verifies it deserializes correctly. We do branch coverage as well so you'll need to test the functions for both of the conditional values of the if statement. You can run the test with `tox -e py27` from the root directory, run the linter with check out the tox file if you want to see how it's getting run. Once it's all done, just plain `tox` should also pass i.e. run all the environments including python3.4 and the coverage should be 100%. Feel free to ask any more questions, or if you give up I can finish it up for you.

jonparrott:
Closing this as it's obsoleted by #516

wasabigeek:
I was trying to follow the documentation here https://developers.google.com/api-client-library/python/guide/django, but it seems like it's out of date now. Are there plans to update it?

jonparrott:
@waprin do you want to take point on updating the docs?

PR #510, opened 2016-05-20
Support token_uri and revoke_uri in ServiceAccountCredentials

thobrla:
This change adds support for custom authorization token
and revocation URIs in the various ServiceAccountCredentials
factory methods.

It also makes the factory method from_p12_keyfile_contents
public for use by clients who have already loaded the file.


oauth2client/service_account.py
```diff
@@ -81,6 +81,10 @@ class ServiceAccountCredentials(AssertionCredentials):
                    service account.
         user_agent: string, (Optional) User agent to use when sending
                     request.
+        token_uri: string, URI for token endpoint. For convenience defaults
+                   to Google's endpoints but any OAuth 2.0 provider can be
+                   used.
+        revoke_uri: string, URI for revoke endpoint.
```
nathanielmanistaatgoogle: Why not describe this parameter as verbosely as the first?

oauth2client/service_account.py
```diff
@@ -145,14 +152,20 @@ def _to_json(self, strip, to_serialize=None):
             strip, to_serialize=to_serialize)
 
     @classmethod
-    def _from_parsed_json_keyfile(cls, keyfile_dict, scopes):
+    def _from_parsed_json_keyfile(cls, keyfile_dict, scopes,
+                                  token_uri=GOOGLE_TOKEN_URI,
```
nathanielmanistaatgoogle: Optional keyword parameters in private behaviors always seem weird to me because all the call sites are completely under the author's control. Should these really be optional? At how many call sites are they passed? At how many call sites are they not passed?

oauth2client/service_account.py
```diff
@@ -176,18 +189,26 @@ def _from_parsed_json_keyfile(cls, keyfile_dict, scopes):
...
         """Factory constructor from JSON keyfile by name.
 
         Args:
             filename: string, The location of the keyfile.
             scopes: List or string, (Optional) Scopes to use when acquiring an
                     access token.
+            token_uri: string, URI for token endpoint. For convenience defaults
+                       to Google's endpoints but any OAuth 2.0 provider can be
+                       used.
+            revoke_uri: string, URI for revoke endpoint.
```
nathanielmanistaatgoogle: Why not document this as verbosely as the other parameter being introduced in this change?

oauth2client/service_account.py
```diff
@@ -200,17 +221,25 @@ def from_json_keyfile_name(cls, filename, scopes=''):
...
 
         Args:
             keyfile_dict: dict-like object, The parsed dictionary-like object
                           containing the contents of the JSON keyfile.
             scopes: List or string, (Optional) Scopes to use when acquiring an
                     access token.
+            token_uri: string, URI for token endpoint. For convenience defaults
+                       to Google's endpoints but any OAuth 2.0 provider can be
+                       used.
+            revoke_uri: string, URI for revoke endpoint.
```
nathanielmanistaatgoogle: Here too.

nathanielmanistaatgoogle:
Who wants to lead review? @dhermes? @jonparrott? Should we be reticent about promoting `ServiceAccountCredentials._from_p12_keyfile_contents` to public visibility?

dhermes:
@nathanielmanistaatgoogle You get the prize (as the official official maintainer).

No need to make `_from_p12_keyfile_contents` public though, maybe @thobrla has a reason I haven't though of?

oauth2client/service_account.py
thobrla: Here and below, this is copy/paste for consistency from the parent class [AssertionCredentials](https://github.com/google/oauth2client/blob/e54e28939cf985e4d25ef34f96162abd3ab65002/oauth2client/client.py#L1594)

oauth2client/service_account.py
thobrla: They're always passed; it seems more convenient to be able to change the default in fewer places, should that ever need to happen.  I don't feel strongly about it, though.

thobrla:
The argument for publicizing from_p12_keyfile_contents is that currently if I have already read the file (say for the purposes of determining whether it's a JSON file or not), there is no need for oauth2client to re-open the file

thobrla:
ping @nathanielmanistaatgoogle (this request blocks a few others in my pipeline)

dhermes:
RE: 
> The argument for publicizing from_p12_keyfile_contents is that currently if I have already read the file (say for the purposes of determining whether it's a JSON file or not), there is no need for oauth2client to re-open the file

A user can just use `from_p12_keyfile_buffer` and put the already read contents in a `StringIO` buffer.

thobrla:
> A user can just use from_p12_keyfile_buffer and put the already read contents in a StringIO buffer.

Done.

thobrla:
FYI, I need to amend this to take token/revoke URIs from a JSON key file to fix https://github.com/google/oauth2client/issues/513.  The new behavior will be:
1. If the user specifies a token/revoke URI in arguments, use that.
2. Else if the key file specifies token/revoke URI, use that.
3. Else use the Google defaults.

thobrla:
Amendments complete, this now fixes https://github.com/google/oauth2client/issues/513

thobrla:
Ping - this should be ready to go.

jonparrott:
@nathanielmanistaatgoogle is out on vacation.

@thobrla I'd still like to see the comments around the `revoke_uri` parameter address. Please completely document or cross-ref that argument.

After that, this change LGTMs. However, I'd like to have @dhermes take a look before I merge if he has the time.

thobrla:
PTAL

Made `revoke_uri` comments more verbose.

oauth2client/service_account.py
```diff
@@ -172,22 +188,41 @@ def _from_parsed_json_keyfile(cls, keyfile_dict, scopes):
         private_key_pkcs8_pem = keyfile_dict['private_key']
         private_key_id = keyfile_dict['private_key_id']
         client_id = keyfile_dict['client_id']
+        if not token_uri:
+            if 'token_uri' in keyfile_dict:
```
theacodes: Why not `token_uri = keyfile_dict.get('token_uri', GOOGLE_TOKEN_URI)`?

oauth2client/service_account.py
```diff
@@ -172,22 +188,41 @@ def _from_parsed_json_keyfile(cls, keyfile_dict, scopes):
...
+                token_uri = keyfile_dict['token_uri']
+            else:
+                token_uri = GOOGLE_TOKEN_URI
+        if not revoke_uri:
+            if 'revoke_uri' in keyfile_dict:
```
theacodes: Same here.

jonparrott:
Two more comments. If @dhermes doesn't object by EOD today, I'll merge once those two comments are addressed.

oauth2client/service_account.py
thobrla: Done - thanks, much cleaner.

oauth2client/service_account.py
thobrla: Also done.  Thanks.

dhermes:
I don't have much bandwidth but this looks mostly OK.

jonparrott:
Thanks, @dhermes

PR #516, opened 2016-06-03
django 1.10 fix

waprin:
@theromis hope you don't mind but I just wrote the tests for you, good call on reusing to_python. Also added to the docs that we are only supporting versions of Django that Django supports.


oauth2client/contrib/django_orm.py
```diff
@@ -16,6 +16,12 @@
 
 Utilities for using OAuth 2.0 in conjunction with
 the Django datastore.
+
+Note that only supported versions of Django are supported
```
theacodes: This sentence reads a little funny, how about just "Only Django versions 1.8+ are supported."?

oauth2client/contrib/django_util/__init__.py
```diff
@@ -20,6 +20,11 @@
 if the user has authorized, and helper shortcuts to create the authorization
 URL otherwise.
 
+Note that only supported versions of Django are supported
```
theacodes: Same here.

oauth2client/contrib/django_orm.py
waprin: done

oauth2client/contrib/django_util/__init__.py
waprin: done

nathanielmanistaatgoogle:
Tell me about the scary flaming red `cla: no` label on this pull request?

dhermes:
@nathanielmanistaatgoogle It's because the Google CLA checker doesn't like PRs by committee (see the "committed with"):

screen_shot_001

jonparrott:
Yeah it's just googlebot being obtuse. @waprin just needs to confirm things are good.

PR #517, opened 2016-06-06
Handle OSError cases in multistore_file

thobrla:
This change fixes an exception in multistore_file that occurs when
the lock file is under conection. Windows may raise an OSError in
this case, where other platforms typically raise in IOError.


jonparrott:
This LGTM.

PR #522, opened 2016-06-08
Add instructions for updating docs

### LINKED ISSUE #521: Docs for doc gen
The process for generating docs isn't documented anywhere in `CONTRIBUTING.md`

elibixby:
Fixes #521

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,9 +26,27 @@ requests. Fork the repository, and make your changes in the forked repository.
 Include tests
 -------------
 
-Be sure to add the relevant tests before making the pull request. Docs will be
-updated automatically when we merge to `master`, but you should also build
-the docs yourself via `tox -e docs` and make sure they're readable.
+Be sure to add the relevant tests before making the pull request.
+
+Update Docs
+-----------
+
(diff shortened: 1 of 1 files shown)


dhermes:
`tox -e docs` is the "correct" / "endorsed" way to build docs

The GAE SDK dependency is a nagging issue that @jonparrott was tackling in a more generic way

jonparrott:
We can use `gcp-repo-tools` like python-docs-samples: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/nox.py#L87

dhermes:
:+1: I didn't realize you had reached a stable state with that

elibixby:
Just FYI, my confusion was that you had to commit after running `tox -e docs` which isn't mentioned anywhere in `CONTRIBUTING.md`

dhermes:
You don't have to commit, you just need to add an RST file documenting new modules. `tox -e docs` nudges in this direction in case the humans forget to add the documentation

elibixby:
You at the very least have to `git add` new files generated by `tox -e docs` which wasn't documented either.... haven't tested `git add`ing without committing.

PR #524, opened 2016-06-16
Populate scopes for gce.AppAssertionCredentials

### LINKED ISSUE #523: Remove http call in `gce.AppAssertionCredentials.service_account_email`
Only `refresh` should issue HTTP calls in credentials classes.

The credentials should be considered invalid until `refresh()` is called, and accessing `service_account_email` should return `None` until they are refreshed. This is in line with other classes' treatment of `scopes`.

(This is assigned to me, but @elibixby will primarily be doing this work)

elibixby:
Additionally provides support for custom GCE service accounts though an `email` kwarg, and fixes #523


elibixby:
@jonparrott I seem to have fouled up the tree somehow. Not sure why this isn't based cleanly on metadata, (which should be even with master)...

jonparrott:
@elibixby can you give me access to your fork? I can fix the tree.

elibixby:
@jonparrott you should already have it!

jonparrott:
@elibixby tree fixed, please verify. You'll probably need to delete your local branch and pull the new version of it.

oauth2client/contrib/gce.py
```diff
@@ -65,24 +61,41 @@ def __init__(self, scope='', **kwargs):
...
-        data = json.loads(_from_bytes(json_data))
-        return AppAssertionCredentials(data['scope'])
+        raise NotImplementedError(
+            'Cannot serialize credentials for GCE service accounts.')
+
+    def to_json(self):
+        raise NotImplementedError(
+            'Cannot serialize credentials for GCE service accounts.')
+
+    def _retrieve_scopes(self, http_request):
```
theacodes: Would this be better named `_retrieve_info`?

oauth2client/contrib/gce.py
```diff
@@ -65,24 +61,41 @@ def __init__(self, scope='', **kwargs):
...
+        """Validates invalid service accounts by retrieving service account info.
+
+        Args:
+            http_request: callable, a callable that matches the method
+                          signature of httplib2.Http.request, used to make the
+                          request to the metadata server
+        """
+        if self.invalid:
+            info = _metadata.get_service_account_info(
```
theacodes, line 114: exceptions?

oauth2client/contrib/gce.py
elibixby: `_retrieve_scopes` is an override of `Credentials._retrieve_scopes` which is called by `Credentials.retrieve_scopes` which references `httplib2` specifically, so this seemed like the best place to hook in.

oauth2client/contrib/gce.py
elibixby, line 114: Do implicitly thrown exceptions go in doc strings? Seems strange.

theacodes, line 114: I'm just wondering if this method should handle exceptions, or, if the call to it at line 113 should wrap exceptions with `HttpAccessTokenRefreshError`?

oauth2client/contrib/gce.py
theacodes: Does `AssertionCredentials` have `_retrieve_scopes`? Either way, I'd prefer this to just be called `_retrieve_info`, it's a bit more descriptive.

elibixby: `AssertionCredentials` doesn't override `_retrieve_scopes`. So I think the alternative would be to override the external method `Credentials.retrieve_scopes(self, http)` which takes an `httplib2.Http` style client (i.e. one that has a `request` member), and then make that call out to an internal `_retrieve_info` method. Thoughts?

theacodes: Yes. Let's do that.

oauth2client/contrib/gce.py
elibixby, line 114: Since it's not refreshing the token I think that error might be misleading. If we want an exception other than the `httplib2.HttpLib2Error` thrown by `_metadata` I would think we should provide it in `_metadata` rather than here, since it would presumably be specific to the metadata server being inaccessible. 

Or we could make a specific error for fetching service account info, but I'm not sure what info that provides.

oauth2client/contrib/gce.py
elibixby: Done.

oauth2client/contrib/gce.py
theacodes, line 114: So I guess what i'm saying/looking for is that `refresh()` on other credentials only raises `AccessTokenRefreshError` (iirc), so, wouldn't an exception raised by `_retrieve_info` in `_refresh` be surfaced to the user as an `AccessTokenRefreshError`?

elibixby, line 114: The `_retrieve_info` call in `_refresh` is outside the `try... except` block that wraps with an `AccessTokenRefreshError` intentionally, since I didn't think that a failure to fetch service account info should register as a failure to refresh a token... I'm still confused as to what is wrong with this...

elibixby, line 114: Resoved!

tests/contrib/test_gce.py
```diff
@@ -77,46 +92,50 @@ def test_refresh_token_failed_fetch(self):
...
     def test_sign_blob_not_implemented(self):
         credentials = AppAssertionCredentials([])
         with self.assertRaises(NotImplementedError):
             credentials.sign_blob(b'blob')
 
     @mock.patch('oauth2client.contrib._metadata.get_service_account_info',
-                return_value={'email': 'a@example.com'})
-    def test_service_account_email(self, metadata):
+                return_value=SERVICE_ACCOUNT_INFO)
+    def test_retrieve_scopes(self, metadata):
```
theacodes, line 116: should the be `test_retrieve_info` now?

oauth2client/contrib/gce.py
```diff
@@ -55,8 +52,7 @@ class AppAssertionCredentials(AssertionCredentials):
     information to generate and refresh its own access tokens.
```
theacodes, line 52: Can you add a note in the docstring about these credentials needing to be refreshed before usage?

theacodes, line 52: On that note, did we require these to be refreshed before usage before? Can you verify that simple calls with google api python client still work as expected with this?

jonparrott:
@elibixby getting close. 👍

tests/contrib/test_gce.py
elibixby, line 116: Testing the externally surfaced function `gce.AppAssertionCredentials.retrieve_scopes` seems preferable to testing an class private function?

theacodes, line 116: Fair enough.

oauth2client/contrib/gce.py
elibixby, line 52: Sure. We didn't previously require that these credentials be refreshed as we never returned `service_account_email` as `None`. But that was seen as preferable to instantiating an `httplib2.Http` client as part of the `service_account_email` method.

theacodes, line 52: Yeah, but I'm wondering if that actually changed any real behavior. Hence trying it out with google-api-python-client

elibixby, line 52: Yep yep! =)

elibixby, line 52: On that note, is there a place where we can put/run integration/system tests?

elibixby, line 52: At the very least is there a way I can put a test in tests/contrib and then tell travis to ignore it?

elibixby, line 52: Yep, confirmed that list GCE instances with both ADC and directly instantiated `gce.AppAssertionCredentials`

elibixby, line 52: Little gist that works on GCE

https://gist.github.com/elibixby/40c07cd47df15f7b0f1db0f575518058

elibixby, line 52: Fudge, tested on the wrong branch. Got some errors.

elibixby, line 52: Alright, that gist now works, and has expected behavior w.r.t scopes and email.

theacodes, line 52: Great.

oauth2client/contrib/gce.py
```diff
@@ -53,10 +50,14 @@ class AppAssertionCredentials(AssertionCredentials):
     This credential does not require a flow to instantiate because it
     represents a two legged flow, and therefore has all of the required
     information to generate and refresh its own access tokens.
+
+    Note that credentials.service_account_email and credentials.scopes
```
theacodes: use `:attr:`service_account_email``, etc.

jonparrott:
I'm pretty satisfied with this. @nathanielmanistaatgoogle do you want to take a look?

elibixby:
@jonparrott @nathanielmanistaatgoogle friendly ping, I know y'all are busy.

jonparrott:
I'll give @nathanielmanistaatgoogle until tomorrow afternoon to respond before merging, and we can always address any post-merge comments.

@elibixby feel free to nag me tomorrow to merge.

elibixby:
@jonparrott Ping to merge =)

PR #529, opened 2016-06-27
Adding exception text to exit message for invalid clientsecrets

pferate:
This is to address a problem found in google/google-api-python-client/issues/228.

This will give users the text from the InvalidClientSecretsError exception, in addition to the generic message that it currently does.  The format of the message can be modified, if needed.


pferate:
Found a problem to fix in one of the tests.  Removing this for now.

PR #530, opened 2016-06-27
Adding exception text to exit message for invalid clientsecrets

pferate:
This is to address a problem found in google/google-api-python-client#228.

This will give users the text from the InvalidClientSecretsError exception, in addition to the generic message that it currently does. The format of the message can be modified, if needed.


oauth2client/client.py
```diff
@@ -2210,8 +2210,10 @@ def flow_from_clientsecrets(filename, scope, redirect_uri=None,
                 client_info['client_id'], client_info['client_secret'],
                 scope, **constructor_kwargs)
 
-    except clientsecrets.InvalidClientSecretsError:
+    except clientsecrets.InvalidClientSecretsError as e:
         if message:
+            if str(e):
+                message = 'Error: %s\n================\n%s' % (str(e), message)
             sys.exit(message)
```
theacodes, line 2217: Holy cow, I'm really appalled we have a `sys.exit` hiding in here. D:

oauth2client/client.py
```diff
@@ -2210,8 +2210,10 @@ def flow_from_clientsecrets(filename, scope, redirect_uri=None,
                 client_info['client_id'], client_info['client_secret'],
                 scope, **constructor_kwargs)
 
-    except clientsecrets.InvalidClientSecretsError:
+    except clientsecrets.InvalidClientSecretsError as e:
         if message:
+            if str(e):
```
theacodes: `if e.args`?

oauth2client/client.py
```diff
@@ -2210,8 +2210,10 @@ def flow_from_clientsecrets(filename, scope, redirect_uri=None,
...
+                message = 'Error: %s\n================\n%s' % (str(e), message)
```
theacodes: I'd prefer just `'The client secrets were invalid: \n {}'.format(e)`

jonparrott:
Thanks for this contribution. Do you mind adding or augmenting the existing test for this?

oauth2client/client.py
pferate: Sounds better.  What about if message is passed as a parameter?

oauth2client/client.py
pferate: better.  I'll update it.

theacodes: I'm unsure about py2.6 / py3 compatibility here so tread carefully.

oauth2client/client.py
theacodes: Works just as well.

oauth2client/client.py
pferate: Just tested this, it works in at least 2.6, 2.7, 3.2, 3.3, 3.4, and 3.5.

theacodes: Thanks.

oauth2client/client.py
```diff
@@ -2210,8 +2210,10 @@ def flow_from_clientsecrets(filename, scope, redirect_uri=None,
                 client_info['client_id'], client_info['client_secret'],
                 scope, **constructor_kwargs)
 
-    except clientsecrets.InvalidClientSecretsError:
+    except clientsecrets.InvalidClientSecretsError as e:
         if message:
```
nathanielmanistaatgoogle: While we're here: `if message is not None:`.

pferate: will do.

pferate:
@jonparrott @nathanielmanistaatgoogle, I updated this PR again and added a test for this (slightly modified a previous test).  Travis passed on my branch, but still waiting to launch for this PR.

I had to use positional argument specifiers in the `.format()` string to have it work with 2.6.

jonparrott:
Great, this looks good to me. @nathanielmanistaatgoogle to do the final review.

nathanielmanistaatgoogle:
Change content looks good; [please amend your commit log message to meet the guidelines](http://chris.beams.io/posts/git-commit/#seven-rules).

pferate:
Commit message updated.

jonparrott:
Waiting for travis.

jonparrott:
Thanks for your contributions, @pferate!

PR #531, opened 2016-06-28
Adding Python 3.5 capability

pferate:
In Python 3.5, `http.HTTPStatus` was added, previously HTTP statuses were int.
Casting `http.HTTPStatus` as int to maintain backwards maintainability.


.travis.yml
```diff
@@ -1,6 +1,12 @@
 language: python
 python: 2.7
 sudo: false
+# A hack to install 3.5, until Travis-CI includes it
```
nathanielmanistaatgoogle: `# TODO(https://github.com/google/oauth2client/issues/xxx): <text>`?

nathanielmanistaatgoogle:
[Does the body of your commit message wrap at seventy-two characters](http://chris.beams.io/posts/git-commit/#seven-rules)?

pferate:
I was using 80 characters, since it looked like a few previous commits were using that.  I'll switch it to 72.

I'll also add an issue, then reference it in the `.travis.yml` file.

tests/test_file.py
```diff
@@ -151,13 +151,13 @@ def test_token_refresh_store_expires_soon(self):
         access_token = '1/3w'
         token_response = {'access_token': access_token, 'expires_in': 3600}
         http = HttpMockSequence([
-            ({'status': str(http_client.UNAUTHORIZED)},
+            ({'status': str(int(http_client.UNAUTHORIZED))},
```
pferate, line 154: These will pass all of the tests if I just use `int` instead of `str`, but I think that it is supposed to be an actual string, so that is why I'm casting as int, then as str.

.travis.yml
theacodes: Just create the issue, no need to add a todo in the code as well. :)

tests/test_file.py
theacodes, line 154: makes sense.

.travis.yml
nathanielmanistaatgoogle: I actually favor the issue-linked TODO here. [When a strange thing appears in code, it's nice for there to be an issue tracking making it unstrange, but it's even nicer if it includes a link to that issue in its strangeness](https://github.com/google/oauth2client/pull/503#discussion_r61331868).

pferate: Issue created (#532).  I can easily change it when I fix the commit messages.

pferate:
Changes made, just waiting for Travis.

jonparrott:
Thanks, this looks good to merge once travis passes.

Issue #465, opened 2016-03-17
contrib.appengine uses its own InvalidClientSecretsError


jonparrott:

pferate:
I don't see any reason within `oauth2client` to not change it to use `clientsecrets.InvalidClientSecretsError`.

I'm not sure if any major projects are trying to catch it.

I can make a PR for this, and leave it up to you guys on when to merge it in.

The other exception in that module is `InvalidXsrfTokenError(Exception)`, and is not used outside of it.  Should I just leave that as is, or have it extend a local base error for the module?

jonparrott:
originally backwards compatibility, but I'm okay with breaking changes as long as they're in a major release.

jonparrott:
Closed by #533

PR #533, opened 2016-06-29
Remove unneeded contrib.appengine exceptions

pferate:
Resolving issue #465.

Possible breaking change, so probably wait to add in next major release.


oauth2client/contrib/appengine.py
```diff
@@ -89,11 +89,11 @@ def _safe_html(s):
...
 
 
-class InvalidClientSecretsError(Exception):
-    """The client_secrets.json file is malformed or missing required fields."""
+class Error(Exception):
+    """Base error for this module."""
 
 
-class InvalidXsrfTokenError(Exception):
+class InvalidXsrfTokenError(Error):
```
theacodes: Seems this is only used in this class during the oauth2callback. Can we just remove it?

jonparrott:
@nathanielmanistaatgoogle interested to see your thoughts on this breaking change.

oauth2client/contrib/appengine.py
```diff
@@ -89,11 +89,11 @@ def _safe_html(s):
     return cgi.escape(s, quote=1).replace("'", '&#39;')
 
 
-class InvalidClientSecretsError(Exception):
-    """The client_secrets.json file is malformed or missing required fields."""
+class Error(Exception):
+    """Base error for this module."""
```
nathanielmanistaatgoogle: What's the reason for this?

tests/contrib/test_appengine.py
```diff
@@ -923,7 +921,7 @@ def test_decorator_from_client_secrets_bad_type(self):
         with loadfile_patch as loadfile_mock:
             loadfile_mock.return_value = ('badtype', None)
             self.assertRaises(
```
nathanielmanistaatgoogle: While the patient is on the table: how about

```
with self.assertRaises(clientsecrets.InvalidClientSecretsError):
    OAuth2DecoratorFromClientSecrets(<arguments>)
```

?

nathanielmanistaatgoogle:
I like the change, mostly. What valid use cases would we be breaking if we were to just incorporate this now? Do we think that developers have written code that catches `appengine.InvalidClientSecretsError` and does something else in response? Note that without "Raises:" in the constructor's doc string, I'm hesitant to call the raising of the exception part of the API in the first place.

jonparrott:
> Do we think that developers have written code that catches appengine.InvalidClientSecretsError and does something else in response?

Probably not, this is likely something experienced when first developing and is uncaught and just manually addressed, if I had to guess. It's not exactly a "recoverable" exception.

> I'm hesitant to call the raising of the exception part of the API in the first place.

Same, but even still we'll probably make the next release 3.0.0.

tests/contrib/test_appengine.py
pferate: Sounds good.  That's how I would do it if I were rewriting it.

There are a lot more instances of this type of syntax in the file, so it may be another PR for that.  Should I go through and make similar changes for all of the test files?  I'm thinking yes.

oauth2client/contrib/appengine.py
pferate: I figured that it may be a good idea to have a module base error to inherit from versus the base python `Exception`.  With only one other module exception, it may be considered overkill.  If there were multiple module exceptions, I think it would be best to have a module base exception.

There are also a few examples of this throughout the rest of the code base.

tests/contrib/test_appengine.py
theacodes: > There are a lot more instances of this type of syntax in the file, so it may be another PR for that. 

please do.

> Should I go through and make similar changes for all of the test files? I'm thinking yes.

Don't sweat it too much. But if you're really bored, I won't stop you.

(One day I'll probably switch this project to py.test)

oauth2client/contrib/appengine.py
theacodes: Yeah, it's overkill. I wouldn't create custom exceptions unless they are critically useful to users.

pferate:
After doing some testing, this may not be considered a breaking change after all.

Since we are importing `InvalidClientSecretsError` in the `appengine` module it can still be caught as `appengine.InvalidClientSecretsError`, even though it is really `clientsecrets.InvalidClientSecretsError`.  Unless users are specifically treating them differently (which isn't likely), the change should be fairly transparent.

jonparrott:
Cool. I'll likely make the next release 3.0.0 anyways because of the other changes.

pferate:
clientsecrets.py
```python
class InvalidClientSecretsError(Exception):
    """Main error"""
```
appengine.py
```python
class InvalidClientSecretsError(Exception):
    """Duplicated error"""
```
appengine2.py
```python
from clientsecrets import InvalidClientSecretsError
```

```python
>>> import appengine
>>> appengine.InvalidClientSecretsError
<class 'appengine.InvalidClientSecretsError'>
>>> import appengine2 as appengine
>>> appengine.InvalidClientSecretsError
<class 'clientsecrets.InvalidClientSecretsError'>
```

oauth2client/contrib/appengine.py
pferate: Are you talking about removing the `InvalidXsrfTokenError`? or keeping it's base class as `Exception`?

theacodes: Removing it.

pferate: I'm thinking we may want to keep it, although I'm not the expert on it.  How else would you want to handle invalid XSRF tokens?

theacodes: So from what I can tell, this is only used in this class. It's raised in the OAuth2 callback which is a request handler, so the user of this library never really gets a chance to catch it - it likely bubbles up as a top-level exception.

I'd recommend we excise it and deal with the bad token the same way as flask util: 
https://github.com/google/oauth2client/blob/master/oauth2client/contrib/flask_util.py#L410

(As an aside, can you file a bug to split the webapp2 parts of this into its own module?)

pferate: I think we can just catch the exception within this module and send the appropriate error response.
https://github.com/google/oauth2client/blob/master/oauth2client/contrib/appengine.py#L816-817

``` python
redirect_uri = _parse_state_value(str(self.request.get('state')), user)
```

change to:

``` python
try:
    redirect_uri = _parse_state_value(str(self.request.get('state')), user)
except InvalidXsrfTokenError:
    self.response.out.write('The authorization request failed: Invalid CSRF Token')
    return # or redo some other logic, so that self.redirect isn't executed below
```

or something along those lines.

theacodes: Yeah, that's what i'm saying, but since InvalidXsrfTokenError is only ever used here, it might as well just be a ValueError.

pferate: Alright, makes sense.  I'll go back and revise this PR to a generalized exception refactoring.  Then file the few other issues that came up, and the associated PRs.

If I do the PRs right away, should I still file the issues separately?

theacodes: Re: filing issues - it's up to you.

oauth2client/contrib/appengine.py
nathanielmanistaatgoogle: ["Modules or packages should define their own domain-specific base exception class, which should inherit from the built-in `Exception` class. The base exception for a module should be called `Error`."](https://google.github.io/styleguide/pyguide.html?showone=Exceptions#Exceptions) is a terrible practice and I've been trying for years to get it removed from the style guide. Thank you for not following it. :-)

I apologize that there are parts of the codebase that define such nonsense per-module base exception classes.

oauth2client/contrib/appengine.py
```diff
@@ -477,15 +469,15 @@ def _parse_state_value(state, user):
         user: google.appengine.api.users.User, The current user.
 
     Raises:
-        InvalidXsrfTokenError: if the XSRF token is invalid.
+        ValueError: if the XSRF token is invalid.
```
nathanielmanistaatgoogle: `Raises: ValueError` is always a code smell because a system should insist on only being used properly and should never describe or guarantee how it will behave when used improperly. Can the user know whether or not the XSRF token is valid before calling any of the system entry points that propagate down into calls to this function? Can the user be held responsible for only making calls that will propagate down to this function in such a way that `ValueError` will never be raised? Does the raising of `ValueError` here represent a programming defect on the part of the user of `oauth2client`?

pferate: This is done behind the scenes during the OAuth callback.  If the client and server tokens differ, the user should would just see the `request failed` response, and that 's all that's being checked here.

I'm of the opinion of keeping the `InvalidXsrfTokenError` for clarity, but @jonparrott said that since that exception is only used internally, to not have a separate custom exception, which also makes sense.

In this case, it is not the result of a programming defect or user error, but rather, a potential CSRF/XSRF vulnerability.

nathanielmanistaatgoogle: `ValueError` indicates that someone made a programming mistake (either the authors of the library that raises the `ValueError` or the users of the library that raises the `ValueError`, depending on context). If the exception being raised here does not indicate a programming defect, we should use some exception other than `ValueError`. This is a non-public method so what is raised from it needn't be a public exception type, but it's starting to sound like it should be a custom exception type.

pferate: Reverted back to `InvalidXsrfTokenError`

nathanielmanistaatgoogle: Well I think the question of whether or not `InvalidXsrfTokenError` should continue to exist has not yet been answered. It is a module-scope module-public code element, but it is not raised by any part of the public API of the module. The smell I'm getting is that it should be either privatized or dropped. @jonparrott, am I making sense?

pferate: Sorry, I saw that part about using a custom exception (which I'm leaning towards), but overlooked the privatized part.  I'm thinking to keep it at module-scope, but private (`appengine._InvalidXsrfTokenError`)?  The method that calls it is module-scope and private.

theacodes: > The smell I'm getting is that it should be either privatized or dropped. @jonparrott, am I making sense?

Right.

I'd prefer if the xsrf check function didn't raise an exception but rather returned True of False. Exceptions are for exceptional circumstances.

pferate: Would it be acceptable to have `_parse_state_value()` return `None` for invalid tokens and a string otherwise?  Similar to `oauth2client.contrib.flask_util._get_flow_for_token()`

theacodes: Yep. In my opinion `_parse_state_value()` shouldn't be validating the state value, just returning it. Let the callback handler deal with the token validation, just like in `flask_util`: https://github.com/google/oauth2client/blob/master/oauth2client/contrib/flask_util.py#L410

pferate: One difference is that `flask_util` stores the server XSRF token in a session, whereas `appengine` regenerates the token during validation.

oauth2client/contrib/appengine.py
```diff
@@ -476,18 +468,14 @@ def _parse_state_value(state, user):
...
     uri, token = state.rsplit(':', 1)
-    if not xsrfutil.validate_token(xsrf_secret_key(), token, user.user_id(),
-                                   action_id=uri):
-        raise InvalidXsrfTokenError()
-
-    return uri
+    if xsrfutil.validate_token(xsrf_secret_key(), token, user.user_id(),
+                               action_id=uri):
+        return uri
+    return
```
pferate: Explicitly returning None, because it has a special meaning here.

pferate:
Updated `_parse_state_value()` to return `None` for invalid tokens and the `uri` otherwise

oauth2client/contrib/appengine.py
nathanielmanistaatgoogle: Make it explicitly explicit by including the `None` in `return None`. Also put it in an `else:` to visually mark its exclusivity with the other branch of the `if`.

jonparrott:
@nathanielmanistaatgoogle does this look good to merge?

pferate:
Hold on @jonparrott , I need to add in @nathanielmanistaatgoogle's last suggestions.  I've been working on moving this to py.test today (you mentioned that at the meetup last month)

pferate:
Ok.  I amended the commit to include explicitly returning `None`.

nathanielmanistaatgoogle:
@jonparrott: to me this now looks ready to merge.

jonparrott:
Woot. Waiting on ~~Godot~~Travis.

jonparrott:
Thanks, @pferate this is a great change. :D

PR #537, opened 2016-07-05
Use context manager for assertRaises

pferate:
Using context manager for assertRaises calls.  Also needed to update files using `unittest` to `unittest2` to be able to use context managers in Python 2.6 (was added to `unittest` in 2.7).  Keeping the same import and usage syntax as the other files using `unittest2`.


tests/contrib/test_multistore_file.py
```diff
@@ -25,7 +25,7 @@
 
 from oauth2client import util
 from oauth2client.client import OAuth2Credentials
-from oauth2client.contrib import locked_file
+from oauth2client.contrib.locked_file import CredentialsFileSymbolicLinkError
```
pferate: FYI: I adjusted the import syntax to keep line 190 under 80 characters.

tests/contrib/test_multistore_file.py
```diff
@@ -186,9 +187,8 @@ def test_multistore_no_symbolic_link_files(self):
             'user-agent/1.0',
             ['some-scope', 'some-other-scope'])
         try:
-            self.assertRaises(
-                locked_file.CredentialsFileSymbolicLinkError,
-                store.get)
+            with self.assertRaises(CredentialsFileSymbolicLinkError):
```
pferate: Was running over 80 characters here, so I had to adjust this.  I figured that breaking this `assertRaises` line into multiple lines could be a little confusing, so instead, I adjusted the way this is imported. (importing just the exception instead of the module)

scripts/run_gce_system_tests.py
```diff
@@ -53,4 +53,4 @@ def test_token_info(self):
 
 
 if __name__ == '__main__':
```
theacodes, line 55: Might as well go ahead and remove these clauses since you're here. They're not needed.

tests/contrib/test_appengine.py
```diff
@@ -922,11 +923,10 @@ def test_decorator_from_client_secrets_bad_type(self):
             'oauth2client.contrib.appengine.clientsecrets.loadfile')
         with loadfile_patch as loadfile_mock:
             loadfile_mock.return_value = ('badtype', None)
-            self.assertRaises(
-                AppEngineInvalidClientSecretsError,
-                OAuth2DecoratorFromClientSecrets,
-                'doesntmatter.json',
-                scope=['foo_scope', 'bar_scope'])
+            with self.assertRaises(AppEngineInvalidClientSecretsError):
+                OAuth2DecoratorFromClientSecrets('doesntmatter.json',
```
theacodes: prefer to break after the opening `(` to avoid these scrunched-up hanging indents.

tests/contrib/test_multistore_file.py
theacodes: Why import the error directly? Typically we just import the module.

tests/contrib/test_multistore_file.py
theacodes: Oh, I see. Hmm... I think I'd prefer you to break assertRaises into multiple lines.

tests/contrib/test_multistore_file.py
pferate: Forgot about that part of the `imports` language rule.  Fixed in new commit.

tests/contrib/test_multistore_file.py
pferate: Fixed.  There wasn't a similar example in the Google Style Guide, so I went with the second acceptable example in the [PEP8 guidelines](https://www.python.org/dev/peps/pep-0008/#indentation).

scripts/run_gce_system_tests.py
pferate, line 55: Fixed as a new commit in this PR.

tests/contrib/test_appengine.py
pferate: Fixed.

jonparrott:
@pferate this is almost ready, only one outstanding comment (about the `if __name__` stuff)

pferate:
Is there something I missed?  I took care of the `if __name__` stuff in 06e7032.

I only left those checks in the `/scripts/*` files.

jonparrott:
Oh, I see, my apologies.

@nathanielmanistaatgoogle I'm merging this tests-only change. Let me know if you have any post-merge concerns.

@pferate thanks for doing this! Much appreciated!

pferate:
No problem @jonparrott!

PR #539, opened 2016-07-07
Use the vendor BIOS file to determine if we are on GCE.

Galabar001:
We currently use a connection test to the GCE metadata server to
determine if we are on a GCE instance.  This can be flaky and may
take a long time.  Instead, we can simply check for "Google" in
the vendor BIOS file.


Galabar001:
This is a proposed change to how we determine if we are on a GCE instance.

Galabar001:
It looks like this will need a little work.  However, I would welcome comments on the feasibility of this change.

jonparrott:
Will this work in docker? E.g., Container Engine or App Engine Flex?

I would be for this change as a short-circuit for the metadata server check.

Galabar001:
Let me look into that.

Galabar001:
A short-circuit -- that sounds like a good idea.  Going to the metadata server check if either (a) the vendor file doesn't exist or (b) does not have the value 'Google' might be a safer first version.

Galabar001:
This is closer, but I'd like to update test_client.py as well.

nathanielmanistaatgoogle:
1. Please squash commits with each round of code review back-and-forth.
2. Ensure that your commit has both an author and committer who have signed the CLA. This pull request currently has the "cla: no" label on it; please get that fixed (probably by ensuring that your commit has your corporate account listed for both author and committer).
3. Include in your commit message at least a paragraph of details answering questions like:
  * Where does Google guarantee that GCE instances will have this file with these contents? And for how far into the future does Google guarantee this?
  * Where does Google guarantee that other computing environments that aren't GCE will not have this file with these contents?

jonparrott:
> Please squash commits with each round of code review back-and-forth.

@nathanielmanistaatgoogle we can do that at the end when we merge, no need for them to squash ahead of time anymore. :)

jonparrott:
@Galabar001 you're a googler, can you make sure your commits are authored with your @google.com address and that you've registered your github username at go/github?

Galabar001:
Jon -- it looks like go/github knows me. :)

Nathaniel -- thank you for the pointers.  I'm still talking to the Vanadium team to determine of the BIOS vendor will only ever be a cloud thing.  If, for example, Google created the BIOS/motherboard for a chromebook, we could get ourselves into trouble.

Another option we are discussing is using an environment variable that could be set by clients (like gcloud).

Galabar001:
I think I may withdraw this pull request for now.  I'm not able to get over the nagging feeling that this mechanism may break at some point in the future.

Let me try the environment variable approach and send out another pull request.

PR #547, opened 2016-07-15
Migrate to Pytest

pferate:
Migrating from `unittest2` to `pytest`.  Some more work could still be done to clean up the fixtures; splitting some into smaller, reusable parts and maybe move them to a separate file and just import them in as needed.

When updating the assert statements, I found errors in some of the tests, mainly when checking for True of False.  `unittest` would incorrectly pass cases like these:

``` python
self.assertTrue('this is a str, not a boolean')
self.assertFalse(None)
self.assertFalse('')
```


waprin:
Much bigger fan of pytest than unittest but nervous about merging this with #546 :-\

jonparrott:
@pferate I definitely appreciate the effort but this is a huge change that will likely require a lot of discussion. I definitely want to punt on this until after 3.0.0.

Once that's released we can take a look at this. I don't even know if it's feasible for this library as it has to play nicely with Google's internal build system (I'll need @nathanielmanistaatgoogle to chime in there).

If we do decide to do this, I'd really like to do this in stages. First stage is to just swap nose with pytest's runner. Second, move to plain asserts. Finally, remove all unit test subclasses and refactor into fixtures.

pferate:
I'd be okay with rebasing and updating this PR after yours gets merged in.

pferate:
> If we do decide to do this, I'd really like to do this in stages. First stage is to just swap nose with pytest's runner. Second, move to plain asserts. Finally, remove all unit test subclasses and refactor into fixtures.

If you notice my 3 commits, that's basically how I did it.

jonparrott:
Great, although it would probably be easier for the reviewers to do it as
three PRs.

nathanielmanistaatgoogle:
@jonparrott: to every reasonable extent we don't adjust the content of this open source library to meet the needs of Google's internal tools.

@pferate: let's imagine for a moment that I were a person of limited vision and narrow mind and had only ever heard stirrings of `pytest` but never actually worked with it. What would be the reason for this migration? There's no attached issue; why is this pull request a good thing? What if I really like my tests with their `assertParticularThing` methods?

jonparrott:
@nathanielmanistaatgoogle I'm definitely a proponent of `py.test`. We switched python-docs-samples over to it and it has made a significant difference in the size and readability of our tests. Notably, using plain asserts, function-based fixtures instead of `setUp` and `tearDown`, no strange camelCase methods, and not having to use classes to organize tests are all huge wins for readability and long-term maintainability.

That said, I'm unsure if it makes sense to switch a well-established library such as this to py.test.

However, I would like to note that nose is essentially unmaintained. Pytest's test runner is compatible with nose and unittest-style tests, so I think *at minimum* we should switch to the py.test runner.

jonparrott:
@nathanielmanistaatgoogle 

> to every reasonable extent we don't adjust the content of this open source library to meet the needs of Google's internal tools.

Yes, but I want to make sure you won't run into trouble running the test suite with blaze.

jonparrott:
I'm closing this PR because this project is now deprecated. Please see the [release notes for v4.1.0](https://github.com/google/oauth2client/releases/tag/v4.1.0) for more details.

PR #551, opened 2016-07-18
Fix Flask Autodoc

### LINKED ISSUE #550: Sphinx Autodoc Breaking on flask_util
Haven't totally narrowed down what Sphinx is doing, but it seems like it's trying to autodocument `current_app` which it can't import outside of an app context. Causing Travis to fail on #547 and #546. 

Removing the line      :undoc-members: from docs/source/flask_util.rst fixes it but I don't thikn this is what we want.

waprin:
Fixes #550


jonparrott:
LGTM, thanks for fixing this, Bill.

Issue #554, opened 2016-07-19
Plan to Implement: oauth2client.transport module


dhermes:
FYI @nathanielmanistaatgoogle @jonparrott @waprin I am going to factor out all of the usage of `httplib2` into a standalone module so we can have a clean break when the time comes to use a better maintained library.

I plan to start soon to please weigh in ASAP with any comments or concerns or requests for API surface area.

Also see #128 for a more "detailed" (but old) plan.

nathanielmanistaatgoogle:
No concerns. I think I've heard rumblings about this for a while.

txomon:
Does this mean we will be able to use our own http clients? I am mostly interested on using aiohttp to do async requests.

This sounds like having the functionality of the oauth2 steps separated from the implementation of the transport, but I don't know if it supports the use case I just mentioned.

jonparrott:
@txomon I'm not sure about async transports, but this will certainly make something like that a lot more feasible in the future.

txomon:
Some difficulties I have found until now is that the docs say that
implementing request function is enough but doesn't give a signature nor
clear signature.

Making that a little more flexible is highly appreciated =) Thanks!

jonparrott:
That's the goal. Ultimately we want to support urllib3/requests as the primary transport and deprecate httplib2. To do that, we more or less have to enable users to bring their own http client. Two birds, one stone.

dhermes:
@txomon There are a few more details in #128. The goal is to define an interface that a BYO transport implementation needs to comply with and then just support that. We'll probably have to support the BYO stuff side-by-side with the default `httplib2` stuff for a generation or two.

dhermes:
Calling this complete and moving discussion back to #128

Issue #558, opened 2016-07-20
Coveralls report not being used


dhermes:
As a result coverage has dropped below 100% and sadness for me since #212 was so hard.

See an [example fail][1] when trying to upload the coverage report to coveralls.io:

```
  File "/home/travis/build/google/oauth2client/.tox/coveralls/lib/python2.7/site-packages/nosegae.py", line 91, in configure
    import dev_appserver
ImportError: No module named dev_appserver
```

[1]: https://travis-ci.org/google/oauth2client/jobs/146207674

dhermes:
@jonparrott Do you know if `/home/travis/.cache/google_appengine` needs to be populated in each of the environments? i.e. why does `gae` work just fine but `py27` fails on the coverage report.

Could we "fix" this with a simple cheat of moving the coverage report into `gae` from `py27`?

dhermes:
OK as can be seen in #560 this does fix the import issue and brings back the `coveralls` report.

As seen in the report https://coveralls.io/builds/7092809 we have 97.06% coverage in `oauth2client.contrib.sqlalchemy and 99.37% in `oauth2client.contrib.appengine`.

It looks like there's a branch mass in `contrib.appengine` in the 807-809, 806->807 range and some other missed in `contrib.sqlalchemy` around 150, 144->150, 146->148, 162->165.

jonparrott:
Well the good news is we're not significantly below 100%.

dhermes:
Indeed. Any idea who the owners should be to get back to 100%?

jonparrott:
I can probably handle it sometime in the next month.

pferate:
One of those was me in #533.  I just wrote a test to get the `appengine` part covered.  Haven't looked at the `sqlalchemy` part yet.

I'll look at that after I finish working on my other PR.

PR #559, opened 2016-07-20
Begin transport module for httplib2 specific pieces.

dhermes:
Towards #554.

### DIFF
--- a/oauth2client/service_account.py
+++ b/oauth2client/service_account.py
@@ -27,14 +27,14 @@ from oauth2client import GOOGLE_REVOKE_URI
 from oauth2client import GOOGLE_TOKEN_URI
 from oauth2client import util
 from oauth2client._helpers import _from_bytes
-from oauth2client.client import _apply_user_agent
-from oauth2client.client import _initialize_headers
 from oauth2client.client import _UTCNOW
 from oauth2client.client import AccessTokenInfo
 from oauth2client.client import AssertionCredentials
-from oauth2client.client import clean_headers
 from oauth2client.client import EXPIRY_FORMAT
 from oauth2client.client import SERVICE_ACCOUNT
+from oauth2client.transport import _apply_user_agent
+from oauth2client.transport import _initialize_headers
+from oauth2client.transport import clean_headers
 
 
 _PASSWORD_DEFAULT = 'notasecret'
--- a/oauth2client/transport.py
+++ b/oauth2client/transport.py
@@ -1,0 +1,198 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+
+import httplib2
+import six
+from six.moves import http_client
+
+from oauth2client._helpers import _to_bytes
+
+
+_LOGGER = logging.getLogger(__name__)
+# Properties present in file-like streams / buffers.
+_STREAM_PROPERTIES = ('read', 'seek', 'tell')
+
+# Google Data client libraries may need to set this to [401, 403].
+REFRESH_STATUS_CODES = (http_client.UNAUTHORIZED,)
+
+
+class MemoryCache(object):
+    """httplib2 Cache implementation which only caches locally."""
+
+    def __init__(self):
+        self.cache = {}
+
+    def get(self, key):
+        return self.cache.get(key)
+
+    def set(self, key, value):
+        self.cache[key] = value
+
+    def delete(self, key):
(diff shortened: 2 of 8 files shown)


oauth2client/transport.py
```diff
@@ -0,0 +1,202 @@
...
+        return self.cache.get(key)
+
+    def set(self, key, value):
+        self.cache[key] = value
+
+    def delete(self, key):
+        self.cache.pop(key, None)
+
+
+def get_cached_http():
```
theacodes, line 48: Is this really needed? Few other transports will provide caching.

jonparrott:
This is a good first step but I'm interested in how you'll make this swappable.

oauth2client/transport.py
dhermes, line 48: Goal of the initial PR is just to move the `httplib2` functionality out of all the modules. Boiling the ocean slowly.

dhermes, line 48: To address the actual question, caching makes sense here (though `MemoryCache` isn't threadsafe so that's stupid). The cache is used for grabbing certs from https://www.googleapis.com/oauth2/v1/certs, which are used a lot for an app that needs to verify ID tokens, but the actual payload itself is rotated infrequently by the backend.

dhermes:
@jonparrott Before making swappable, the goal is to just make the boundary explicit (see the plan in #128).

jonparrott:
Gotcha. Okay. I'm fine with this. Nathaniel to merge.

dhermes:

oauth2client/transport.py
```diff
@@ -0,0 +1,202 @@
...
+    """Creates a copy of the headers.
+
+    Args:
+        headers: dict, request headers to copy.
+
+    Returns:
+        dict, the copied headers or a new dictionary if the headers
+        were null.
+    """
+    if headers is None:
```
nathanielmanistaatgoogle: `return {} if headers is None else dict(headers)`?

oauth2client/transport.py
```diff
@@ -0,0 +1,202 @@
...
+
+def _initialize_headers(headers):
+    """Creates a copy of the headers.
+
+    Args:
+        headers: dict, request headers to copy.
+
+    Returns:
+        dict, the copied headers or a new dictionary if the headers
+        were null.
```
nathanielmanistaatgoogle: `None`. There is no `null` in Python - or am I missing something context-specific here?

oauth2client/service_account.py
```diff
@@ -27,14 +27,14 @@
...
 from oauth2client._helpers import _from_bytes
-from oauth2client.client import _apply_user_agent
-from oauth2client.client import _initialize_headers
 from oauth2client.client import _UTCNOW
 from oauth2client.client import AccessTokenInfo
 from oauth2client.client import AssertionCredentials
-from oauth2client.client import clean_headers
 from oauth2client.client import EXPIRY_FORMAT
 from oauth2client.client import SERVICE_ACCOUNT
+from oauth2client.transport import _apply_user_agent
```
nathanielmanistaatgoogle, line 35: Have we today any better options for avoiding violations of "don't import private members of other modules" than we did when the left-side-of-the-diff code was written?

oauth2client/transport.py
```diff
@@ -0,0 +1,202 @@
...
+    """Adds a user-agent to the headers.
+
+    Args:
+        headers: dict, request headers to add / modify user
+                 agent within.
+        user_agent: str, the user agent to add.
+
+    Returns:
+        dict, the original headers passed in, but modified if the
+        user agent is not null.
```
nathanielmanistaatgoogle: `user agent is not None.`

tests/test_transport.py
```diff
@@ -0,0 +1,131 @@
...
+
+from oauth2client import client
+from oauth2client import transport
+
+
+class TestMemoryCache(unittest2.TestCase):
+
+    def test_get_set_delete(self):
+        cache = transport.MemoryCache()
+        self.assertEqual(None, cache.get('foo'))
```
nathanielmanistaatgoogle: `assertIsNone` here and throughout.

oauth2client/transport.py
dhermes: I just copy-pasted. Will change it.

oauth2client/transport.py
dhermes: You say potato, I say tomato. On it.

nathanielmanistaatgoogle:
Looks mostly good.

oauth2client/transport.py
dhermes: null is me referring to `None` in a language-agnostic way. Changing to `None`

oauth2client/service_account.py
dhermes, line 35: I am going to rip out the equivalent of `wrap_http_for_auth` from `service_account.py` in a future PR. This will solve that problem. On the subject, `clean_headers` should have never been public in `oauth2client.client`.

tests/test_transport.py
dhermes: Was copy-pasta. On it.

dhermes:
@nathanielmanistaatgoogle Addressed the comments I could and gave an answer to the one I couldn't about non-public imports

nathanielmanistaatgoogle:
Looks great; merge (with squashed commits) when tests repass.

dhermes:
See #558 about coverage failing. Not my sword to fall on here. Merging.

PR #560, opened 2016-07-20
Sending the coverage report to coveralls in gae env.

dhermes:
This is because the GAE import failed in the py27 env.


dhermes:
See #558

jonparrott:
This isn't the way I'd like to do this, I'd like to combine the reports from all runs. I'll merge this for now though.

dhermes:
RE: combine from all runs, `tox -e coveralls` still does `py27` and `gae` together, but not the Python 3 reports. Is that what you meant?

PR #561, opened 2016-07-21
Remove httplib2 parts from service_account module.

dhermes:
Towards #554.


jonparrott:
This LGTM, though again interested in how the contract will change - some transports likely won't need to wrap the http object.

PR #563, opened 2016-07-22
Getting code coverage back to 100%

pferate:


pferate:
:100:
For resolving #558

jonparrott:
LGTM, thank you for doing this!

dhermes:
@pferate @jonparrott There are two branch misses in `sqlalchemy` that are not reported in coveralls.io (it's a [frustrating missing feature][1])

I'm trying to find the branches but can't figure out how to generate the HTML coverage report. @jonparrott testing wizard, I can haz help?

[1]: https://github.com/lemurheavy/coveralls-public/issues/31

jonparrott:
@dhermes 

```
$ tox -e cover
$ coverage html
```

Just in case, I uploaded it here: http://temp.theadora.io/oauth2client/oauth2client_contrib_sqlalchemy_py.html

dhermes:
Great, so the branches are [here][1] and [here][2].

I tried `tox -e cover; coverage html` and got "No data to report". Maybe I have a stupid local config?

[1]: https://github.com/google/oauth2client/blob/ae73312942d3cf0e98f097dfbb40f136c2a7c463/oauth2client/contrib/sqlalchemy.py#L148
[2]: https://github.com/google/oauth2client/blob/ae73312942d3cf0e98f097dfbb40f136c2a7c463/oauth2client/contrib/sqlalchemy.py#L164

jonparrott:
Not sure, outdated coverage maybe? Either way, it seems we have what we need. :)

pferate:
I was looking into another issue this morning (#357) and I noticed that `contrib/_win32_opener.py` and `contrib/_fcntl_opener.py` were not in the cover report.  I'm not sure what's going on with that yet.  Looks like some of the calls are being mocked, but I didn't see where it was actually getting tested.

jonparrott:
They're excluded on purpose because they're platform-specific. Either way, don't sweat it because we can actually just remove them now. :)

pferate:
OK.  I kind of figured that for the `win32`, but I thought that `fcntl` would have been covered.

PR #565, opened 2016-07-25
Fix `self.assertTrue(isinstance(` statements

### LINKED ISSUE #542: Replace self.assertTrue(isinstance(<value expression>, <class expression>)) with self.assertIsInstance
Something I noticed in an unrelated pull request.

pferate:
Replace self.assertTrue(isinstance()) with self.assertIsInstance()

Resolves #542


jonparrott:
LGTM, thanks for this.

PR #569, opened 2016-07-26
Migrate test runner to py.test

pferate:
Maintainer of pytest_gae needs to merge open PR and push to PyPI.
In the mean time, get code directly from GitHub mirror.


oauth2client/crypt.py
```diff
@@ -248,3 +237,25 @@ def verify_signed_jwt_with_certs(jwt, certs, audience=None):
...
+
+
+def configure_module():
+    """Configure module-wide variables.
+
+    Not all testing frameworks load modules directly, so we need a way to
+    define these on set up.
+    """
+    global Signer
+    global Verifier
```
dhermes: 👎  on the use of `global`

pferate: They were global before, just being defined at the module level.  Pytest doesn't load module/global variables, which is why it was failing (unless someone knows of a way to do that).

I'll see if I can come up with a cleaner/better way to do that.

I came across [this answer](http://stackoverflow.com/a/7668273/398514) on StackOverflow regarding `__getattr__` at the module level, which referenced a hack that [Guido van Rossum suggested](https://mail.python.org/pipermail/python-ideas/2012-May/014969.html).

pferate:
@jonparrott Should we worry about populating the `conftest.py` file now?  Or should we wait until we start moving everything to fixtures (phase 3)?

oauth2client/crypt.py
dhermes: The issue isn't the use of globals / module level constants, just the use of the `global` keyword. It makes the module harder to understand.

pferate: Guido's hack only works on Python 3, so that is not an option.

I was incorrect in my statement that Pytest didn't load module variables, modules are just reused between tests.  I've added some setUp/tearDown code to do some housecleaning when module variables are changed (at least for `oauth2client.crypt`)

@dhermes Thanks for making me take another look at things!

pferate:
There is a bug in the 0.9 release of `flake8-import-order` (released yesterday) that is causing the `flake8` test to incorrectly fail.

I've already created an [issue](/PyCQA/flake8-import-order/issues/80) for it.

jonparrott:
@pferate if we need app engine stuff to work (which we do) we might as well populate conftest.py just for that. You only need a subset of the stuff in the example I linked.

pferate:
I copied the code from App Engine file you linked.  I moved the imports within `setup_sdk_imports` to within try/except blocks, so that the tests don't fail outside of App Engine.

I just added the fixture code for this PR to `pytest_configure()`.

These will be expanded more when we transition the tests here to fixtures.

tests/conftest.py
```diff
@@ -0,0 +1,149 @@
...
+
+    # Fixes timezone and other os-level items.
+    try:
+        import google.appengine.tools.os_compat
+        (google.appengine.tools.os_compat)
+    except ImportError:
+        pass
+
+
+def import_appengine_config():
```
theacodes: This probably isn't needed for oauth2client.

tests/conftest.py
```diff
@@ -0,0 +1,149 @@
...
+    tb.init_user_stub()
+    tb.init_logservice_stub()
+    # tb.init_taskqueue_stub(root_path='tests/resources')
+    tb.init_taskqueue_stub()
+    tb.taskqueue_stub = tb.get_stub(gaetestbed.TASKQUEUE_SERVICE_NAME)
+
+    return tb
+
+
+def run_taskqueue_tasks(testbed, app):
```
theacodes: This won't be needed here either.

tests/conftest.py
```diff
@@ -0,0 +1,149 @@
...
+        testbed.setup_env(
+            user_email=email,
+            user_id=id,
+            user_is_admin='1' if is_admin else '0',
+            overwrite=True)
+    return _login
+
+
+@pytest.fixture
+def run_tasks(testbed):
```
theacodes: Won't be needed.

tests/conftest.py
```diff
@@ -0,0 +1,149 @@
...
+    return _run_tasks
+
+
+def pytest_configure():
+    """conftest.py hook function for setting up SDK imports."""
+    setup_sdk_imports()
+    util.positional_parameters_enforcement = util.POSITIONAL_EXCEPTION
+
+
+def pytest_runtest_call(item):
```
theacodes: Can remove this.

tests/conftest.py
```diff
@@ -0,0 +1,149 @@
...
+    """py.test fixture for running GAE tasks."""
+    def _run_tasks(app):
+        run_taskqueue_tasks(testbed, app)
+    return _run_tasks
+
+
+def pytest_configure():
+    """conftest.py hook function for setting up SDK imports."""
+    setup_sdk_imports()
+    util.positional_parameters_enforcement = util.POSITIONAL_EXCEPTION
```
theacodes: Add a comment here as to why this is needed?

tox.ini
```diff
@@ -18,74 +18,70 @@ deps = {[testenv]basedeps}
...
-      --logging-level=INFO \
+      --gae-path={env:GAE_PYTHONPATH:google_appengine} \
+      --gae-project-path=tests/data \
       tests/contrib/test_appengine.py \
       tests/contrib/test__appengine_ndb.py
 deps = {[testenv]deps}
     coverage
-    nosegae
+    pytest-cov
+    # pytest_gae  # TODO: Move to official source when it is updated
```
theacodes: With the conftest.py, pytest_gae won't be needed anymore.

tests/conftest.py
pferate: Is this in reference to the `import_appengine_config` function or the try/except blocks above it?

tests/conftest.py
pferate: Similar question here. `run_taskqueue_tasks()` or `tb.` code?

theacodes: this whole function

tests/conftest.py
theacodes: the whole function.

tox.ini
pferate: Where does the reference to `GAE_PYTHONPATH` get used?  Should I add a new fixture to update the `PATH` instead?

theacodes: Probably just do it in setup_sdk_imports. (Alternatively other projects like python-docs-samples has nox/tox do it)

pferate: I ended up updating `PYTHONPATH` in Tox for the `gae` environment.  I'm catching `ImportError` when running `setup_sdk_imports`, which may be a bit too generalized.  If `ImportError` is raised there, I'm assuming that it is not running within the `gae` test and those modules aren't needed.  If, for some reason, we are in the `gae` test and `ImportError` is raised, then the tests should fail.

jonparrott:
Please don't merge this until after the 3.0.0 release. I'll remove the label when it's safe.

jonparrott:
v3.0.0 is released so this can now be consider for merging.

pferate:
Should there be a 4.x or Dev branch, so that they can be kept separate from 3.x updates?

jonparrott:
> Should there be a 4.x or Dev branch, so that they can be kept separate from 3.x updates?

I don't think so, but @nathanielmanistaatgoogle is welcome to overrule that.

tests/conftest.py
```diff
@@ -0,0 +1,106 @@
...
+        # namespace package. This prevents that.
+        reload_module(sys.modules['google'])
+
+    # This sets up google-provided libraries.
+    import dev_appserver
+    dev_appserver.fix_sys_path()
+
+    # Fixes timezone and other os-level items.
+    import google.appengine.tools.os_compat
+    (google.appengine.tools.os_compat)
```
nathanielmanistaatgoogle: I don't understand this at all?

tests/conftest.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def setup_testbed():
+    """Sets up the GAE testbed and enables common stubs."""
+    from google.appengine.datastore import datastore_stub_util
+    from google.appengine.ext import testbed as gaetestbed
+
+    # Setup the datastore and memcache stub.
+    # First, create an instance of the Testbed class.
+    tb = gaetestbed.Testbed()
```
nathanielmanistaatgoogle: Initialisms are discouraged for the same reason that single-letter identifiers are banned. Spell out something descriptive?

tests/conftest.py
```diff
@@ -0,0 +1,106 @@
...
+    # This sets up google-provided libraries.
+    import dev_appserver
+    dev_appserver.fix_sys_path()
+
+    # Fixes timezone and other os-level items.
+    import google.appengine.tools.os_compat
+    (google.appengine.tools.os_compat)
+
+
+def setup_testbed():
```
nathanielmanistaatgoogle: "setup" is a noun; "set up" is a verb phrase. This should be `set_up_testbed`.

nathanielmanistaatgoogle:
Commit log message should be longer and describe why this change is being made.

tests/conftest.py
theacodes: Do you really want to understand App Engine magic?

Or do you mean the usage of `()` here? It just makes the linter ignore an unused import (we're importing for side-effects).

nathanielmanistaatgoogle:
No dev branches. No branches other than `master` and `gh_pages`, actually. We're not that complicated a project.

tests/conftest.py
theacodes: This might be slightly _too_ nitpicky. 😛

tests/conftest.py
theacodes: `testbed` would be a fine name here.

tests/conftest.py
nathanielmanistaatgoogle: Then is there a comment-level indication that can be used instead of a needless expression? `# pylint: disable=unused-import` or something perfectly clear like that?

theacodes: This probably came from a project using flake8 instead of pylint, which has historically had pretty opaque error messages (E103, etc.).

tests/conftest.py
pferate: Sounds good.  I just copied and pasted everything from GCP except for `util.positional_parameters_enforcement = util.POSITIONAL_EXCEPTION`.

tests/conftest.py
pferate: Removed `()` line and added a comment level flag to the import

pferate:
@jonparrott, I've been going over various ways of handling the GAE in this PR and I'd like some feedback.

Initially, I was using a fork of [pytest_gae](https://pypi.python.org/pypi/pytest_gae), due to the main library not working with newer versions of app engine.  Last week I got a PR merged to fix this and the library owner pushed it to PyPI.

You said to just use the code from the App Engine repo, but not all of it was utilized.  I also ran into an issue with the `cover` tox environment; one test expects GAE to be available and the other expects it to not be available.  If we had it always available, there is some trivial code that doesn't execute, because it's expecting the GAE library to not be there. There was no simple way of doing this without reproducing code from a plug-in.  Should I add a CLI option to load GAE (which would be added from plug-in)?  Also, we don't load the `google` module directly, so the line about reloading it always gets skipped.

I also came across another plug-in, (pytest-beds)[https://pypi.python.org/pypi/pytest-beds), which adds the following fixtures:

- testbed
- mailer
- channel
- urlfetch
- memcache
- taskqueue
- blobstore
- ndb
- users

`pytest-beds` seems like the most flexible option to me.

What are your thoughts on going forward with this?

jonparrott:
@pferate I'm going to send a PR to your fork to get the App Engine stuff where I want it.

jonparrott:
@pferate see https://github.com/pferate/oauth2client/pull/2

jonparrott:
I am now okay with this. @nathanielmanistaatgoogle to do the final review and merge.

(also there's a conflict, please fix)

pferate:
Rebased onto master.

Thanks for the help @jonparrott!

tests/contrib/appengine/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+        set_up_gae_environment(config.getoption('gae_sdk'))
+
+
+def pytest_ignore_collect(path, config):
+    """Runs when py.test discovers tests.
+
+    Return True to skip collection. Skips App Engine tests when --gae-sdk
+    is not specified.
+    """
+    if 'contrib/appengine' in str(path):
```
nathanielmanistaatgoogle: Does

```
return 'contrib/appengine' in str(path) and config.getoption('gae_sdk') is None
```

fit on one line?

nathanielmanistaatgoogle:
Looks mostly good; just that one comment.

tests/contrib/appengine/conftest.py
pferate: No.  If we do it that way, it would need to be split into 3 lines:

``` python
    return (
        'contrib/appengine' in str(path) and
        config.getoption('gae_sdk') is None)
```

We could take away the explicit `else`'s, but I know you prefer them.

``` python
    if 'contrib/appengine' in str(path):
        if config.getoption('gae_sdk') is None:
            return True
    return False
```

A one-line `if` statement is one character over the limit as well.

``` python
    if 'contrib/appengine' in str(path) and config.getoption('gae_sdk') is None:
```

Redefining `path` can get it all to fit in 2 lines:

``` python
    path = str(path)
    return 'contrib/appengine' in path and config.getoption('gae_sdk') is None
```

Not explicitly checking for `None` can get it into one line:

``` python
    return 'contrib/appengine' in str(path) and not config.getoption('gae_sdk')
```

nathanielmanistaatgoogle: I like the three-line form best (it most clearly communicates "this is a simple conjuction") but could live with what's currently drafted. How much do you like the three-line form?

(One day an autoformatter will solve this!)

pferate:
@jonparrott, with us moving the app engine tests into their own directory, should we move the app engine modules into their own directory as well?

Another coupled set of files are `contrib/gce` and `contrib/_metadata`.

tests/contrib/appengine/conftest.py
theacodes: This is where we disagree - I don't like combining branching if statements like this unless they're directly related (e.g., `if x > 0 and x < screen.w`). But, I'll defer to you on this.

jonparrott:
> with us moving the app engine tests into their own directory, should we move the app engine modules into their own directory as well?

I don't think so. There's no hard and fast rule that tests must be organized exactly as their module-under-test. I think defaulting to that but making exceptions where it makes testing easier (like we did here) makes sense. Practicality beats purity.

tests/contrib/appengine/conftest.py
pferate: Personally, I would go with:

``` python
    if 'contrib/appengine' in str(path):
        if config.getoption('gae_sdk') is None:
            return True
```

since, in this case, `True` is the only return value that does something special, all other return values are treated the same.

nathanielmanistaatgoogle: I happen to think that `True` and `False` are equally special. Go with the conjunction.

tests/contrib/appengine/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+    import google.appengine.tools.os_compat  # noqa: unused import
+
+
+def pytest_configure(config):
+    """Configures the App Engine SDK imports on py.test startup."""
+    if config.getoption('gae_sdk') is not None:
+        set_up_gae_environment(config.getoption('gae_sdk'))
+
+
+def pytest_ignore_collect(path, config):
```
nathanielmanistaatgoogle, line 49: Am I right to learn from there being no call sites of this function in added code that this, due to its name, is called by py.test? If so, maybe a better doc string for it would be something like `Directs py.test to skip App Engine tests when --gae-sdk is not specified.`?

tests/contrib/appengine/conftest.py
pferate: Will do.

pferate:
@nathanielmanistaatgoogle: PR updated, PTAL.

tests/contrib/appengine/conftest.py
```diff
@@ -0,0 +1,57 @@
...
+    """Direct py.test to skip App Engine tests when --gae-sdk is not specified.
+
```
nathanielmanistaatgoogle: I think lines 51-53 can be dropped; a one-line summary is now fine for this function. "Return True to skip collection" probably belongs in the abstract "super" function, wherever that's defined.

nathanielmanistaatgoogle:
One last tweak; sorry.

tests/contrib/appengine/conftest.py
pferate: Dropped.  I had to reword it because the `"""` made it go over 80 characters.

pferate:
PR updated.  @nathanielmanistaatgoogle PTAL.

nathanielmanistaatgoogle:
What's with that test failure?

Is it desired that this go in as two commits? (No pushing in either direction; just want to confirm.)

pferate:
Looks like it may be related to #590.  This is what is in my Travis logs:
```
# Travis ships with an old version of PyPy, so install at least version 2.6.
if [[ "${TOX_ENV}" == "pypy" ]]; then
    git clone https://github.com/yyuu/pyenv.git ${HOME}/.pyenv
    ${HOME}/.pyenv/bin/pyenv install pypy-2.6.0
fi
fatal: destination path '/home/travis/.pyenv' already exists and is not an empty directory.
```
In my repo, `pypy` passed the first time, but failed when I reran the test.

As for the commits, one was a merged PR from @jonparrott helping me sort out some of the GAE tests, so I didn't want to lose his work within mine.  Functionally, they could be squashed.

jonparrott:
The pypy failure is my fault. We probably just need to check for the non-existence of {$HOME}/.pyenv before doing the git clone.

We can do it in this PR or I can send a separate one.

nathanielmanistaatgoogle:
Separate pull request with the fix in isolation, please.

jonparrott:
@pferate It's fixed, rebase and try again?

pferate:
@jonparrott Rebased and waiting on Travis.

One weird thing that I've been seeing recently is `py33` tests have been [failing in my fork on Travis](https://travis-ci.org/pferate/oauth2client/jobs/150778885), but have been [passing on the main repo](https://travis-ci.org/google/oauth2client/jobs/150778904).  They also pass when I run them on my local system.

The 2 linked Travis jobs are from the same commit from yesterday.  It looks like my fork is collecting the django tests (it shouldn't) and the main repo correctly skips them.  Before I dig much further, do you have any idea of what's going on?

As long as it passes on the main repo, it doesn't really matter; but this seems strange.

nathanielmanistaatgoogle:
@pferate: are you waiting for answers from @jonparrott before this gets merged?

(I am completely fine with the pull request content and metadata; feel free to merge when conversation is again concluded.)

pferate:
No need to wait.  I just had some weird, unique test results that I'm trying to figure out on my end.

PR #572, opened 2016-07-27
Add warnings and helpers for prompt='consent'.

### LINKED ISSUE #453: approval_prompt=force no longer works
After upgrading to 2.0.0 approval_prompt=force no longer works.

The difference in the generated url seems to be the reason

1.5.2:
https://accounts.google.com/o/oauth2/auth?access_type=offline&approval_prompt=force

2.0.0:
https://accounts.google.com/o/oauth2/v2/auth?access_type=offline&approval_prompt=force

I'm not sure where to report this, as the real issue seems to be the new v2 endpoint.

theacodes:
Fixes #453

This is the last one for 3.0.0. After this, I'll send the release PR.

:)

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -1762,6 +1762,33 @@ class DeviceFlowInfo(collections.namedtuple('DeviceFlowInfo', (
         return cls(**kwargs)
 
 
+def _oauth2_web_server_flow_params(kwargs):
+    """Configures redirect URI parameters for OAuth2WebServerFlow."""
+    params = {
+        'access_type': 'offline',
+        'response_type': 'code',
+    }
+
+    params.update(kwargs)
+
+    # Check for the presence of the deprecated approval_prompt param and
+    # warn appropriately.
+    approval_prompt = params.get('approval_prompt')
+    if approval_prompt is not None:
+        logger.warning(
+            'The approval_prompt parameter for OAuth2WebServerFlow is '
+            'deprecated. Please use the prompt parameter instead.')
+
+        if approval_prompt == 'force':
+            logger.warning(
+                'approval_prompt="force" has been adjusted to '
+                'prompt="consent"')
+            params['prompt'] = 'consent'
+            del params['approval_prompt']
+
+    return params
+
+
 class OAuth2WebServerFlow(Flow):
     """Does the Web Server Flow for OAuth 2.0.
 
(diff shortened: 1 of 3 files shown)


oauth2client/client.py
```diff
@@ -1843,6 +1843,22 @@ def __init__(self, client_id,
         }
         self.params.update(kwargs)
 
+        self._check_for_approval_prompt()
+
+    def _check_for_approval_prompt(self):
+        approval_prompt = self.params.get('approval_prompt')
+        if approval_prompt:
```
nathanielmanistaatgoogle: `is not None`? You are comparing to `None`, aren't you?

theacodes: Yes, fixed.

oauth2client/client.py
```diff
@@ -1843,6 +1843,22 @@ def __init__(self, client_id,
         }
         self.params.update(kwargs)
 
+        self._check_for_approval_prompt()
```
nathanielmanistaatgoogle: Do not call instance methods from the constructor of the instance.

`_check_for_approval_prompt` _barely_ uses `self`. I think in this constructor you should just change

```
self.params = {
    'access_type': 'offline',
    'response_type': 'code',
}
self.params.update(kwargs)

self._check_for_approval_prompt()
```

into

```
self.params = _oauth2_web_server_flow_params(kwargs)
```

.

nathanielmanistaatgoogle:
First pass completed.

oauth2client/client.py
theacodes: Cool, so just pull that out into a module-level function? Sounds good.

I did this originally because I didn't feel like this logic belonged in the constructor, so I'm fine with pulling it into a separate function.

theacodes: Done.

tests/test_client.py
```diff
@@ -1679,6 +1679,16 @@ def test_override_flow_via_kwargs(self):
         self.assertEqual(client.OOB_CALLBACK_URN, q['redirect_uri'][0])
         self.assertEqual('online', q['access_type'][0])
 
+    def test__check_for_approval_prompt(self):
+        self.flow.params['approval_prompt'] = 'force'
+        self.flow._check_for_approval_prompt()
```
nathanielmanistaatgoogle: This method no longer exists, right?

theacodes: Yes, sorry, I wasn't as done as I thought. :)

jonparrott:
@nathanielmanistaatgoogle this should be ready for another look. I was going to wait on travis but it seems we've been put back in the slow lane.

jonparrott:
Travis is (finally) green. :)

jonparrott:
Thanks for all the reviews today, @nathanielmanistaatgoogle. I'll be cutting 3.0.0 tomorrow morning.

PR #573, opened 2016-07-27
Fix system test reference to GOOGLE_TOKEN_URI.

theacodes:
We broke the system tests when @pferate fixed the import nonsense. This should get it back in line.

@nathanielmanistaatgoogle I'm merging this, it's a small test-only change and I can't verify it works until it's in master.


jonparrott:
Success, our builds are passing again.

PR #576, opened 2016-07-28
Warn user if storage file is missing

### LINKED ISSUE #338: storage class does not throw exceptions or errors when creds file is missing
Calling `storage = Storage('a_credentials_file')` does not give any errors or warnings when the file passed does not exist. This should be helpful in debugging.

pferate:
Also warn user if the given filename is a directory.

Resolves #338


oauth2client/file.py
```diff
@@ -20,13 +20,20 @@
...
+import warnings
 
 from oauth2client import client
 
 
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 
 
+_IS_DIR_MESSAGE = '{0}: Is a directory'
+
```
theacodes: Why so many newlines?

tests/test_file.py
```diff
@@ -51,12 +53,21 @@ def tearDown(self):
         try:
             os.unlink(FILENAME)
         except OSError:
+            try:
+                os.rmdir(FILENAME)
+            except OSError:
+                pass
```
pferate: These can be moved to a separate fixture once we migrate to fixtures.

oauth2client/file.py
```diff
@@ -42,6 +49,10 @@ def _validate_file(self):
         if os.path.islink(self._filename):
             raise CredentialsFileSymbolicLinkError(
                 'File: {0} is a symbolic link.'.format(self._filename))
+        elif os.path.isdir(self._filename):
```
theacodes: This should actually be exceptional. A `put()` won't be able to write the file.

oauth2client/file.py
pferate: I was going by the blank lines guidence in [PEP8](https://www.python.org/dev/peps/pep-0008/#blank-lines)

> Surround top-level function and class definitions with two blank lines.

But after seeing it explicitly says functions and classes, this may not need to follow that.

pferate:
PR updated.  I moved the directory housekeeping into the test function, similar to the sym link test.  Both of those can be pulled into fixtures when we get to that point.

Also, should I submit an issue about possible removing the "no symbolic link" requirement?  That way we can host any discussion there?

jonparrott:
> Also, should I submit an issue about possible removing the "no symbolic link" requirement? That way we can host any discussion there?

Sounds reasonable.

oauth2client/file.py
```diff
@@ -20,17 +20,28 @@
...
 import threading
+import warnings
 
 from oauth2client import client
 
 
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 
+_SYM_LINK_MESSAGE = 'File: {0}: Is a symbolic link.'
+
```
theacodes: I feel like there really isn't much need for a blank newline between these, but that's a super nit.

oauth2client/file.py
```diff
@@ -20,17 +20,28 @@
...
+_IS_DIR_MESSAGE = '{0}: Is a directory'
+
+_MISSING_FILE_MESSAGE = 'Cannot access {0}: No such file or directory'
+
 
 class CredentialsFileSymbolicLinkError(Exception):
     """Credentials files must not be symbolic links."""
 
 
+class CredentialsFileDirectoryError(Exception):
```
theacodes: Great, another exception class. :P 

Could this at least extend from IOError? @nathanielmanistaatgoogle WDYT?

pferate: I was just trying to be consistent within the file...  IOError sounds reasonable to me.

If I use IOError, and we remove the Symbolic Link parts, then there will only be 2 real Exceptions defined outside of `client` (in the core library).

nathanielmanistaatgoogle: Subclass `IOError`? Why not simply use `IOError`?

The circumstance at play here is a programming defect, right? If this exception (or these exceptions) are ever raised in the wild, no user of `oauth2client` is going to modify their code to handle the exception, right? They're just going to say "oh, my code has a bug" and they're going to fix it?

Why not just use `IOError` to communicate the problem? It's the exception's message and not its class that seems to be doing all the communication anyway?

tests/test_file.py
```diff
@@ -74,11 +77,23 @@ def _create_test_credentials(self, client_id='some_client_id',
             user_agent)
         return credentials
 
-    def test_non_existent_file_storage(self):
+    @mock.patch('warnings.warn')
+    def test_non_existent_file_storage(self, warn_mock):
         s = file.Storage(FILENAME)
```
nathanielmanistaatgoogle: Single-letter identifiers are banned by the style guide.

tests/test_file.py
```diff
@@ -74,11 +77,23 @@ def _create_test_credentials(self, client_id='some_client_id',
...
         credentials = s.get()
+        warn_mock.assert_called_with(
+            file._MISSING_FILE_MESSAGE.format(FILENAME))
         self.assertEquals(None, credentials)
```
nathanielmanistaatgoogle: `assertIsNone`.

nathanielmanistaatgoogle:
Commit message is stale, but no need to worry about updating it until we've found consensus.

jonparrott:
> Commit message is stale, but no need to worry about updating it until we've found consensus.

& we can always re-write it when we merge.

tests/test_file.py
pferate: That syntax is used throughout this file.  Should I update them all in this PR, even though they aren't related to this change?

oauth2client/file.py
pferate: Both of these may be end-user bugs.  If they specify a directory instead of a file, or if they try to get credentials from a file that doesn't exist.  I can see the argument of raising `IOError` in both cases.

I suppose it depends on how you want to define the specification of `Storage.get()`.  Right now, the specification of `Storage.get` says that it only returns `Credentials`; nothing about any exceptions or `None`.  It isn't my decision on which way to go.

If we go the exceptions route, we could have a base exception, like `client.NoCredentialsFound`, and change `Storage.get` to analyze the return value of `locked_get()` and raise an exception if it's `None`.  For the cases in this PR, we could an exception that subclasses `NoCredentialsFound` and `IOError`.  This way any existing classes that extend `Storage` can return `None` (which will raise `NoCredentialsFound` in `Storage.get()`) or raise their own exception that extends `NoCredentialsFound`.

Or we could just go with the `None` route and change any exceptions to `return None` and leave it up to the end-user to investigate.  In that case, we could still issue warnings.

theacodes: I greatly prefer not raising an exception for `get()` not finding credentials. Based on the usage of this library, that is not an error. The usage pattern mostly seen is:

```
credentials = storage.get()
if not credentials():
    # obtain them somehow.
    storage.put(credentials)
```

theacodes: Based on that usage, `get()`ing credentials from a non-existent file is also _not_ an error.

oauth2client/file.py
pferate: Went ahead and got rid of them

oauth2client/file.py
pferate: Alright, I changed it to `IOError` for the directory instead of filename.

tests/test_file.py
pferate: Went ahead and fixed throughout file.

tests/test_file.py
pferate: Fixed throughout file, also did `assertIsNotNone`.

pferate:
Updated PR and commit message

tests/test_file.py
```diff
@@ -74,40 +77,52 @@ def _create_test_credentials(self, client_id='some_client_id',
...
             os.unlink(SYMFILENAME)
 
     def test_pickle_and_json_interop(self):
         # Write a file with a pickled OAuth2Credentials.
         credentials = self._create_test_credentials()
 
-        f = open(FILENAME, 'wb')
-        pickle.dump(credentials, f)
-        f.close()
+        cred_file = open(FILENAME, 'wb')
```
nathanielmanistaatgoogle: Also avoid abbreviations; this can and should be `credentials_file`.

oauth2client/file.py
```diff
@@ -20,12 +20,17 @@
...
 
 
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 
+_SYM_LINK_MESSAGE = 'File: {0}: Is a symbolic link.'
+_IS_DIR_MESSAGE = '{0}: Is a directory'
+_MISSING_FILE_MESSAGE = 'Cannot access {0}: No such file or directory'
+
 
 class CredentialsFileSymbolicLinkError(Exception):
```
nathanielmanistaatgoogle: Is this exception class not also on the chopping block? Why should this be used rather than `IOError`?

pferate: I went ahead and removed it, in favor of `IOError`.  When grepping for that exception, I found a duplicated `validate_file` function, so I moved `validate_file` to `oauth2client.util` and used that instead.

tests/test_file.py
pferate: Fixed

nathanielmanistaatgoogle:
I'm satisfied; @jonparrott will re-review since this has changed a bit.

jonparrott:
Thanks, @pferate!

pferate:
No problem @jonparrott!

PR #577, opened 2016-07-29
Remove httplib2 imports from non-transport modules.

dhermes:
Now all usage of `httplib2` is concentrated in tests and in the `oauth2client.transport` module. This does not yet cover all behavior that implicitly relies on `httplib2` in these modules, so there is still work to be done.

---

``` bash
$ git grep import | grep httplib2
oauth2client/transport.py:import httplib2
samples/oauth2_for_devices.py:import httplib2
scripts/run_gce_system_tests.py:import httplib2
scripts/run_system_tests.py:import httplib2
tests/contrib/test_appengine.py:import httplib2
tests/contrib/test_flask_util.py:import httplib2
tests/contrib/test_gce.py:import httplib2
tests/contrib/test_metadata.py:import httplib2
tests/http_mock.py:import httplib2
tests/test_client.py:import httplib2
tests/test_service_account.py:import httplib2
tests/test_transport.py:import httplib2
```

### DIFF
--- a/oauth2client/contrib/_metadata.py
+++ b/oauth2client/contrib/_metadata.py
@@ -20,7 +20,6 @@ See https://cloud.google.com/compute/docs/metadata
 import datetime
 import json
 
-import httplib2
 from six.moves import http_client
 from six.moves.urllib import parse as urlparse
 
@@ -51,7 +50,8 @@ def get(http_request, path, root=METADATA_ROOT, recursive=None):
         A dictionary if the metadata server returns JSON, otherwise a string.
 
     Raises:
-        httplib2.Httplib2Error if an error corrured while retrieving metadata.
+        http_client.HTTPException if an error corrured while
+        retrieving metadata.
     """
     url = urlparse.urljoin(root, path)
     url = util._add_query_parameter(url, 'recursive', recursive)
@@ -68,7 +68,7 @@ def get(http_request, path, root=METADATA_ROOT, recursive=None):
         else:
             return decoded
     else:
-        raise httplib2.HttpLib2Error(
+        raise http_client.HTTPException(
             'Failed to retrieve {0} from the Google Compute Engine'
             'metadata service. Response:\n{1}'.format(url, response))
 
--- a/oauth2client/contrib/gce.py
+++ b/oauth2client/contrib/gce.py
@@ -20,7 +20,7 @@ Utilities for making it easier to use OAuth 2.0 on Google Compute Engine.
 import logging
 import warnings
 
-import httplib2
+from six.moves import http_client
 
 from oauth2client import client
 from oauth2client.contrib import _metadata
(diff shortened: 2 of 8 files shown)


oauth2client/contrib/gce.py
```diff
@@ -134,8 +133,8 @@ def _refresh(self, http_request):
             self._retrieve_info(http_request)
             self.access_token, self.token_expiry = _metadata.get_token(
                 http_request, service_account=self.service_account_email)
-        except httplib2.HttpLib2Error as e:
-            raise client.HttpAccessTokenRefreshError(str(e))
+        except transport.transport_error() as err:
```
dhermes: @jonparrott @elibixby It seems this exception is only relevant here because of the usage of `httplib2.HttpLib2Error` in `oauth2client.contrib._metadata.get()`.

Is this accurate? Could a different, library-agnostic exception be used instead?

theacodes: Yes, I'm not sure I see what needs to be changed here. The idea is if the transport fails to talk to the metadata server, that's equivalent to an AccessTokenRefreshError.

I'm open to ideas, what would you recommend instead?

dhermes: Well we can raise something other than `httplib2.HttpLib2Error` in the `get()` method. This starts to get brittle when methods can raise exceptions (instead of just returning a bad status code).

theacodes: Yeah, I mean, the transport needs to have some way of communicating connection errors. Are you saying `metadata.get()` should return something like `None` instead of raising an exception?

oauth2client/contrib/_metadata.py
```diff
@@ -68,7 +68,8 @@ def get(http_request, path, root=METADATA_ROOT, recursive=None):
         else:
             return decoded
     else:
-        raise httplib2.HttpLib2Error(
+        err_klass = transport.transport_error()
```
theacodes: would `error_class` be okay?

dhermes: Sure

oauth2client/contrib/gce.py
dhermes: No, I'm just saying `httplib2.HttpLib2Error` doesn't need to be the error that communicates the failure. Something more generic would suffice.

Though now that you mention it, the only users of `_metadata.get` is `gce.AppAssertionCredentials._refresh` (via `_metadata.get_token`) and `gce.AppAssertionCredentials._retrieve_info` (via `_metadata.get_service_account_info`) so it'd be easy to change. As currently written,  `get_service_account_info` would return `None` and `get_token` would result in a `TypeError` (via `__getitem__` on `None`).

theacodes: Feel free to change it to whatever you think is most appropriate.

dhermes: @jonparrott Right now I am going with `httplib.HTTPException` (the base for all exceptions defined in `httplib` / `http.client`).

It still feels too specific, though at least it's in the standard library.

dhermes:
@jonparrott PTAL

oauth2client/contrib/gce.py
theacodes: > It still feels too specific, though at least it's in the standard library.

It's also gross for various reasons, but it'll suffice for now. It's easy to change before we make the transport concrete if needed.

dhermes:
Also LMK if you want the two commits squashed into one before merging

jonparrott:
LGTM, and please squash and change the voice to active. `Remove httplib2 imports ...`

Issue #251, opened 2015-08-10
util.py and _helpers.py should be merged


dhermes:
That's my fault for creating `_helpers` and not realizing `util` was there.

dhermes:
(As of 5dda03224fbb408fbb3ac450733b997b525496c6)

----

Things defined in `util`:

- `positional`
- `scopes_to_string`
- `string_to_scopes`
- `dict_to_tuple_key`
- `_add_query_parameter`
- `POSITIONAL_WARNING` (not really relevant, but is in `__all__`)
- `POSITIONAL_EXCEPTION` (not really relevant, but is in `__all__`)
- `POSITIONAL_IGNORE` (not really relevant, but is in `__all__`)

Things defined in `_helpers`:

- `_parse_pem_key`
- `_json_encode`
- `_to_bytes`
- `_from_bytes`
- `_urlsafe_b64encode`
- `_urlsafe_b64decode`

PR #579, opened 2016-08-01
Merge util.py and _helpers.py

pferate:
A new file, `_helpers.py`, was created without realizing that `utils.py` existed for the same purpose.

Resolves: #251, (partially) #428 (partial duplicate of #251)


dhermes:
@pferate I am the originator of `_helpers` and it was very much intentional. If you are merging them then `utils` should go away since it is public.

Also, I seem to recall there being an issue with the LICENSE on that file since it came from Guido on the `protorpc` project

pferate:
Thanks for the info.  I guess I misunderstood your comment from #251:
> That's my fault for creating _helpers and not realizing util was there.
and assumed that the code in `_helpers` should go into `utils`.

dhermes:
Ahh that's a failure of my memory then, the thing I wrote at that time is the real truth.

The goal is a smaller surface area, not a larger one and none of the functions in `_helpers` are in the API surface area, so big :+1: there.

In addition to nuking `util.py`, I'd love to see a plan for deprecating use of the `positional` decorator.

jonparrott:
>  In addition to nuking util.py, I'd love to see a plan for deprecating use of the positional decorator.

File a bug and put it in the 4.0.0 milestone.

jonparrott:
In general, I'd prefer *less* public stuff in this library.

dhermes:
> File a bug and put it in the 4.0.0 milestone.

I apparently already filed #428 for just this and you and @pferate got it in the right milestone

nathanielmanistaatgoogle:
... so is this pull request actionable or should it be closed? Is this where `util.py` should be killed off?

jonparrott:
I don't see anything in `utils` that want to be public. Should merge `utils` into `_helpers` and remove `utils`?

dhermes:
Yes that

pferate:
OK, I'll update this PR.  With `_helpers` being private, should all of the functions and global variables within it be private as well?  Or would that already be handled from the private module?

dhermes:
That is @nathanielmanistaatgoogle's territory (of having more refined opinions than others). IIUC he prefers that imported functions be public, and since the module is non-public, that will be very much OK. (I would suggest holding off of renames like `_helpers._from_bytes` for a separate PR.)

jonparrott:
I also would prefer `_helpers.from_bytes` instead of `_helpers._from_bytes`.

Basically, functions that start with `_` are *module*-private and modules that start with `_` are *package*-private. A package-private module still has a right to privacy.

pferate:
Thanks!

pferate:
PR has been updated (`utils` -> `_helpers`).

oauth2client/_helpers.py
```diff
@@ -11,12 +11,216 @@
...
+    """
+    if not scopes:
+        return []
+    if isinstance(scopes, six.string_types):
+        return scopes.split(' ')
+    else:
+        return scopes
+
+
+def _add_query_parameter(url, name, value):
```
theacodes, line 187: want to go ahead and remove the `_` here?

jonparrott:
This change looks mostly good to me, @nathanielmanistaatgoogle to do the final review.

oauth2client/_helpers.py
pferate, line 187: @dhermes suggested to remove the leading `_`'s in a separate PR.  Or is this one different since I'm already touching the function?

oauth2client/_helpers.py
```diff
@@ -11,12 +11,216 @@
...
 import six
+from six.moves import urllib
+
+
+__author__ = [
+    'rafek@google.com (Rafe Kaplan)',
+    'guido@google.com (Guido van Rossum)',
+]
+
+__all__ = [
```
nathanielmanistaatgoogle: Is the value of this constant correct or stale?

oauth2client/_helpers.py
```diff
@@ -11,12 +11,216 @@
...
 import json
+import logging
+import os
+import warnings
 
 import six
+from six.moves import urllib
+
+
+__author__ = [
```
nathanielmanistaatgoogle: Never write a list literal where a tuple literal will do. _Especially_ for a value like this intended to be an immutable constant.

oauth2client/_helpers.py
```diff
@@ -11,12 +11,216 @@
...
+
+    Args:
+        scopes: a string or iterable of strings, the scopes.
+
+    Returns:
+        The scopes in a list.
+    """
+    if not scopes:
+        return []
+    if isinstance(scopes, six.string_types):
```
nathanielmanistaatgoogle: `elif` rather than `if`.

oauth2client/_helpers.py
```diff
@@ -11,12 +11,216 @@
...
+        value: string, query parameter value.
+
+    Returns:
+        Updated query parameter. Does not update the url if value is None.
+    """
+    if value is None:
+        return url
+    else:
+        parsed = list(urllib.parse.urlparse(url))
+        q = dict(urllib.parse.parse_qsl(parsed[4]))
```
nathanielmanistaatgoogle: Single-letter identifiers are banned by the style guide.

I recognize that this commit is just a code move, but... the patient is kind of on the table. Only kind of, though. Up to you.

nathanielmanistaatgoogle:
Done with my first round of review.

oauth2client/_helpers.py
pferate: Looks like no.  Is there any reason for us to define `__all__` here?  It looks like this file and `tools.py` are the only files that explicitly define `__all__`.  With this being a private module, I think it can be removed.

oauth2client/_helpers.py
pferate: Will do, I just copy-pasted these changes.  Should I leave the values as-is?

oauth2client/_helpers.py
nathanielmanistaatgoogle: I agree that it isn't justifying its existence and should be removed.

oauth2client/_helpers.py
nathanielmanistaatgoogle: One of the many reasons I despise author attribution in code is that there's never a good answer for what to do when a file is modified like we're modifying this file.

I'm not sure of the implications of removing it, so I can't tell you to remove it, so... either keep it as-is or add yourself and the authors of the preëxisting content in `_helpers.py` to it?

tests/test__helpers.py
```diff
@@ -11,13 +11,133 @@
...
+
+class PositionalTests(unittest2.TestCase):
+
+    def test_usage(self):
+        _helpers.positional_parameters_enforcement = (
+            _helpers.POSITIONAL_EXCEPTION)
+
+        # 1 positional arg, 1 keyword-only arg.
+        @_helpers.positional(1)
+        def fn(pos, kwonly=None):
```
pferate: @nathanielmanistaatgoogle, I'll also go ahead and expand `fn`

oauth2client/_helpers.py
pferate: Since the bulk of the file is someone else's code (and I'm not a Googler), I won't add myself.  Looking at the [history](https://github.com/google/oauth2client/commit/68a8cfe26dc8523f91ac8655aae772d8527fca2e#diff-25e202e943c248567c6fe9f00d8c5ee2), I think I'd be more comfortable removing the authors once remove `positional()`.

pferate:
PR has been updated.  I will be creating another PR to remove the `_` from functions in `_helpers`.

PR #583, opened 2016-08-02
Adding transport helper for http.request.

dhermes:
This further makes clear the assumptions about `httplib.Http()` objects and their methods.

Also removing imports / usage of `httplib2` in system tests.

Towards #128.

---
- I'm happy to discuss the usefulness of this abstraction.
- I have left out using `get_callable_http` in unit tests, though I can do that if desired.
- I've [enumerated](https://github.com/google/oauth2client/issues/128#issuecomment-237057097) the various places an `http` object or `http.request` is used. We could just make all the `http_request` arguments in non-public methods into `http` arguments instead


tests/test_transport.py
```diff
@@ -136,3 +136,24 @@ def test_wrap(self):
         self.assertIsNone(result)
         self.assertNotEqual(http.request, orig_req_method)
         self.assertIs(http.request.credentials, credentials)
+
+
+class Test_get_callable_http(unittest2.TestCase):
+
+    def test_it(self):
+        http_obj = mock.Mock()
+        http_obj.request = req_callable = object()
```
nathanielmanistaatgoogle: One assignment per line.

tests/test_transport.py
```diff
@@ -136,3 +136,24 @@ def test_wrap(self):
         self.assertIsNone(result)
         self.assertNotEqual(http.request, orig_req_method)
         self.assertIs(http.request.credentials, credentials)
+
+
+class Test_get_callable_http(unittest2.TestCase):
+
+    def test_it(self):
```
nathanielmanistaatgoogle: No better method names, here and below?

dhermes:
@nathanielmanistaatgoogle I addressed the comments. PTAL.

It's worth discussing more the actual approach here. We may be better off replacing all arguments of the form `http_request` by an `http` and then just have a helper in transport which takes `http` and some arguments and spits out a pair of response (interface TBD) and content (in bytes).

```python
response, content = transport.make_request(http, *args, **kwargs)
```

jonparrott:
I'd prefer `get_requestor` or similar to `get_callable_http`. 

Actually, seriously, I'd rather just `transport.request` and stop passing around http objects. Seems obstinate at this point. You and @nathanielmanistaatgoogle might (and are totally allowed to) disagree here.

jonparrott:
For more clarity:

All internal methods that need http (refresh, info update, etc.) should just call `transport.request()`. We can provide some sort of hooks to transport to switch between them (e.g., `transport.use_httplib2()`, `transport.use_urllib3()`, and maybe even `transport.use_custom(some_function)`.

The only method I feel like should deal with a user-provided http object is `Credentials.authorize`. We need to cleanly separate those concerns. If you even want to focus on removing all usage of httplib2 *except* that usage, we can separately address how `authorize` works.

Does this sound completely crazy or am I making some amount of sense?

jonparrott:
More nonsense:

What I'm getting at here is that what http client oauth2client uses internally and what the user uses can be completely different. In fact, it should be possible for the user to use multiple http clients without issue. In the perfect world, this kind of thing would be possible within the same program:

```python
# urllib3
http = urllib3.PoolManager()
http = oauth2client.authorize(http, credentials)
# httplib2
http = httplib2.Http()
http = oauth2client.authorize(http, credentials)
# requests
http = requests.Session()
http = oauth2client.authorize(http, credentials)
```

All the while never configuring oauth2client itself. We could use urllib3 for all internal calls in all three cases.

dhermes:
@jonparrott Let's go with baby steps and have that conversation over in #128.

I think for now, you've shined (shone?) a light on using `oauth2client.transport.request`:

```python
def request(http, uri, method='GET', ...):
    ...
```

This means getting rid of all the `http_request` arguments to non-public methods. This may break some people, so we could sniff the input to allow both for a few releases.

I'll leave this code here for now and put up a competing PR?

jonparrott:
> This may break some people, so we could sniff the input to allow both for a few releases.

I'm open to that, but definitely want to have a plan for outright removing that.

> I'll leave this code here for now and put up a competing PR?

Sounds good.

dhermes:
Closing in favor of #586

PR #586, opened 2016-08-03
Using transport helper for calling http.request().

dhermes:
This (reference to `http.request`) assumes (for now) that `http` is an instance of `httplib.Http`.

Alternative approach to #583. Towards #128.

Note that this PR stops short of
- just using `http` instead of `http.request`
- changing the input argument in all helpers from `http_request` to `http`


tests/contrib/test_metadata.py
```diff
@@ -28,13 +28,23 @@
...
 
 
 def request_mock(status, content_type, content):
     response = http_mock.ResponseMock(
         {'status': status, 'content-type': content_type})
-    return mock.Mock(return_value=(response, content.encode('utf-8')))
+    request_method = mock.Mock(
+        return_value=(response, content.encode('utf-8')))
+    # Make sure the mock doesn't have a request attr.
+    del request_method.request
```
dhermes, line 46: Here and elsewhere that I have to `del foo.request`, I'd rather just update the code to properly use our `http_mock.HttpMock` test class.

jonparrott:
It still seems strange to me that `http` is an argument to `transport.request`.

dhermes:
@jonparrott Baby steps

jonparrott:
> @jonparrott Baby steps

But it makes me feel icky.

But more seriously this looks fine so long as there's another PR to nix the http arg.

jonparrott:
Actually, why not go ahead and make http and optional arg to `transport.request`?

dhermes:
I'd rather have a discussion first about your idea of "under the covers" vs. up-front `http`. Let's hold off on making it optional until we can hash out more details.

`transport` is essentially complete, except for possibly adding a helper to get status or other headers from a response object.

tests/test_client.py
```diff
@@ -1381,6 +1390,10 @@ def _do_retrieve_scopes_test_helper(self, response, content,
         assertUrisEqual(self, token_uri, called_uri)
         logger.info.assert_called_once_with('Refreshing scopes')
 
+        http_request.assert_called_once_with(
+            token_uri, method='GET', body=None, headers=None,
+            redirections=5, connection_type=None)
```
dhermes: @jonparrott I (think I) now see why you didn't make this assertion, the `token_uri` and `called_uri` might not be literally equal due to dictionary order in Python (flaky failed tests).

Fail by me for not noticing above

``` python
self.assertEqual(len(http_request.mock_calls), 1)
scopes_call = http_request.mock_calls[0]
```

jonparrott:
> I'd rather have a discussion first about your idea of "under the covers" vs. up-front http. Let's hold off on making it optional until we can hash out more details.

SG.

dhermes:
@jonparrott Good to merge? @nathanielmanistaatgoogle any thoughts?

jonparrott:
I'm fine with it for now.

nathanielmanistaatgoogle:
Resolve conflicts and poke me again?

dhermes:
@nathanielmanistaatgoogle resolved

PR #589, opened 2016-08-04
Remove contrib.multistore_file

### LINKED ISSUE #470: Remove multistore file/locked file

theacodes:
Remove dependent modules as well. Resolves #470.


nathanielmanistaatgoogle:
I regret that I have but one "looks fantastic" to award this deletion-only commit.

Merge at your leisure.

jonparrott:
gone-done

PR #590, opened 2016-08-04
Drop support for Python 2.6

### LINKED ISSUE #298: Drop support for Python 2.6

theacodes:
Resolves #298.


jonparrott:
(Don't merge until I verify that Travis still works)

.travis.yml
```diff
@@ -1,15 +1,7 @@
 language: python
-python: 2.7
```
nathanielmanistaatgoogle, line 2: I don't understand the changes on lines two through nine of this file.

nathanielmanistaatgoogle:
[I thought we were committed through 1 September](https://github.com/google/oauth2client/issues/298#issuecomment-215211821). [What am I missing](https://days.to/1-september/2016)? @thobrla, has something changed?

.travis.yml
theacodes, line 2: `-python 2.7` isn't required as by default travis will expose all available python interpreters.

The `matrix` part shouldn't be needed anymore, as travis should provide the 3.5 interpreter. (this needs to be verified)

thobrla:
No change on the gsutil side, still planning on removing Py 2.6. support from gsutil releases made after Sept 1.

jonparrott:
@nathanielmanistaatgoogle I don't anticipate us releasing 4.0.0 until september 1st, and going ahead and getting 2.6 out of the way will allow us to continue our hygenie cleanups unencumbered by 2.6

gcloud-python also dropped 2.6 support today. We're not alone.

jonparrott:
Travis flailing concluded. Everything is green and okay to merge.

pferate:
Should we take care of `unittest2` in this PR?  Or create another PR?

jonparrott:
@pferate another PR, I just wanted to focus on removing the build system support and updated `setup.py`.

nathanielmanistaatgoogle:
If we integrate this pull request now, doesn't that put us in an unreleasable state until 1 September? Or would we say something like "oh, use 3.x if you want 2.6 support"?

jonparrott:
> If we integrate this pull request now, doesn't that put us in an unreleasable state until 1 September?

Only if we really want to hold steadfast to that date. We have essentially 3 downstream dependencies to worry about:

1. google-api-python-client, which you and I maintain.
2. gsutil, which hasn't updated to 3.0.0 yet.
3. gcloud-python, who just dropped 2.6 today.

nathanielmanistaatgoogle:
So if we adopt this now, gsutil would have to pin their dependency on us to 3.0.0-or-lower through the end of this month? I recognize that the likelihood of problems is very small; I just want to make sure that the correct bounding box has been drawn around them.

jonparrott:
@thobrla can confirm, but I think they're currently bundling < 3.0.0 (maybe even less than 2.0.0).

thobrla:
Correct - we would need to continue to pin to 3.0.0 or lower.  We're already pinned to 2.2.0 because we have porting and integration testing work needed to adopt @jonparrott 's replacement for multistore_file.

nathanielmanistaatgoogle:
... so @thobrla, it sounds like your explicit consent is needed for this merge? For you to say something like "gsutil will not move its `oauth2client` pin past 3.0.0 before 1 September"?

thobrla:
Sorry - let me be clear.  gsutil is fine with this merge; we won't move past 3.0.0 before 1 September.  The caveat is if we find a critical bug in the 2.x branch, we'll need to fix it there and release.

nathanielmanistaatgoogle:
Squash and merge!

.travis.yml
```diff
@@ -1,29 +1,35 @@
...
+  - python: 3.4
+    env: TOX_ENV=py34
+  - python: 3.5
+    env: TOX_ENV=py35
+  - python: pypy
+    env: TOX_ENV=pypy
+  - python: 2.7
+    env: TOX_ENV=system-tests
+  - python: 3.4
+    env: TOX_ENV=system-tests3
```
pferate, line 25: Sorry for the late comment.  What do you think using something like [tox-travis](https://pypi.python.org/pypi/tox-travis) to simplify this part.

dhermes, line 25: @pferate Thanks for bringining `tox-travis` to my attention!

theacodes, line 25: I'm ambivalent.

dhermes:
@pferate @jonparrott `unittest2` can be squashed now if someone is up for it

pferate:
We can easily use `sed` to go from`unittest2` -> `unittest`, or we can move the tests to `pytest`.  I still have my branch for reference of the headaches I dealt with the first time I did that.

dhermes:
Isn't the move to `pytest` still in progress?

pferate:
I had a huge PR (#547) that did the full migration, but it was requested to break it up to make it easier to review.

dhermes:
Gotcha. Brain dump PRs are difficult to review, I can sympathize. e.g. `httplib2` wouldn't need to be in that PR.

I'll take a stab at it right now.

PR #591, opened 2016-08-04
Open up MultiprocessStorageBackend.

bochunz:
1. Open up the class MultiprocessStorageBackend to be used by oauth2l.
2. Add a clear credential method that remove all credential stored.

oauth2l needs this change to implement a revoke feature where all cached
tokens are removed from cache and revoked.

### DIFF
--- a/oauth2client/contrib/multiprocess_file_storage.py
+++ b/oauth2client/contrib/multiprocess_file_storage.py
@@ -186,7 +186,7 @@ def _write_credentials_file(credentials_file, credentials):
     credentials_file.truncate()
 
 
-class _MultiprocessStorageBackend(object):
+class MultiprocessStorageBackend(object):
     """Thread-local backend for multiprocess storage.
 
     Each process has only one instance of this backend per file. All threads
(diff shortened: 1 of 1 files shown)


jonparrott:
We are strongly trying not to introduce more public API surface area. Why should we make this public?

bochunz:
I'm implementing a revoke method for oauth2l where I will need to clear all credentials stored in the local cache.

craigcitro:
I guess I'm not clear on why we need this for `oauth2l`?

1. The storage class already has a public delete method.
2. If you're really looking to blow away all credentials, you can just drop the file.

I would be interested to know precisely what semantics you're trying to implement on the `oauth2l` side, and why the existing surface doesn't already support them.

bochunz:
To revoke all tokens in the cache, I will need to
1. get all credentials in the cache
2. call something like https://<revoke_url>?token=xxx for each token
3. delete them from local cache
- So just drop the file won't work for number 2

As disussed with Hong, I will put this PR on hold since we have a deadline to meet. I will first only implement revoking a particular credential, which doesn't need any change here.

jonparrott:
I'd be amenable to adding a top-level module function to get a list of all keys in the file. I would not be amenable to making the entire backend publicly visible.

I'm closing this PR, but feel free to continue discussion.

PR #598, opened 2016-08-09
Remove all uses of MagicMock

dhermes:
`MagicMock` is a version of `Mock` with extra features for tests that need to test usage of magic methods (i.e. `__nonzero__`, `__cmp__`, etc.).

None of the `oauth2client` tests actually need these extra features though.

### DIFF
--- a/tests/contrib/django_util/test_decorators.py
+++ b/tests/contrib/django_util/test_decorators.py
@@ -63,7 +63,7 @@ class OAuth2EnabledDecoratorTest(TestWithDjangoEnvironment):
     @mock.patch('oauth2client.client.OAuth2Credentials')
     def test_has_credentials_in_storage(self, OAuth2Credentials):
         request = self.factory.get('/test')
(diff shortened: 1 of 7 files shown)


dhermes:
It seems py33 has started failing because of `django`. FWIW `django` [dropped support][1] for 3.3 on Dec. 1, 2015.

The `gcloud-python` project has also stopped supporting 3.3

[1]: https://github.com/GoogleCloudPlatform/gcloud-python/issues/995#issuecomment-168126798

jonparrott:
We might as well drop 3.3 as well. This LGTM, the travis failure should be addressed separately.

nathanielmanistaatgoogle:
Commit itself looks fine; will this not get a passing build until 3.3 support is dropped?

jonparrott:
@nathanielmanistaatgoogle correct. I'll try to do that today (dealing with a product launch).

dhermes:
TADA! I just made #604

PR #602, opened 2016-08-10
Remove distinction between http AND http.request

dhermes:
Towards #128.

It's unclear why `http_request` was passed around instead of `http`, other than possible to make mocking easier, but it became a rampant pattern.


dhermes:
@jonparrott I think you should give this a once-over to make sure we're on the same page here.

jonparrott:
@dhermes I think this is fine. Your next step is to make the http arg optional for all of these? 

Or would you prefer to go ahead and remove the http arg altogether as the next step?

PR #605, opened 2016-08-10
Clean up usage of HTTP mocks in tests.

dhermes:
This is a follow-on to #602, and partially making good on a [comment](https://github.com/google/oauth2client/pull/586/files#r73388113) I made in #586.

I checked the types of `http` values sent to `transport.request` by adding:

``` python
from tests import http_mock
import types
if type(http) == http_mock.HttpMock:
    pass
elif type(http) == http_mock.HttpMockSequence:
    pass
elif (type(http) == types.MethodType and
      http.__self__.__class__ == http_mock.HttpMock):
    pass
elif (type(http) == types.MethodType and
      http.__self__.__class__ == http_mock.HttpMockSequence):
    pass
elif (type(http) == types.FunctionType and
      http.__name__ == 'new_request'):
    pass
else:
    raise MyErr(http)
```

I also did `git grep http_request` to find some other iffy usage. I'd also like to track down usage of `http.request` (or just `.request` on an object).

The method type and function type checks are to accommodate the hacky way an `http` object is monkey-patched, so I'd also like to fix those up as part of #128.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -12,10 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Oauth2client tests
-
-Unit tests for oauth2client.
-"""
+"""Unit tests for oauth2client.client."""
 
 import base64
 import contextlib
@@ -705,8 +702,9 @@ class GoogleCredentialsTests(unittest2.TestCase):
         # Make sure the well-known file actually doesn't exist.
         self.assertTrue(os.path.exists(get_well_known.return_value))
 
-        method_name = \
-            'oauth2client.client._get_application_default_credential_from_file'
+        method_name = (
+            'oauth2client.client.'
+            '_get_application_default_credential_from_file')
         result_creds = object()
         with mock.patch(method_name,
                         return_value=result_creds) as get_from_file:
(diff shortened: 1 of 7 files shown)


tests/test_client.py
```diff
@@ -705,8 +702,8 @@ def test_get_adc_env_from_well_known(self, *stubs):
         # Make sure the well-known file actually doesn't exist.
         self.assertTrue(os.path.exists(get_well_known.return_value))
 
-        method_name = \
-            'oauth2client.client._get_application_default_credential_from_file'
+        method_name = ('oauth2client.client.'
```
theacodes: break after opening `(`?

jonparrott:
This looks good to me.

tests/test_client.py
dhermes: I assume you mean break == line break == newline.

Going with the general philosophy that multiline continuations should just jump to the 2nd line immediately?

I folded the change into the original commit

theacodes: Yar. Thx.

dhermes:
@jonparrott What do you have against the acronym LGTM :wink:

jonparrott:
I don't know.

PR #606, opened 2016-08-10
Use transport helpers in system tests

dhermes:
Was supposed to go in #586 (was originally in #583) but it got lost

Towards #128


jonparrott:
LGTM

PR #608, opened 2016-08-11
Remove backslash line continuations

dhermes:
@nathanielmanistaatgoogle Should I add a rule in `CONTRIBUTING` that says

> Pretty please don't use backslash line continuations, they aren't needed.


nathanielmanistaatgoogle:
No, `CONTRIBUTING` already requires new code to conform to the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html); that's enough.

PR #610, opened 2016-08-11
Drop unittest2 dependency

### LINKED ISSUE #609: Drop unittest2 dependency since Python 2.6 is gone
Tackling this right now

/cc @pferate

dhermes:
No longer needed since Python 2.6 support has been dropped.

Fixes #609

/cc @pferate

### DIFF
--- a/tests/contrib/django_util/test_django_models.py
+++ b/tests/contrib/django_util/test_django_models.py
@@ -19,17 +19,16 @@ Unit tests for models and fields defined by the django_util helper.
 
 import base64
 import pickle
+import unittest
 
 from tests.contrib.django_util.models import CredentialsModel
-
-import unittest2
 
 from oauth2client import _helpers
 from oauth2client.client import Credentials
 from oauth2client.contrib.django_util.models import CredentialsField
 
 
-class TestCredentialsField(unittest2.TestCase):
+class TestCredentialsField(unittest.TestCase):
 
     def setUp(self):
         self.fake_model = CredentialsModel()
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -24,12 +24,12 @@ import os
 import pickle
 import stat
 import tempfile
+import unittest
 import warnings
 
 import mock
 import six
 from six.moves import http_client
-import unittest2
 
 from oauth2client import _helpers
 from oauth2client import client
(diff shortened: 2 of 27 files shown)


tests/test_tools.py
```diff
@@ -12,24 +12,20 @@
...
 from six.moves.urllib import request
-import unittest2
 
 from oauth2client import client
 from oauth2client import tools
 
-try:
-    import argparse
-except ImportError:  # pragma: NO COVER
-    raise unittest2.SkipTest('argparase unavailable.')
```
dhermes, line 29: REVIEWER(S): This part is the only non-trivial change, since the `argparse` import will succeed for any version after  (but not including) Python 2.5

nathanielmanistaatgoogle:
I'm satisfied. @jonparrott: will you be taking a look?

tests/test_file.py
```diff
@@ -25,11 +25,11 @@
 import stat
 import tempfile
 import warnings
+import unittest
```
pferate, line 27: Move up a line to avoid the `flake8` error.

tests/contrib/django_util/test_django_models.py
```diff
@@ -19,17 +19,15 @@
...
-
-from tests.contrib.django_util.models import CredentialsModel
-
-import unittest2
+import unittest
 
 from oauth2client import _helpers
 from oauth2client.client import Credentials
 from oauth2client.contrib.django_util.models import CredentialsField
+from tests.contrib.django_util.models import CredentialsModel
```
pferate: Move this line back for now, to appease `flake8`.  I'll take care of it in another PR today, since the `flake8` config and a few more files need to be updated.

pferate:
LGTM, other than a couple minor `flake8` errors.  I saw some other imports errors that I will resolve in a separate PR, since it's outside the scope of this one.

tests/test_file.py
dhermes, line 27: D'oh! Thanks

dhermes:
@nathanielmanistaatgoogle I don't think @jonparrott will mind not reviewing this mindless change. FWIW I assign both of you just to reach one of you (and here I knew @pferate was interested as well).

jonparrott:
I didn't mind not reviewing it. As long as the test god is pleased.

PR #612, opened 2016-08-11
Use transport module for GCE environment check.

### LINKED ISSUE #599: Should we use a more generic transport for _detect_gce_environment()
The current [source][1] uses the core `httplib` / `http.client` machinery to get the info.

1.  The original reason we used the low-level library is because the timeout in `httplib2` was taking too long. This may no longer be an issue, i.e. if the GCE check is the last one
1.  It should probably use `oauth2client.contrib._metadata` utils

I'm filing this because after #586 all requests (except for this one) are "guaranteed" to pass through the `transport.request` method

[1]: https://github.com/google/oauth2client/blob/b1a17432731344d2e44b1ef78ab52f2202bee703/oauth2client/client.py#L995-L1022

dhermes:
Fixes #599.

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -108,9 +108,10 @@ except ValueError:  # pragma: NO COVER
     GCE_METADATA_TIMEOUT = 3
 
 _SERVER_SOFTWARE = 'SERVER_SOFTWARE'
-_GCE_METADATA_HOST = '169.254.169.254'
-_METADATA_FLAVOR_HEADER = 'Metadata-Flavor'
+_GCE_METADATA_URI = 'http://169.254.169.254'
+_METADATA_FLAVOR_HEADER = 'metadata-flavor'  # lowercase header
 _DESIRED_METADATA_FLAVOR = 'Google'
+_GCE_HEADERS = {_METADATA_FLAVOR_HEADER: _DESIRED_METADATA_FLAVOR}
 
 # Expose utcnow() at module level to allow for
 # easier testing (by replacing with a stub).
@@ -997,21 +998,16 @@ def _detect_gce_environment():
     #       could lead to false negatives in the event that we are on GCE, but
     #       the metadata resolution was particularly slow. The latter case is
     #       "unlikely".
-    connection = six.moves.http_client.HTTPConnection(
-        _GCE_METADATA_HOST, timeout=GCE_METADATA_TIMEOUT)
-
+    http = transport.get_http_object(timeout=GCE_METADATA_TIMEOUT)
     try:
-        headers = {_METADATA_FLAVOR_HEADER: _DESIRED_METADATA_FLAVOR}
-        connection.request('GET', '/', headers=headers)
-        response = connection.getresponse()
-        if response.status == http_client.OK:
-            return (response.getheader(_METADATA_FLAVOR_HEADER) ==
-                    _DESIRED_METADATA_FLAVOR)
+        response, _ = transport.request(
+            http, _GCE_METADATA_URI, headers=_GCE_HEADERS)
+        return (
+            response.status == http_client.OK and
+            response.get(_METADATA_FLAVOR_HEADER) == _DESIRED_METADATA_FLAVOR)
     except socket.error:  # socket.timeout or socket.error(64, 'Host is down')
         logger.info('Timeout attempting to reach GCE metadata service.')
         return False
-    finally:
-        connection.close()
 
 
 def _in_gae_environment():
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -42,8 +42,6 @@ from tests import http_mock
 DATA_DIR = os.path.join(os.path.dirname(__file__), 'data')
 
 
-# TODO(craigcitro): This is duplicated from
-# googleapiclient.test_discovery; consolidate these definitions.
 def assertUrisEqual(testcase, expected, actual):
     """Test that URIs are the same, up to reordering of query parameters."""
     expected = urllib.parse.urlparse(expected)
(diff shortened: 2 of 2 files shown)


oauth2client/client.py
```diff
@@ -110,8 +110,8 @@
     GCE_METADATA_TIMEOUT = 3
 
 _SERVER_SOFTWARE = 'SERVER_SOFTWARE'
-_GCE_METADATA_HOST = '169.254.169.254'
-_METADATA_FLAVOR_HEADER = 'Metadata-Flavor'
+_GCE_METADATA_URI = 'http://169.254.169.254'
+_METADATA_FLAVOR_HEADER = 'metadata-flavor'  # lowercase header
```
nathanielmanistaatgoogle, line 112: I don't understand the significance of `# lowercase header` on this line.

oauth2client/client.py
```diff
@@ -999,21 +999,19 @@ def _detect_gce_environment():
     #       could lead to false negatives in the event that we are on GCE, but
     #       the metadata resolution was particularly slow. The latter case is
     #       "unlikely".
-    connection = six.moves.http_client.HTTPConnection(
-        _GCE_METADATA_HOST, timeout=GCE_METADATA_TIMEOUT)
-
+    http = transport.get_http_object(timeout=GCE_METADATA_TIMEOUT)
+    headers = {_METADATA_FLAVOR_HEADER: _DESIRED_METADATA_FLAVOR}
```
nathanielmanistaatgoogle: Does this have to be function-scope? It looks like it can be a module-scope constant.

tests/test_client.py
```diff
@@ -357,67 +357,48 @@ def test_environment_caching(self):
...
-    def _environment_check_gce_helper(self, status_ok=True, socket_error=False,
+    def _environment_check_gce_helper(self, status_ok=True,
                                       server_software=''):
-        response = mock.Mock()
         if status_ok:
-            response.status = http_client.OK
-            response.getheader = mock.Mock(
-                name='getheader',
-                return_value=client._DESIRED_METADATA_FLAVOR)
+            headers = {
```
nathanielmanistaatgoogle: Can this be a module-scope constant? Does placing it at module-scope save having to line-wrap one of its key-value pairs?

tests/test_client.py
```diff
@@ -426,8 +407,23 @@ def test_environment_check_gce_prod_with_working_gae_imports(self):
...
+                return_value=object())
+    @mock.patch('oauth2client.transport.request',
+                side_effect=socket.timeout())
+    def test_environment_check_gce_timeout(self, mock_request, new_http):
+        self.assertFalse(client._in_gae_environment())
+        self.assertFalse(client._in_gce_environment())
+
+        # Verify mocks.
+        new_http.assert_called_once_with(timeout=client.GCE_METADATA_TIMEOUT)
+        headers = {
```
nathanielmanistaatgoogle: If a constant was defined in the `client` module, use it here?

oauth2client/client.py
```diff
@@ -999,21 +999,19 @@ def _detect_gce_environment():
...
     try:
-        headers = {_METADATA_FLAVOR_HEADER: _DESIRED_METADATA_FLAVOR}
-        connection.request('GET', '/', headers=headers)
-        response = connection.getresponse()
+        response, _ = transport.request(
+            http, _GCE_METADATA_URI, headers=headers)
         if response.status == http_client.OK:
```
nathanielmanistaatgoogle: This can be structurally simplified to

```
return (
    response.status == http_client.OK and
    response.get(_METADATA_FLAVOR_HEADER) == _DESIRED_METADATA_FLAVOR)
```

. Maybe with shorter names for the constants it would look good laid out as well?

oauth2client/client.py
nathanielmanistaatgoogle, line 112: How essential is "_HEADER" in this name?

oauth2client/client.py
```diff
@@ -110,8 +110,8 @@
     GCE_METADATA_TIMEOUT = 3
 
 _SERVER_SOFTWARE = 'SERVER_SOFTWARE'
-_GCE_METADATA_HOST = '169.254.169.254'
-_METADATA_FLAVOR_HEADER = 'Metadata-Flavor'
+_GCE_METADATA_URI = 'http://169.254.169.254'
+_METADATA_FLAVOR_HEADER = 'metadata-flavor'  # lowercase header
 _DESIRED_METADATA_FLAVOR = 'Google'
```
nathanielmanistaatgoogle, line 113: How essential is "_METADATA" in this name?

oauth2client/client.py
dhermes, line 112: > the significance of `# lowercase header`

The previous value broke `httplib2` (headers in `httplib2` are all lowercased so avoid allowing case-insensitive header lookups)

> How essential is "_HEADER" in this name?

Making the name shorter just to be short is a loss for code readability. The name describes what the variable is.

oauth2client/client.py
dhermes, line 113: Ditto on my answer above

oauth2client/client.py
dhermes: Argument against making it module-scoped: `dict` is mutable and `frozendict` does not exist.

Though I'm happy to do it, I was just keeping the status quo from the old version (which I also wrote).

oauth2client/client.py
dhermes: :-1: on shorter names for "look good". Your suggestion fits in the first 78 columns anyhow.

tests/test_client.py
dhermes: Sure thing.

tests/test_client.py
dhermes: Would use instead of the `test_client` module level constant, but the "if in `client`" discussion is still pending

dhermes:
@nathanielmanistaatgoogle PTAL I pushed a 2nd commit with some of the changes you suggest (see my responses for the changes I didn't make).

I also removed a no-longer relevant TODO (about separation from `google-api-python-client`) and the `__author__` from the `test_client` file. (I did this because I noticed them when adding a module level constant in `test_client`.)

oauth2client/client.py
```diff
@@ -999,21 +999,17 @@ def _detect_gce_environment():
     #       could lead to false negatives in the event that we are on GCE, but
     #       the metadata resolution was particularly slow. The latter case is
     #       "unlikely".
-    connection = six.moves.http_client.HTTPConnection(
-        _GCE_METADATA_HOST, timeout=GCE_METADATA_TIMEOUT)
-
+    http = transport.get_http_object(timeout=GCE_METADATA_TIMEOUT)
```
theacodes, line 1001: Is it not possible to have timeout be an argument on `request`?

dhermes, line 1001: It is "possible", though I'm not sure how it would work with existing `httplib2.Http` objects passed in.

We could file an issue and then follow up after / during redesign? (Now that the quarantine is essentially done.)

theacodes, line 1001: I would greatly prefer that, as when we switch to urllib3 we will want to use a single PoolManager instead of individual http instances.

dhermes, line 1001: Filed #618 for now.

jonparrott:
LGTM for now.

dhermes:
@jonparrott Can you think of any reason that this wouldn't have been the original implementation?

@nathanielmanistaatgoogle Any issues?

jonparrott:
I'm not sure, I didn't write the original environment check. /shruggie

dhermes:
Got heem

oauth2client/client.py
nathanielmanistaatgoogle: Please promote to a module-scope constant. Right now its location alone suggests to the reader that it needs to be where it is.

I also feel the pain of `frozendict` not existing.

dhermes:
@nathanielmanistaatgoogle PTAL. Pretty please remind me to squash in your LGTM. (@jonparrott is there no such thing as a squash-bot / remind-bot service for GitHub?)

jonparrott:
@dhermes what would you want it to do?

dhermes:
Block a merge until commits with the word `SQUASH` are gone? Or until a `squash` label is removed.

jonparrott:
I could probably add that to @dpebot once he's functional.

dhermes:
Ha awesome!

nathanielmanistaatgoogle:
Change content looks good; squash and merge.

Issue #594, opened 2016-08-09
The django_util session data does not work with Django 1.10


huwshimi:
The session serialiser with Django 1.10 does not support picked data, like that attached to the session here: https://github.com/google/oauth2client/blob/master/oauth2client/contrib/django_util/views.py#L74

This is the error received:
```
Traceback (most recent call last):
  File "/home/user/project/env/lib/python3.5/site-packages/django/core/handlers/base.py", line 131, in get_response
    response = middleware_method(request, response)
  File "/home/user/project/env/lib/python3.5/site-packages/opbeat/utils/wrapt/wrappers.py", line 562, in __call__
    args, kwargs)
  File "/home/user/project/env/lib/python3.5/site-packages/opbeat/contrib/django/middleware/__init__.py", line 86, in process_response_wrapper
    response = wrapped(*args, **kwargs)
  File "/home/user/project/env/lib/python3.5/site-packages/django/contrib/sessions/middleware.py", line 58, in process_response
    request.session.save()
  File "/home/user/project/env/lib/python3.5/site-packages/django/contrib/sessions/backends/db.py", line 83, in save
    obj = self.create_model_instance(data)
  File "/home/user/project/env/lib/python3.5/site-packages/django/contrib/sessions/backends/db.py", line 69, in create_model_instance
    session_data=self.encode(data),
  File "/home/user/project/env/lib/python3.5/site-packages/django/contrib/sessions/backends/base.py", line 98, in encode
    serialized = self.serializer().dumps(session_dict)
  File "/home/user/project/env/lib/python3.5/site-packages/django/core/signing.py", line 93, in dumps
    return json.dumps(obj, separators=(',', ':')).encode('latin-1')
  File "/usr/lib/python3.5/json/__init__.py", line 237, in dumps
    **kw).encode(obj)
  File "/usr/lib/python3.5/json/encoder.py", line 198, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.5/json/encoder.py", line 256, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.5/json/encoder.py", line 179, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: b'\x80\x03coauth2client.client\nOAuth2WebServerFlow\nq\x00)\x81q\x01}q\x02(X\n\x00\x00\x00revoke_uriq\x03X+\x00\x00\x00[...scrubbed]' is not JSON serializable
```
[...]

PR #614, opened 2016-08-11
Switch from pickle to jsonpickle

waprin:
Addresses #594. As the reporter mentions the PickleSerializer is not recommended. The flow object can't be encoded with `json.dumps`, we could add a custom serializer but this jsonpickle library seems to work and is super simple so figured I would try that.

/cc @huwshimi

### DIFF
--- a/oauth2client/contrib/django_util/views.py
+++ b/oauth2client/contrib/django_util/views.py
@@ -22,13 +22,13 @@ in the configured storage."""
 import hashlib
 import json
 import os
-import pickle
 
 from django import http
 from django import shortcuts
 from django.conf import settings
 from django.core import urlresolvers
 from django.shortcuts import redirect
+import jsonpickle
 from six.moves.urllib import parse
 
 from oauth2client import client
@@ -71,7 +71,7 @@ def _make_flow(request, scopes, return_url=None):
             urlresolvers.reverse("google_oauth:callback")))
 
     flow_key = _FLOW_KEY.format(csrf_token)
-    request.session[flow_key] = pickle.dumps(flow)
+    request.session[flow_key] = jsonpickle.encode(flow)
     return flow
 
 
@@ -89,7 +89,7 @@ def _get_flow_for_token(csrf_token, request):
         CSRF token.
     """
     flow_pickle = request.session.get(_FLOW_KEY.format(csrf_token), None)
-    return None if flow_pickle is None else pickle.loads(flow_pickle)
+    return None if flow_pickle is None else jsonpickle.decode(flow_pickle)
 
 
 def oauth2_callback(request):
--- a/tests/contrib/django_util/test_views.py
+++ b/tests/contrib/django_util/test_views.py
@@ -152,8 +152,8 @@ class Oauth2CallbackTest(TestWithDjangoEnvironment):
         self.user = User.objects.create_user(
             username='bill', email='bill@example.com', password='hunter2')
 
-    @mock.patch('oauth2client.contrib.django_util.views.pickle')
-    def test_callback_works(self, pickle):
+    @mock.patch('oauth2client.contrib.django_util.views.jsonpickle')
+    def test_callback_works(self, jsonpickle_mock):
         request = self.factory.get('oauth2/oauth2callback', data={
             'state': json.dumps(self.fake_state),
             'code': 123
@@ -169,9 +169,10 @@ class Oauth2CallbackTest(TestWithDjangoEnvironment):
             redirect_uri=request.build_absolute_uri("oauth2/oauth2callback"))
 
         name = 'google_oauth2_flow_{0}'.format(self.CSRF_TOKEN)
-        self.session[name] = pickle.dumps(flow)
+        pickled_flow = object()
+        self.session[name] = pickled_flow
         flow.step2_exchange = mock.Mock()
-        pickle.loads.return_value = flow
+        jsonpickle_mock.decode.return_value = flow
 
         request.session = self.session
         request.user = self.user
@@ -180,9 +181,10 @@ class Oauth2CallbackTest(TestWithDjangoEnvironment):
         self.assertEqual(
             response.status_code, django.http.HttpResponseRedirect.status_code)
         self.assertEqual(response['Location'], self.RETURN_URL)
+        jsonpickle_mock.decode.assert_called_once_with(pickled_flow)
 
-    @mock.patch('oauth2client.contrib.django_util.views.pickle')
-    def test_callback_handles_bad_flow_exchange(self, pickle):
+    @mock.patch('oauth2client.contrib.django_util.views.jsonpickle')
+    def test_callback_handles_bad_flow_exchange(self, jsonpickle_mock):
         request = self.factory.get('oauth2/oauth2callback', data={
             "state": json.dumps(self.fake_state),
             "code": 123
@@ -198,17 +200,19 @@ class Oauth2CallbackTest(TestWithDjangoEnvironment):
             redirect_uri=request.build_absolute_uri('oauth2/oauth2callback'))
 
         session_key = 'google_oauth2_flow_{0}'.format(self.CSRF_TOKEN)
-        self.session[session_key] = pickle.dumps(flow)
+        pickled_flow = object()
+        self.session[session_key] = pickled_flow
 
         def local_throws(code):
             raise FlowExchangeError('test')
 
         flow.step2_exchange = local_throws
-        pickle.loads.return_value = flow
+        jsonpickle_mock.decode.return_value = flow
 
         request.session = self.session
         response = views.oauth2_callback(request)
         self.assertIsInstance(response, http.HttpResponseBadRequest)
+        jsonpickle_mock.decode.assert_called_once_with(pickled_flow)
 
     def test_error_returns_bad_request(self):
         request = self.factory.get('oauth2/oauth2callback', data={
--- a/tox.ini
+++ b/tox.ini
@@ -14,6 +14,7 @@ basedeps = mock>=1.3.0
 deps = {[testenv]basedeps}
        django
        keyring
+       jsonpickle
 setenv =
     pypy: with_gmp=no
     DJANGO_SETTINGS_MODULE=tests.contrib.django_util.settings


oauth2client/contrib/django_util/views.py
```diff
@@ -36,6 +35,8 @@
 from oauth2client.contrib.django_util import get_storage
 from oauth2client.contrib.django_util import signals
 
+import jsonpickle
```
dhermes: 3rd-party?

dhermes: Either way, this import should go with the `django` (i.e. 3rd-party) imports, right?

tests/contrib/django_util/test_views.py
```diff
@@ -169,9 +169,11 @@ def test_callback_works(self, pickle):
             redirect_uri=request.build_absolute_uri("oauth2/oauth2callback"))
 
         name = 'google_oauth2_flow_{0}'.format(self.CSRF_TOKEN)
-        self.session[name] = pickle.dumps(flow)
+        self.session[name] = jsonpickle.encode(flow)
+        self.session['google_oauth2_flow_{0}'.format(self.CSRF_TOKEN)] \
+            = jsonpickle.encode(flow)
```
dhermes: Please don't use backslash continuation (in the Google style guide) Instead:

``` python
        key = 'google_oauth2_flow_{0}'.format(self.CSRF_TOKEN)
        self.session[key] = jsonpickle.encode(flow)
```

I just fixed this in #608, and you put it back :cry:

tests/contrib/django_util/test_views.py
```diff
@@ -152,8 +152,8 @@ def setUp(self):
         self.user = User.objects.create_user(
             username='bill', email='bill@example.com', password='hunter2')
 
-    @mock.patch('oauth2client.contrib.django_util.views.pickle')
-    def test_callback_works(self, pickle):
+    @mock.patch('oauth2client.contrib.django_util.views.jsonpickle')
+    def test_callback_works(self, jsonpickle):
```
dhermes: 1. Suggest rename as `jsonpickle_mock`
2. Can you verify the `jsonpickle` mock gets called or that `jsonpickle.encode` gets called, etc.? (I realize the `pickle` mock wasn't verified, but "you break it, you but it" has been @nathanielmanistaatgoogle rule.)

tests/contrib/django_util/test_views.py
```diff
@@ -197,14 +199,15 @@ def test_callback_handles_bad_flow_exchange(self, pickle):
             state=json.dumps(self.fake_state),
             redirect_uri=request.build_absolute_uri('oauth2/oauth2callback'))
 
-        session_key = 'google_oauth2_flow_{0}'.format(self.CSRF_TOKEN)
-        self.session[session_key] = pickle.dumps(flow)
+
+        session_key ='google_oauth2_flow_{0}'.format(self.CSRF_TOKEN)
```
dhermes: I was wondering why this showed up in the diff, then I noticed you're missing a space after the `=` sign

tests/contrib/django_util/test_views.py
```diff
@@ -181,8 +183,8 @@ def test_callback_works(self, pickle):
             response.status_code, django.http.HttpResponseRedirect.status_code)
         self.assertEqual(response['Location'], self.RETURN_URL)
 
-    @mock.patch('oauth2client.contrib.django_util.views.pickle')
-    def test_callback_handles_bad_flow_exchange(self, pickle):
+    @mock.patch('oauth2client.contrib.django_util.views.jsonpickle')
+    def test_callback_handles_bad_flow_exchange(self, jsonpickle):
```
dhermes: Ditto here, can you verify the mock was called and name the var. like a mock?

tox.ini
```diff
@@ -14,6 +14,7 @@ basedeps = mock>=1.3.0
 deps = {[testenv]basedeps}
        django
        keyring
+       jsonpickle
```
dhermes, line 17: Will `django` users have this dependency? Should we add some extras to `setup.py` for people that want the `django` dependencies?

waprin, line 17: they won't have any particular reason to have it. what do you have in mind? not totally sure how extra dependencies work.

tests/contrib/django_util/test_views.py
waprin: done

tests/contrib/django_util/test_views.py
waprin: done

tests/contrib/django_util/test_views.py
waprin: done

tests/contrib/django_util/test_views.py
waprin: done

oauth2client/contrib/django_util/views.py
waprin: done

tox.ini
dhermes, line 17: See an [example](https://github.com/GoogleCloudPlatform/gcloud-python/blob/f02d7a28cc835790cf7991da7323416769463888/setup.py#L49) of extras. You'd just give a list like

``` python
DJANGO_EXTRAS = ('django', 'jsonpickle')

setup(
    ...
    extras_require={'django': DJANGO_EXTRAS},
    ...
)
```

and then people could install via

```
pip install oauth2client[django]
```

waprin, line 17: I do think making the contrib stuff more optional, as well as improving tox so it runs the tests on the django version/python version matrix would be good. But I think given this is a bugfix maybe hold that change for a separate PR?

oauth2client/contrib/django_util/views.py
```diff
@@ -22,20 +22,21 @@
...
 from django.shortcuts import redirect
+import jsonpickle
 from six.moves.urllib import parse
 
 from oauth2client import client
 from oauth2client.contrib import django_util
 from oauth2client.contrib.django_util import get_storage
 from oauth2client.contrib.django_util import signals
 
+
```
dhermes: I don't have a preference here but I think @nathanielmanistaatgoogle wants 1 line between imports and the constants? I dunno.

tests/contrib/django_util/test_views.py
```diff
@@ -152,8 +152,8 @@ def setUp(self):
         self.user = User.objects.create_user(
             username='bill', email='bill@example.com', password='hunter2')
 
-    @mock.patch('oauth2client.contrib.django_util.views.pickle')
-    def test_callback_works(self, pickle):
+    @mock.patch('oauth2client.contrib.django_util.views.jsonpickle')
+    def test_callback_works(self, jsonpickle_mock):
```
dhermes, line 156: Thanks for the rename, still no verification of how the mock was used though

tests/contrib/django_util/test_views.py
```diff
@@ -198,17 +199,18 @@ def test_callback_handles_bad_flow_exchange(self, pickle):
...
             raise FlowExchangeError('test')
 
         flow.step2_exchange = local_throws
-        pickle.loads.return_value = flow
+        jsonpickle_mock.decode.return_value = flow
 
         request.session = self.session
         response = views.oauth2_callback(request)
         self.assertIsInstance(response, http.HttpResponseBadRequest)
+        self.assertTrue(jsonpickle_mock.encode.called)
```
dhermes: Can you `assert_called_once_with` to be more specific

tox.ini
dhermes, line 17: :+1: to a separate PR

tests/contrib/django_util/test_views.py
waprin: Hmm shouldn't we stick with unittest methods though? Or is that actually a method I can call on self, or will calling it on the mock play nice with the unittest test runner?

waprin: of `self.assertEqual(jsonpickle_mock.called_count, 1)` perhaps?

dhermes: Calling on the mock will place nice with the test runner.

dhermes: No `assert_called_once_with` is what is desired. It actually specifies which arguments were passed in.

tests/contrib/django_util/test_views.py
waprin, line 156: done

tests/contrib/django_util/test_views.py
waprin: ah ok didn't realize it was ok not to use the self.asserts. done.

tests/contrib/django_util/test_views.py
```diff
@@ -180,9 +181,10 @@ def test_callback_works(self, pickle):
         self.assertEqual(
             response.status_code, django.http.HttpResponseRedirect.status_code)
         self.assertEqual(response['Location'], self.RETURN_URL)
+        jsonpickle_mock.decode.assert_called_once_with(pickled_flow)
```
dhermes, line 184: Was `encode` called (other than your `jsonpickle_mock.encode(flow)` above)? Also, what is the point of that? Wouldn't `pickled_flow = object()` work just fine?

waprin, line 184: Done, yes now that you mention it setting a value to the return of a mock you just created _is_ kind of silly. I'm guessing there was an iteration of the test that at some point made sense and then not-thoughtful refactors made it otherwise, thanks.

oauth2client/contrib/django_util/views.py
```diff
@@ -71,7 +71,7 @@ def _make_flow(request, scopes, return_url=None):
             urlresolvers.reverse("google_oauth:callback")))
 
     flow_key = _FLOW_KEY.format(csrf_token)
-    request.session[flow_key] = pickle.dumps(flow)
+    request.session[flow_key] = jsonpickle.encode(flow)
```
waprin: sorry need to change here too

waprin: err misread this is not the test

dhermes:
LGTM. I'll let someone else have a pass first before merging

waprin:
@jonparrott hey, i just met you, and this is crazy, but LGTM, so merge me maybe?

jonparrott:
:D I never get to use any Carly Rae Jepsen gifs.

carly

PR #616, opened 2016-08-11
Change status access from header to attribute.

dhermes:
This is to unify all the times a status is used from the response (containing header info) of an HTTP request.

---

@jonparrott @nathanielmanistaatgoogle I was considering adding `transport.get_status(response)` and `transport.get_header(response)`, these are essentially the only remnants of `httplib2` outside of the `transport` module. (The only other is that `http.request` returns `response, content` and that `content` is always `bytes`.)


dhermes:
@jonparrott @nathanielmanistaatgoogle Can you weigh in on `transport.get_status(response)` and `transport.get_header(response)`?

jonparrott:
Let's table it until we actually try using a different transport.

dhermes:
SGTM. I'm more concerned about flagging / fencing off all the `httplib2` features.

PR #622, opened 2016-08-11
Correct query loss when using parse_qsl to dict

dhermes:
**NOTE**: I noticed this when doing a fairly comprehensive source check to find the `httplib2` footprint (for #128).

The issue is that repeated params could just be dropped:

``` python
>>> from six.moves import urllib
>>> params_list = urllib.parse.parse_qsl('a=1&a=2')
>>> params_list
[('a', '1'), ('a', '2')]
>>> dict(params_list)
{'a': '2'}
```

### DIFF
--- a/oauth2client/_helpers.py
+++ b/oauth2client/_helpers.py
@@ -179,6 +179,54 @@ def string_to_scopes(scopes):
         return scopes
 
 
+def parse_unique_urlencoded(content):
+    """Parses unique key-value parameters from urlencoded content.
+
+    Args:
+        content: string, URL-encoded key-value pairs.
+
+    Returns:
+        dict, The key-value pairs from ``content``.
+
+    Raises:
+        ValueError: if one of the keys is repeated.
+    """
+    urlencoded_params = urllib.parse.parse_qs(content)
+    params = {}
+    for key, value in six.iteritems(urlencoded_params):
+        if len(value) != 1:
+            msg = ('URL-encoded content contains a repeated value:'
+                   '%s -> %s' % (key, ', '.join(value)))
+            raise ValueError(msg)
+        params[key] = value[0]
+    return params
+
+
+def update_query_params(uri, params):
+    """Updates a URI with new query parameters.
+
+    If a given key from ``params`` is repeated in the ``uri``, then
+    the URI will be considered invalid and an error will occur.
+
+    If the URI is valid, then each value from ``params`` will
+    replace the corresponding value in the query parameters (if
+    it exists).
+
+    Args:
+        uri: string, A valid URI, with potential existing query parameters.
+        params: dict, A dictionary of query parameters.
+
+    Returns:
+        The same URI but with the new query parameters added.
+    """
+    parts = urllib.parse.urlparse(uri)
+    query_params = parse_unique_urlencoded(parts.query)
+    query_params.update(params)
+    new_query = urllib.parse.urlencode(query_params)
+    new_parts = parts._replace(query=new_query)
+    return urllib.parse.urlunparse(new_parts)
+
+
 def _add_query_parameter(url, name, value):
     """Adds a query parameter to a url.
 
@@ -195,11 +243,7 @@ def _add_query_parameter(url, name, value):
     if value is None:
         return url
     else:
-        parsed = list(urllib.parse.urlparse(url))
-        query = dict(urllib.parse.parse_qsl(parsed[4]))
-        query[name] = value
-        parsed[4] = urllib.parse.urlencode(query)
-        return urllib.parse.urlunparse(parsed)
+        return update_query_params(url, {name: value})
 
 
 def validate_file(filename):
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -438,23 +438,6 @@ class Storage(object):
             self.release_lock()
 
 
-def _update_query_params(uri, params):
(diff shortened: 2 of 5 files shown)


jonparrott:
This LGTM, but I wonder if there's not already something that handles this well?

dhermes:
I just did `git grep parse_q`. It's crazy it was used incorrectly in 4 different places.

dhermes:
@jonparrott I want to hold off on the merge until @nathanielmanistaatgoogle weighs in

jonparrott:
@dhermes good plan. This seems like hackery I'm not sure I like.

dhermes:
> This seems like hackery I'm not sure I like.

What is "this"?

jonparrott:
Doing urlparsing and such.

dhermes:
Yeah I dunno. You can file an issue to follow-up?

jonparrott:
Just more to think about before we close #128 I think.

dhermes:
Currently typing one-handed w/baby in my arms, can you add a checkbox about it in the long desc. for #128 for me?

jonparrott:
I read that as "typing with baby arms". :D 

Will do.

oauth2client/_helpers.py
```diff
@@ -199,12 +199,7 @@ def _add_query_parameter(url, name, value):
     """
     if value is None:
```
nathanielmanistaatgoogle, line 243: Does

`return url if value is None else update_query_params(url, {name: value})`

fit? If not, retain the `else:` in the `if:`.

oauth2client/_helpers.py
```diff
@@ -300,3 +295,55 @@ def _urlsafe_b64decode(b64string):
     b64string = _to_bytes(b64string)
     padded = b64string + b'=' * (4 - len(b64string) % 4)
     return base64.urlsafe_b64decode(padded)
+
+
+def update_query_params(uri, params):
+    """Updates a URI with new query parameters.
+
+    If a given key from ``params`` is repeated in the ``uri``, then
```
nathanielmanistaatgoogle: Sorcerery!

These are really strange semantics; is there some resource to which you can link that explains why this is correct or useful? (Yes, there's a lot I don't know about the web.)

oauth2client/_helpers.py
```diff
@@ -300,3 +295,55 @@ def _urlsafe_b64decode(b64string):
     b64string = _to_bytes(b64string)
     padded = b64string + b'=' * (4 - len(b64string) % 4)
     return base64.urlsafe_b64decode(padded)
+
+
+def update_query_params(uri, params):
```
nathanielmanistaatgoogle: When a code element is both introduced to a file and used in the same file, please place the code element earlier in the file ("above") all of its use-sites.

dhermes: That is provably impossible for all possible code that can be written, but doable here.

oauth2client/_helpers.py
dhermes: I don't have an opinion on what should happen, so just tell me what you want. The issue is that casting the results of `parse_qsl` (a list of tuples, with repeats allowed) directly to a `dict` is that query params can be lost.

It's possible / probable that all call sites assumed the same thing that `parse_unique_urlencoded` does and then meant to replace (if existing) or add the param.

dhermes:
@nathanielmanistaatgoogle PTAL.

In particular, we should resolve what we think the actual goal of all those `dict(parse_qsl(...))` calls were.

nathanielmanistaatgoogle:
Agreed that this needs a deeper look; probably site-by-site (luckily there are few). Replace-except-if-the-value-is-a-list-of-two-or-more-then-add is just weird semantics.

If it's the case that all sites in the library are correct with replace semantics, then our library-private `_helpers` module should make available a helper function that implements replace semantics.

If it's the case that all sites in the library are correct with add-to semantics, then our library-private `_helpers` module should make available a helper function that implements add-to semantics.

If it's the case that some sites in the library are correct with replace semantics and some are correct with add-to semantics, then our library-private `_helpers` module should make available a helper function that implements replace semantics _and_ a helper function that implements add-to semantics.

I seriously doubt that it's the case that there's a site in the library where the correct semantics are replace-except-if-the-value-is-a-list-of-two-or-more-then-add.

Am I making sense?

dhermes:
Yes. But the issue with replace is if the original had multiple values, then what does replace mean? Which is why I suggested we first verify that each parameter shows up exactly once, then replace.

i.e. `a=1&b=2` replace `b=3` (or `c=3`) is easy but what if the original was `b=1&b=2`

nathanielmanistaatgoogle:
For private helper methods, replace means what we implement it to mean. I could see it going one of two ways: either "replace" semantics mean "associate this key with this value, no matter whether zero, one, two, or more values were associated with the key previously" or "replace" semantics mean "associate this key with this value, and if there were zero values with the key previously, that's fine, and if there was one value with the key previously, that's also fine, but if there were two or more values associated with the key previously, that's a problem so raise an exception". While either of these choices would be acceptable to be blessed as correct, I suspect it's the first that is going to be more useful and applicable in our codebase. Is my suspicion correct?

dhermes:
@nathanielmanistaatgoogle Hopefully this works. Don't forget to remind me to squash on LGTM

nathanielmanistaatgoogle:
Looks good (after out-of-band discussion); squash and merge.

PR #624, opened 2016-08-11
Remove `unittest2`

pferate:
This test snuck by the recent removal of `unittest2`


jonparrott:
Thanks! Sorry for letting this slip through.

PR #625, opened 2016-08-11
Import clean up

pferate:
- Treat `tests` as part of the local package (for import order)
- Import modules instead of objects (some slipped by since this was cleaned up previously)


jonparrott:
lgtm. I don't know why django-related code always looks so gross to me.

jonparrott:
(waiting for the test god to merge)

pferate:
Flask > Django?

jonparrott:
> Flask > Django?

At least to me. I've never been much of a fan of Django (though I love their community!)

PR #627, opened 2016-08-11
Remove __author__ lines, add contributors.md

### LINKED ISSUE #626: Remove __author__ from all files
Might be worth putting a section in `CONTRIBUTING` giving a shout to all the author's over time?

theacodes:
Resolves #626

### DIFF
--- a/oauth2client/_helpers.py
+++ b/oauth2client/_helpers.py
@@ -26,11 +26,6 @@ import six
 from six.moves import urllib
 
 
-__author__ = (
-    'rafek@google.com (Rafe Kaplan)',
-    'guido@google.com (Guido van Rossum)',
-)
-
 logger = logging.getLogger(__name__)
 
 POSITIONAL_WARNING = 'WARNING'
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,0 +1,95 @@
+# Contribors to oauth2client
+
+## Maintainers
+
+* [Nathaniel Manista](https://github.com/nathanielmanistaatgoogle)
+* [Jon Wayne Parrott](https://github.com/jonparrott)
+* [Danny Hermes](https://github.com/dhermes)
+
+Previous maintainers:
+
+* [Craig Citro](https://github.com/craigcitro)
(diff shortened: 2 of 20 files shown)


jonparrott:
I knew our author lines were inaccurate at best, but they were vastly underrepresenting the people who have contributed to this library.

dhermes:
Insane number of contributors!

Do we want a special section for past maintainers like Joe and Craig?

dhermes:
LGTM save for my comment.

We should wait for @nathanielmanistaatgoogle to weigh in

oauth2client/_helpers.py
```diff
@@ -26,11 +26,6 @@
 from six.moves import urllib
 
 
-__author__ = (
-    'rafek@google.com (Rafe Kaplan)',
-    'guido@google.com (Guido van Rossum)',
-)
-
```
waprin, line 33: These guys are getting removed from here but aren't in contributors section, not sure if intentional?

jonparrott:
Happy to add a previous maintainers section if both you and @nathanielmanistaatgoogle  think it's important. :)

oauth2client/_helpers.py
theacodes, line 33: neither have actual commits in the repo. Looking into the history, it's likely that the very first version of this file was copied from elsewhere in Google and guido wrote that file.

dhermes, line 33: They came from `protorpc.util`, where `@positional` was cribbed from

waprin, line 33: Ok, if code they wrote is used in the repo seems like a very small and easy bone to throw them to put them in the contributors file. Whoever this Guido guy is could probably use the acknowledgement and recognition.

pferate:
I agree with @dhermes on having a separate section for former maintainers.

oauth2client/_helpers.py
theacodes, line 33: Done.

jonparrott:
Former maintainers section added. @nathanielmanistaatgoogle PTAL?

PR #629, opened 2016-08-12
Complete branches from partial test coverages

pferate:
Some tests didn't cover all branches of flow, so this is to fix that.


pferate:
D'Oh!  Gotta fix some Python 3 issues.

jonparrott:
Thanks, @pferate!

Issue #596, opened 2016-08-09
PyPy tests fail after first run


pferate:
Looks like this was introduced in #590.

Travis was configured to cache the `.pyenv` directory between builds and cloning the repo fails if the directory is not empty.

We have a few options on how to proceed:

1. Not cache the directory between builds (the previous behavior)
2. Only clone the `pyev` repo if the directory doesn't exist (similar to App Engine SDK)

jonparrott:
Fixed by #595.

pferate:
@jonparrott Looks like Travis creates the cached directories if they don't exists, so the check in #595 didn't work for me.  Created #630.

PR #595, opened 2016-08-09
Fix pypy environment in Travis.

theacodes:

PR #630, opened 2016-08-15
Check for `bin` directory before cloning pyenv

pferate:
Travis creates cached directories if they are not present, so the pyenv repo was never getting cloned if the cache was empty.

Related to #595 and #596 

cc/ @jonparrott


jonparrott:
Thanks, @pferate!

PR #631, opened 2016-08-15
Update helper b64 encode/decode tests

pferate:
While looking at some other tests, I noticed an inconsistency in `Test__urlsafe_b64decode.test_valid_input_unicode()`.  This addresses that and a few more related inconsistencies.


jonparrott:
Thanks again, @pferate!

PR #647, opened 2016-09-15
Remove transport.get_cached_http, add certificate cache

theacodes:


jonparrott:
(I love deleting code)

nathanielmanistaatgoogle (approved):

oauth2client/client.py
```diff
@@ -1533,15 +1549,16 @@ def verify_id_token(id_token, audience, http=None,
...
 
-    resp, content = transport.request(http, cert_uri)
-    if resp.status == http_client.OK:
-        certs = json.loads(_helpers._from_bytes(content))
-        return crypt.verify_signed_jwt_with_certs(id_token, certs, audience)
-    else:
-        raise VerifyJwtTokenError('Status code: {0}'.format(resp.status))
+    try:
+        certs = _get_certificates(http, cert_uri)
+    except http_client.HTTPException as e:
```
dhermes, line 1558: Blerg! The `as e` killing me. Why the allergy to actual names?

dhermes:
@jonparrott It just occurred to me reading this that the old implementation never flushed the cache.

I guess there is no mention of a time-out in:
https://www.googleapis.com/oauth2/v1/certs

oauth2client/client.py
theacodes: Do you have a better recommendation? I've honestly never seen anyone else complain about `e` except in cases where there are multiple exceptions around.

dhermes: `exc`

jonparrott:
@dhermes indeed. We can file an issue to revisit it, if you want?

oauth2client/client.py
theacodes: (Fixed in `transport-refactor` branch)

dhermes:
Naw. I seemed to recall endpoints had an eviction policy. They use the GAE `memcache` service for caching:
https://github.com/cloudendpoints/endpoints-python/blob/067490edee4d157f2ff919d4d3e85d38d73a832c/endpoints/users_id_token.py#L200-L201

Also the evict bashed on the cache control header:
https://github.com/cloudendpoints/endpoints-python/blob/067490edee4d157f2ff919d4d3e85d38d73a832c/endpoints/users_id_token.py#L434

dhermes:
The `oauth2client` certs URI also has `max-age=23377` in the cache control.

```python
>>> import requests
>>> resp = requests.get('https://www.googleapis.com/oauth2/v1/certs')
>>> print(json.dumps(dict(resp.headers), indent=2))
{
  "content-length": "2145", 
  "x-xss-protection": "1; mode=block", 
  "x-content-type-options": "nosniff", 
  "content-encoding": "gzip", 
  "age": "2346", 
  "expires": "Fri, 16 Sep 2016 22:30:03 GMT", 
  "vary": "Origin, X-Origin", 
  "server": "GSE", 
  "cache-control": "public, max-age=23377, must-revalidate, no-transform", 
  "date": "Fri, 16 Sep 2016 16:00:26 GMT", 
  "x-frame-options": "SAMEORIGIN", 
  "alt-svc": "quic=\":443\"; ma=2592000; v=\"36,35,34,33,32\"", 
  "content-type": "application/json; charset=UTF-8"
}
```

Seems like we need to file a bug? (Feeling :sheep:-ish)

jonparrott:
Filed #649

PR #607, opened 2016-08-10
Use transport.request in tests.

dhermes:
In the process
- "spring clean" the modules that were touched
- use `HttpMock` when `HttpMockSequence` not needed
- add some verifications on new `HttpMock`'s

PR #648, opened 2016-09-15
	Remove inject_assertion_credentials, add Credentials.before_request

theacodes:
This one is a big one:
1. Add a new method to `OAuth2Credentials`: `before_request`. This is used to apply headers and do any pre-request logic.
2. Add an override for `before_request` to `_JWTAccessCredentials` that allows it to generate a one-time token.
3. Remove separate `inject_assertion_credential` method, as the differing logic is now contained in the `before_request` methods.

@dhermes is concerned about the new method being public. I don't care if it's public or private so I'll defer to you two. However, I do really like that `transport` no longer calls _any_ private methods on `Credentials`.


oauth2client/service_account.py
```diff
@@ -683,3 +664,41 @@ def _create_token(self, additional_claims=None):
...
+    def apply(self, headers, access_token=None):
+        """Add the access token as the Authorization header in headers.
+
+        Args:
+            headers: dict, the headers to add the Authorization header to.
+            access_token: An access token to use instead of the credential's
+                current access token.
+        """
+        if not access_token:
+            access_token = self.access_token
```
dhermes, line 679: I forget, why is `self.access_token` insufficient?

oauth2client/service_account.py
```diff
@@ -683,3 +664,41 @@ def _create_token(self, additional_claims=None):
...
+        if 'aud' in self._kwargs:
+            # Preemptively refresh token. This is not done for regular
+            # OAuth2 credentials, but is useful here because it avoids
+            # unnecessary requests.
+            if self.access_token is None or self.access_token_expired:
+                self.refresh(None)
+            self.apply(headers)
+        else:
+            # Generate a one-time token using the request URI as the audience.
+            uri_root = uri.split('?', 1)[0]
```
dhermes: This is an error prone way to parse a URI.

oauth2client/transport/httplib2.py
```diff
@@ -62,98 +62,56 @@ def inject_credentials(credentials, http, refresh_status_codes):
...
 
-        return resp, content
+            credentials.refresh(orig_request_method)
 
-    # Replace the request method with our own closure.
-    http.request = new_request
+            # Remove the existing Authorization header, as the credentials
+            # may set the header as u'Authorization' but clean_headers set it
+            # as b'Authorization'.
+            headers.pop(b'Authorization', None)
```
dhermes, line 100: Is this necessary? Pretty hacky

tests/conftest.py
```diff
@@ -14,9 +14,14 @@
 
 """Py.test hooks."""
 
+import logging
+
 from oauth2client import _helpers
 
 
+logging.basicConfig(level=logging.INFO)
```
dhermes, line 22: Why?

tests/test_service_account.py
```diff
@@ -601,19 +601,23 @@ def test_authorize_stale_token(self, utcnow):
...
+        # Third request should use the second access token.
         http = http_mock.HttpMockSequence([
             ({'status': http_client.OK}, b''),
             ({'status': http_client.UNAUTHORIZED}, b''),
             ({'status': http_client.OK}, b''),
         ])
-        self.jwt.authorize(http)
-        transport.request(http, self.url)
+        http = self.jwt.authorize(http)
+        http.request(self.url)
```
dhermes, line 616: Please use `transport.request` (in other words, don't unravel my work in #607)

tests/test_service_account.py
```diff
@@ -601,19 +601,23 @@ def test_authorize_stale_token(self, utcnow):
...
         token_1 = self.jwt.access_token
 
         utcnow.return_value = T2_DATE
-        response, _ = transport.request(http, self.url)
+        response, _ = http.request(self.url)
```
dhermes, line 620: Ditto

tests/test_service_account.py
theacodes: I am intentionally doing this, as these tests should match the user's usage, not the library's internal usage:

```
http = {httplib2.Http(), urllib3.PoolManager(), requests.Session(), http_mock.HttpMock()}
authed_http = credentials.authorize(http)
response, data = authed_http.request(...)
```

(I will clean up all of these once I have a separate strong test for `transport.httplib2.request`)

tests/conftest.py
theacodes: so that if a test fails, all of the logs will show up in pytest's captured output.

oauth2client/transport/httplib2.py
theacodes: For now, I'm going to make sure that all credential classes set headers as bytestrings instead of unicode in a follow-up.

oauth2client/service_account.py
theacodes: see the `before_request` logic below, sometimes the credentials need to generate a one-time token for the url. It passes that one-time token into here.

oauth2client/service_account.py
theacodes: I just copypasted but I can switch to urlparse if you'd like.

oauth2client/service_account.py
dhermes: Ah.

oauth2client/service_account.py
dhermes: Just file issue?

oauth2client/transport/httplib2.py
dhermes: So is it necessary?

tests/test_service_account.py
dhermes: Can you do that then, i.e.

``` python
authed_http = self.jwt.authorize(http)
```

theacodes: (That is to say, the interface for transport.request is still in flux, so sit tight)

oauth2client/service_account.py
theacodes: It's a super easy fix, will do.

oauth2client/transport/httplib2.py
theacodes: It is for now, because the tests will non-deterministically fail.

dhermes: I feel like I'm falling down a rabbit hole. Why do they non-deterministically fail?

tests/test_service_account.py
theacodes: @dhermes I'm going to do that in the next next PR to remove the monkey-patching. Can you sit tight with the promise that I've already done this in a later commit?

dhermes: [Show me the commit](https://www.youtube.com/watch?v=FCgy0dphOGc)

oauth2client/transport/httplib2.py
theacodes: So here's what happens:
1. First request
   1. `credentials.apply` sets `headers[u'Authorization']`
   2. `clean_headers` renames `headers[u'Authorization']` to `headers[b'Authorization']`
   3. Receive 401 response, so retry the request. _The current headers are re-used_.
2. Retry request.
   1. `credentials.refresh` is called.
   2. `credentials.apply` sets `headers[u'Authorization']`.
   3. At this point `headers` is `{b'Authorization': 'old token', u'Authorization': 'new token'}`
   4. clean_headers may or may not use the second one or the first one when sanitizes because of non-deterministic dictionary ordering.

Imagine how long it too me to figure this out. :)

tests/test_service_account.py
theacodes: https://github.com/jonparrott/oauth2client/blob/tr-3-http-wrap/tests/test_service_account.py#L615

oauth2client/service_account.py
theacodes: Fixed. :)

tests/test_service_account.py
dhermes: f516ce33039690b82d4447112bb62346cd91481e

dhermes: Thanks

oauth2client/transport/httplib2.py
dhermes: So why don't we use `b'Authorization'` instead of `u'Authorization'` in all of those places?

theacodes: I plan to in a separate PR. I can file a bug if you want (although it's on my list of things todo before I call this finished).

dhermes: Please do. I don't have much fail in human memory.

theacodes: Filed #650

(It's not in my head, it's on a big txt file that I'm nursing while doing this)

jonparrott:
@dhermes @nathanielmanistaatgoogle any thoughts on making `before_request` public or private? As I mentioned, the only thing I have in favor of `before_request` being public is that `transport` currently doesn't have to use any private methods, but considering it's part of oauth2client it's not a big deal if it does.

dhermes:
@jonparrott I'm always for making private if not user-facing.

jonparrott:
@nathanielmanistaatgoogle PTAL when you get a chance. :)

oauth2client/service_account.py
```diff
@@ -683,3 +666,44 @@ def _create_token(self, additional_claims=None):
...
+        """Add the access token as the Authorization header in headers.
+
+        Args:
+            headers: dict, the headers to add the Authorization header to.
+            access_token: An access token to use instead of the credential's
+                current access token.
+        """
+        if not access_token:
+            access_token = self.access_token
+        headers['Authorization'] = 'Bearer ' + access_token
```
nathanielmanistaatgoogle, line 680: I happen to work on [an HTTP/2-based library](http://www.grpc.io/) and recently learned in the course of a painful user-visible outage that [in HTTP/2 all header names must contain no majuscule characters](http://http2.github.io/http2-spec/index.html#rfc.section.8.1.2). Can this header name be "authorization"? Do we support use of `oauth2client` with HTTP/2? How aware are we that `oauth2client` is already in use with HTTP/2 whether we support it or not?

oauth2client/transport/httplib2.py
```diff
@@ -62,98 +62,56 @@ def inject_credentials(credentials, http, refresh_status_codes):
...
-            credentials._refresh(orig_request_method)
 
         # Clone and modify the request headers to add the appropriate
         # Authorization header.
         headers = _helpers.initialize_headers(headers)
-        credentials.apply(headers)
         _helpers.apply_user_agent(headers, credentials.user_agent)
 
+        credentials.before_request(
+            orig_request_method, uri, headers)
```
nathanielmanistaatgoogle: Unnecessary line-wrap?

oauth2client/transport/httplib2.py
nathanielmanistaatgoogle: Let me just go ahead and make life harder by adding letter-case to this conversation!

oauth2client/transport/httplib2.py
```diff
@@ -62,98 +62,56 @@ def inject_credentials(credentials, http, refresh_status_codes):
...
 
-    # Set credentials as a property of the request method.
-    http.request.credentials = credentials
-
-
-def inject_assertion_credentials(
-        credentials, http, refresh_status_codes):
-    """Prepares an HTTP object's request method for JWT access.
+            credentials.before_request(
+                orig_request_method, uri, headers)
```
nathanielmanistaatgoogle: Unnecessary line-wrap?

tests/http_mock.py
```diff
@@ -101,6 +101,7 @@ def request(self, uri,
                 redirections=1,
                 connection_type=None):
         resp, content = self._iterable.pop(0)
+        headers = headers.copy() if headers is not None else headers
```
nathanielmanistaatgoogle: The possible assignment of `None` reads funny to me here. Why shouldn't this line be

```
headers = {} if headers is None else dict(headers)
```

?

oauth2client/service_account.py
dhermes: At current, `httplib2`[ normalizes](https://github.com/jcgregorio/httplib2/blob/010b4e399c04a199e7ca25995286f3db0337500d/python2/httplib2/__init__.py#L263-L264) the headers to be lowercase

theacodes: I will do a follow-up PR soon to normalize all header usage in oauth2client to lowercase bytestrings, cool?

oauth2client/transport/httplib2.py
theacodes: Fixed

oauth2client/transport/httplib2.py
theacodes: Fixed

tests/http_mock.py
theacodes: Fixed

jonparrott:
@nathanielmanistaatgoogle comments addressed.

There's still the open issue of if `before_request` should be public or not.

(I'm punting the header issue because I really want to keep each of these PRs as small and focused as possible to avoid every PR spiraling into trying to fix all the problems at once. I am keeping a list of things that need to be fixed before the final merge)

oauth2client/service_account.py
nathanielmanistaatgoogle: Supercool.

nathanielmanistaatgoogle:
`before_request` can't be library-public without a compelling use case. Preferably more than one. Am I understanding correctly that we'd like it to be class-public-but-library-private? Since we lack a means to express that visibility (right?) it should be class-private and we'll just live with (`TODO`? issue?) the abstraction violation.

jonparrott:
@nathanielmanistaatgoogle I don't expect consumers of this library to call it, but they make want to override it for custom credential classes.

It seems like you're saying you'd prefer it to be private, so I'll make that change.

jonparrott:
Done.

jonparrott:
@nathanielmanistaatgoogle @dhermes any reservations or can I merge this and send over the next PR?

dhermes:
None

nathanielmanistaatgoogle:
I have larger questions that don't have to be addressed as part of this pull request but I think should be part of the 4.0 story or later, but let's lay them out now:

1. Affording developer-users an opportunity to change the behavior of an implementation provided by us in a small, parameterized way seems like a classic inheritance-versus-composition scenario. Why would providing a method to be (possibly) overridden be better than providing a construction parameter to be (possibly) passed?
2. Is it forecast that overriding implementations of the method will need to `super`-call the overridden implementation, want to in some circumstances, or never need to? If it weren't for the use of inheritance structurally forcing subclass authors' having the freedom to call the overridden implementation, would we require them to do so, require them not to do so, or neither?

jonparrott:
For (1), the implementation of `before_request` is likely credential-class specific, and it fits in well with my exploration into the post-4 refactor.

For (2), during my post-4 refactor tests, no subclasses have needed to call super for before_request. Only one class overrides it, and behaves quite differently.

PR #651, opened 2016-09-16
Fix django authorization redirect by correctly checking validity of credentials

waprin:
I was checking for credentials but I need to use the `has_credentials` method for cases of existing but invalid credentials.

### DIFF
--- a/oauth2client/contrib/django_util/views.py
+++ b/oauth2client/contrib/django_util/views.py
@@ -170,7 +170,10 @@ def oauth2_authorize(request):
          A redirect to Google OAuth2 Authorization.
     """
     return_url = request.GET.get('return_url', None)
+    if not return_url:
+        return_url = request.META.get('HTTP_REFERER', '/')
 
+    scopes = request.GET.getlist('scopes', django_util.oauth2_settings.scopes)
     # Model storage (but not session storage) requires a logged in user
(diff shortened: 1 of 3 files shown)


theacodes (approved):

jonparrott:
@waprin looks like we have a coverage miss:

```
oauth2client/contrib/django_util/views.py              73      1     14      1    98%   187, 186->187
```

Can you fix?

waprin:
@jonparrott Looks like all my tests pass and django_util now has 100% coverage, failing on cover because of other code I think?

waprin:
Ah never mind, now I see still some missing lines in the tests themselves

oauth2client/contrib/django_util/views.py
```diff
@@ -170,21 +170,22 @@ def oauth2_authorize(request):
          A redirect to Google OAuth2 Authorization.
     """
     return_url = request.GET.get('return_url', None)
+    if not return_url:
```
nathanielmanistaatgoogle, line 173: Prefer `is None`/`is not None` when comparing to the `None` object.

If `request.GET.get('return_url', None)` is the empty string, that's a bug, right? And by unintentionally sweeping it along the same code path as `request.GET`-does-not-contain-a-key-value-mapping-for-key-`return_url` we make finding and fixing the bug harder, right?

Issue #597, opened 2016-08-09
[Discussion] post 4.0.0 refactor


theacodes:
With 3.0.0 out and 4.0.0 on the way, we've finally addressed most of the "cruft" in this library without making huge, sweeping, breaking changes. However, some huge obstacles remain - such as the twisted class hierarchy and the strange naming of some of the credentials classes.

I'd like to start a discussion on refactoring this library post-4.0.0. I'm not one to throw out well-tested and proven code, but the code that is here needs to be better organized to suit our downstream clients and our users.

I want to put forth the idea of this package slowly migrating to two new packages: `google.auth` and `google.oauth2`. During the initial phases we'll retain code inside of and continue to publish the `oauth2client` package, but once complete the `oauth2client` package will be permanently deprecated.

The `google.auth` package will focus solely on server-to-server authentication. A rough idea of what this package will look like is:
- `google.auth.default()` returns the application default credentials.
- `google.auth.service_account.Credentials` credentials using a service account key to obtain an access token.
- `google.auth.jwt.Credentials` credentials using a service account key to directly assert credentials.
- `google.auth.gae.Credentials` credentials using App Engine identity credentials.
- `google.auth.gce.Credentials` credentials using the Compute Engine metadata server.
- `google.auth.access_token.Credentials` credential using a bare access token.

This package won't contain anything related to storage or end-user based credentials.

The `google.oauth2` package will contain the remainder of the current library and focus solely on oauth2 user-specific stuff, better organized:
- `google.oauth2.flow` - generic flow that can be easily used with web frameworks.
- `google.oauth2.credentials` - user credentials with access and refresh tokens.
- `google.oauth2.storage` - a wrapper that allows credentials to be stored.
  etc.

What are you initial thoughts @nathanielmanistaatgoogle @dhermes @waprin @thobrla @anthmgoogle @elibixby @pferate  & any others.

waprin:
First pass sounds reasonable to me. I assume the work you're doing on GAE SDK regarding the `google` package being properly namespaces is a prereq to this. Is there a need to protect this library from any other `google` packages that *dont* do it correctly? I worry that even if you do things "right" and others do it wrong, it's poor consolation if your library won't install correctly.

I also would like people to consider the topic of the Storage classes, and how `get` does not take a key as a parameter. The "construct with a key then re-use" always seemed awkward to me and forced awkward code. Wondering if it's just me.

jonparrott:
[...]

PR #654, opened 2016-09-20
Solidify transport interface

theacodes:
These changes solidify the transport interface.
- Add a transport compliance test (see note below). This is use to qualify transport implementations work as expected.
- Bring httplib2 transport into compliance.
- Make `transport._request` private.
- Define the `transport._request` return value.
- Remove all usage of `transport.request` in tests in favor of using `authed_http` directly, just as users would.

Note about compliance tests:
This is the first test in the module to use pytest fixtures as well as use pytest's test & assertion styles. We eventually want to move toward pytest style tests (I plan to do so as I refactor each module for #597), so I'm hoping this won't be controversial. If it is, I'm happy to switch to unittest-style, but I'll be keeping the extremely useful pytest-localserver fixture.

(Also I'm starting to see the light at the end of the tunnel here)


tests/transport/test_transport_compliance.py
```diff
@@ -0,0 +1,153 @@
...
+        yield
+        server.stop()
+
+    def test_request_basic(self):
+        http = self.transport.get_http_object()
+        response = self.transport._request(
+            http, self.server.url + '/basic')
+
+        assert response.status == http_client.OK
+        print(response.headers.keys())
```
dhermes: Why the `print()`?

tests/transport/test_transport_compliance.py
```diff
@@ -0,0 +1,153 @@
...
+        self.user_agent = 'test-credentials'
+
+    def _before_request(self, http, uri, headers):
+        headers[b'authorization'] = 'Bearer {}'.format(self.token)
+
+    def refresh(self, http):
+        self.token = '{}{}'.format(self.token, '-new')
+
+
+class TransportComplianceTests(object):
```
dhermes, line 76: Why not start name with `Test`? `py.test` style thing?

tests/transport/test_transport_compliance.py
```diff
@@ -0,0 +1,153 @@
...
+        response = self.transport._request(
+            http, self.server.url + '/server_error')
+
+        assert response.status == http_client.INTERNAL_SERVER_ERROR
+        assert response.data == b'Error'
+
+    def test_connection_error(self):
+        http = self.transport.get_http_object()
+
+        with pytest.raises(Exception):
```
dhermes: Can we get a more specific exception here?

oauth2client/transport/httplib2.py
```diff
@@ -135,30 +133,24 @@ def make_authorized_http(credentials, http, refresh_status_codes):
     return _AuthorizedHttp(http, credentials, refresh_status_codes)
 
 
-def request(http, uri, method='GET', body=None, headers=None,
-            redirections=httplib2.DEFAULT_MAX_REDIRECTS,
-            **kwargs):
+class _ResponseWrapper(object):
```
dhermes, line 136: Shouldn't this be in a more general place? Or at least the interface required be documented? (`.data`, `.status` and `.headers`)

Also, where does this desired interface come from?

dhermes: Also I know it's private, but you should still have a docstring

tests/transport/test_httplib2_transport.py
```diff
@@ -50,42 +51,28 @@ def test_wrap(self):
...
 
 class Test_request(unittest.TestCase):
-
-    uri = 'http://localhost'
-    method = 'POST'
-    body = 'abc'
-    redirections = 3
-
-    def test_with_request_attr(self):
+    def test_it(self):
```
dhermes, line 54: I know there is someone who will ask you to rename this.

oauth2client/client.py
```diff
@@ -836,13 +837,13 @@ def _do_revoke(self, http, token):
...
-        if resp.status == http_client.OK:
+        response = transport._request(http, token_revoke_uri)
+        if response.status == http_client.OK:
             self.invalid = True
         else:
-            error_msg = 'Invalid response {0}.'.format(resp.status)
+            error_msg = 'Invalid response {0}.'.format(response.status)
             try:
-                d = json.loads(_helpers._from_bytes(content))
+                d = json.loads(_helpers._from_bytes(response.data))
```
dhermes: Since you dropped by, might as well give it a better name than `d`. `error_dict`?

oauth2client/client.py
```diff
@@ -985,11 +986,12 @@ def _detect_gce_environment():
...
     http = transport.get_http_object(timeout=GCE_METADATA_TIMEOUT)
     try:
-        response, _ = transport.request(
+        response = transport._request(
             http, _GCE_METADATA_URI, headers=_GCE_HEADERS)
         return (
             response.status == http_client.OK and
-            response.get(_METADATA_FLAVOR_HEADER) == _DESIRED_METADATA_FLAVOR)
+            response.headers.get(
+                _METADATA_FLAVOR_HEADER) == _DESIRED_METADATA_FLAVOR)
```
dhermes: These two lines are pretty hard to read. Instead:

``` python
        flavor = response.headers.get(_METADATA_FLAVOR_HEADER)
        return (
            response.status == http_client.OK and
            flavor == _DESIRED_METADATA_FLAVOR)
```

oauth2client/client.py
```diff
@@ -2084,13 +2086,14 @@ def step2_exchange(self, code=None, http=None, device_flow_info=None):
...
-            logger.info('Failed to retrieve access token: %s', content)
+            logger.info('Failed to retrieve access token: %s', response.data)
             if 'error' in d:
                 # you never know what those providers got to say
                 error_msg = (str(d['error']) +
                              str(d.get('error_description', '')))
             else:
-                error_msg = 'Invalid response: {0}.'.format(str(resp.status))
+                error_msg = 'Invalid response: {0}.'.format(
+                    str(response.status))
```
dhermes, line 2096: Isn't `str()` superfluous here since `format()` will do that?

tests/test_client.py
```diff
@@ -901,9 +901,9 @@ def test_recursive_authorize(self):
         token_response = {'access_token': '1/3w', 'expires_in': 3600}
         encoded_response = json.dumps(token_response).encode('utf-8')
         http = http_mock.HttpMock(data=encoded_response)
-        http = self.credentials.authorize(http)
-        http = self.credentials.authorize(http)
-        transport.request(http, 'http://example.com')
+        authed_http = self.credentials.authorize(http)
+        authed_http = self.credentials.authorize(http)
```
dhermes: is-there-an

oauth2client/client.py
```diff
@@ -771,11 +771,11 @@ def _do_refresh_request(self, http):
         headers = self._generate_refresh_request_headers()
 
         logger.info('Refreshing access_token')
-        resp, content = transport.request(
+        response = transport._request(
```
dhermes: Comment for the entire commit where `transport._request` goes private:

Doesn't that break your rule of module-private for private-module?

oauth2client/transport/__init__.py
```diff
@@ -64,7 +64,27 @@ def make_authorized_http(
...
 
-def request(http_object, uri, method='GET', body=None, headers=None, **kwargs):
-    """Makes an HTTP request."""
-    return get_default_transport().request(
+def _request(http_object, uri, method='GET', body=None, headers=None,
+             **kwargs):
+    """Makes an HTTP request.
+
+    Args:
+        http_object: The transport-specific http object to be used to make
```
dhermes: Not that I care but http -> HTTP?

oauth2client/transport/__init__.py
```diff
@@ -64,7 +64,27 @@ def make_authorized_http(
...
+        uri: string, The URI to be requested.
+        method: string, The HTTP method to use for the request. Defaults
+            to 'GET'.
+        body: string, The payload / body in HTTP request. By default
+            there is no payload.
+        headers: dict, Key-value pairs of request headers. By default
+            there are no headers.
+
+    Returns:
+        An response object that will contain at least the following properties:
```
dhermes: An response, a response, am response, spam response

oauth2client/transport/__init__.py
```diff
@@ -64,7 +64,27 @@ def make_authorized_http(
...
+
+        * status: int, the HTTP status code.
+        * headers: dict, the HTTP response headers.
+        * data: bytes, the HTTP response body.
+    """
+    return get_default_transport()._request(
```
dhermes: I know Pylint would kill you here, using a private method of an object outside of one of that objects methods.

oauth2client/transport/httplib2.py
```diff
@@ -59,7 +59,6 @@ def __init__(self, http, credentials, refresh_status_codes):
         self.refresh_status_codes = refresh_status_codes
 
     def request(self, uri, method='GET', body=None, headers=None,
-                redirections=httplib2.DEFAULT_MAX_REDIRECTS,
```
dhermes, line 62: A triple plus

scripts/run_gce_system_tests.py
```diff
@@ -43,10 +43,10 @@ def test_token_info(self):
         query_params = {'access_token': credentials.access_token}
         token_uri = (oauth2client.GOOGLE_TOKEN_INFO_URI + '?' +
                      urllib.parse.urlencode(query_params))
-        response, content = transport.request(http, token_uri)
+        response = transport._request(http, token_uri)
```
dhermes, line 46: Wouldn't you rather use an `authed_http`?

scripts/run_system_tests.py
```diff
@@ -57,7 +57,7 @@ def _require_environ():
 
 def _check_user_info(credentials, expected_email):
     http = credentials.authorize(transport.get_http_object())
-    response, content = transport.request(http, USER_INFO)
+    response, content = transport._request(http, USER_INFO)
```
dhermes: Ditto on using an `authed_http`

tests/transport/test_transport_compliance.py
```diff
@@ -0,0 +1,153 @@
...
+    @app.route('/server_error')
+    def server_error():
+        return 'Error', http_client.INTERNAL_SERVER_ERROR
+
+    @app.route('/authorized')
+    def authorized():
+        """Echoes the request auth and user agent headers."""
+        return flask.jsonify({
+            'authorization': flask.request.headers.get('authorization'),
+            'user_agent': flask.request.headers.get('user-agent')
```
dhermes: Trailing comma, for diff's sake?

tests/transport/test_transport_compliance.py
```diff
@@ -0,0 +1,153 @@
...
+        """Returns UNAUTHORIZED if the request token doesn't contain
+        '-new', otherwise returns the auth and user agent headers."""
+        authorization = flask.request.headers.get('authorization', '')
+
+        if '-new' not in authorization:
+            return 'Not authorized', http_client.UNAUTHORIZED
+
+        return flask.jsonify({
+            'authorization': flask.request.headers.get('authorization'),
+            'user_agent': flask.request.headers.get('user-agent')
```
dhermes: Trailing comma, for diff's sake?

tests/transport/test_transport_compliance.py
```diff
@@ -0,0 +1,153 @@
...
+        headers[b'authorization'] = 'Bearer {}'.format(self.token)
+
+    def refresh(self, http):
+        self.token = '{}{}'.format(self.token, '-new')
+
+
+class TransportComplianceTests(object):
+    """Test base class to confirm that a transport satisfies the interface
+    expected by oauth2client."""
+    transport = None
```
dhermes, line 79: Yay! (I am happy because this used to be the `httplib2` transport.)

tests/transport/test_transport_compliance.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @pytest.fixture(autouse=True)
+    def testserver(self):
+        """Provides a test HTTP server that is automatically created before
+        a test and destoryed at the end. This server is running the test
```
dhermes: s/destoryed/destroyed

tests/transport/test_transport_compliance.py
```diff
@@ -0,0 +1,153 @@
...
+        authed_http = self.transport.make_authorized_http(
+            credentials, http, (http_client.UNAUTHORIZED,))
+
+        # Use transport._request instead of authed_http request because this
+        # has to work across multiple transports - the http object is
+        # essentially opaque to these tests.
+        response = self.transport._request(
+            authed_http, self.server.url + '/authorized')
+
+        content = json.loads(response.data.decode('utf-8'))
```
dhermes, line 129: There is `_helpers.to_bytes` for this decode.

tests/transport/test_transport_compliance.py
```diff
@@ -0,0 +1,153 @@
...
+
+        assert credentials.token == 'token'
+        assert content['authorization'] == 'Bearer {}'.format(
```
dhermes: Need a `b'...'` in this header?

dhermes (changes requested):

oauth2client/client.py
theacodes: My rule? Maybe you mean @nathanielmanistaatgoogle's rule. I'm not sure what you're referring to here.

oauth2client/client.py
theacodes: Done.

oauth2client/client.py
theacodes: done.

oauth2client/client.py
theacodes: Probably, but can we punt this?

oauth2client/transport/__init__.py
theacodes: done.

oauth2client/transport/__init__.py
theacodes: I can switch request back to public if you want?

oauth2client/transport/__init__.py
theacodes: Fixed

oauth2client/transport/httplib2.py
theacodes: ?

oauth2client/transport/httplib2.py
theacodes: It's specific to httplib2, urllib3 and requests will already be in the right format. ;)

theacodes: (docstring added)

scripts/run_gce_system_tests.py
theacodes: I'll eventually make this test use every transport, so no.

scripts/run_system_tests.py
theacodes: Same situation as above.

tests/test_client.py
theacodes: Haha, the intent of the original code is to double-wrap the object, but I messed that up. Fixed.

tests/transport/test_httplib2_transport.py
theacodes: @dhermes I used the same name you used in the test right above this one. :P

tests/transport/test_transport_compliance.py
theacodes: done.

tests/transport/test_transport_compliance.py
theacodes: done.

tests/transport/test_transport_compliance.py
theacodes: I don't want py.test to discover and run this. It should only run subclasses (which start with `Test`)

tests/transport/test_transport_compliance.py
theacodes: Arg. fixed.

tests/transport/test_transport_compliance.py
theacodes: because I forgot to remove it. Fixed.

tests/transport/test_transport_compliance.py
theacodes: So that's actually an outstanding question. There's a lot of errors that a transport can raise. Should we force the transports to raise a consistent exception? oauth2client basically never tried to catch any transport-level exceptions- preferring instead to let them work up the stack and have the user deal with them.

tests/transport/test_transport_compliance.py
theacodes: I'd prefer to reserve usage of that for when I don't know if the item is already a string or bytes. I control the type here, so I'd prefer not to have that indirection.

tests/transport/test_transport_compliance.py
theacodes: Nope, but you did make me rethink how I asserted this.

theacodes (commented):
Thanks for the review, @dhermes.

oauth2client/client.py
dhermes: Maybe I misremembered? In this library there have been discussion about not using non-public members of a module outside of that module.

tests/test_client.py
```diff
@@ -901,9 +901,10 @@ def test_recursive_authorize(self):
         token_response = {'access_token': '1/3w', 'expires_in': 3600}
         encoded_response = json.dumps(token_response).encode('utf-8')
         http = http_mock.HttpMock(data=encoded_response)
-        http = self.credentials.authorize(http)
-        http = self.credentials.authorize(http)
-        transport.request(http, 'http://example.com')
+        authed_http = self.credentials.authorize(http)
+        # Double-wrap
```
dhermes: This is necessary?

oauth2client/client.py
dhermes: Did you try and it caused issues? Are you worried about something? (i.e. why didn't you just put it in the "address review" commit?)

oauth2client/transport/__init__.py
dhermes: It seems necessary?

oauth2client/transport/httplib2.py
dhermes: A+++ (but in words). Like it's good.

oauth2client/transport/httplib2.py
dhermes: RE: `requests`, Am I missing something?

``` python
>>> import requests
>>> resp = requests.get('https://google.com')
>>> resp
<Response [200]>
>>> hasattr(resp, 'data')
False
>>> hasattr(resp, 'status')
False
>>> hasattr(resp, 'headers')
True
```

scripts/run_gce_system_tests.py
dhermes: Ah I see. (Just being allergic to the module-private helper.)

tests/transport/test_httplib2_transport.py
dhermes: Yes. I personally am OK with the name. My comment still is true.

tests/transport/test_transport_compliance.py
dhermes: Can't you just mark it with a special sentinel for `py.test` to go away? Seems silly to have to remember to name it right.

tests/transport/test_transport_compliance.py
dhermes: For the time being (i.e. for the transports you've implemented) can you provide a base type or a list of base types?

tests/transport/test_transport_compliance.py
dhermes: Meh.

tests/transport/test_transport_compliance.py
theacodes: Unfortunately no. The options are:
1. Name the class so that pytest doesn't treat it as a test
2. Name the module (test_transport_compliance) something else (transport_compliance, perhaps).

tests/transport/test_transport_compliance.py
theacodes: I'm not sure what you mean.

oauth2client/transport/httplib2.py
theacodes: Should've been more specific: the requests transport will use `response.raw` which is the underlying `urllib3.HTTPResponse` object.

jonparrott:
@dhermes it seems like the remainder of your concerns (save the one about exceptions) can be fixed by making `transport.request` public again. It's used throughout the package, but I don't not want users to use it directly. Are the concerns you have reason enough to make it public?

oauth2client/client.py
theacodes: I would just rather not fix this right now. This PR has changed enough without adding this in. We can file an issue, if you want.

(I know we generally have a rule of thumb of sprucing up every block of code we touch, but these transport changes are going to touch a majority of this project and I want to be focused on just making transport-related changes. We'll hopefully be combing over every module during the great refactor)

tests/transport/test_transport_compliance.py
dhermes: Ok. Sad face.

tests/test_client.py
theacodes: The comment or the double-wrapping? Either way - yes.

tests/transport/test_transport_compliance.py
dhermes: Instead of using `pytest.raises(Exception)` have `pytest.raises(self.transport.exceptions)` or maybe just `self.exceptions` (would be easier)

oauth2client/transport/httplib2.py
dhermes: Ah.

dhermes:
> Are the concerns you have reason enough to make it public?

Not sure.

jonparrott:
I'll let @nathanielmanistaatgoogle decide.

oauth2client/client.py
dhermes: OK. I'm still curious why this is different than the `d` --> `error_dict` rename (or similar).

tests/test_client.py
dhermes: Can you explain why it gets double wrapped in the comment?

tests/transport/test_transport_compliance.py
theacodes: Done, although this is not surfaced to the top-level `oauth2client.transport`.

tests/test_client.py
theacodes: Done.

oauth2client/client.py
theacodes: Because this issue is all over `client.py`, and I'd prefer to fix all occurrences of it at once instead of doing piecemeal changes which lead to overall inconsistency.

tests/transport/test_transport_compliance.py
dhermes: SGTM

oauth2client/client.py
dhermes: Now we're talking! For the first time I understand. Filed #656 for you.

oauth2client/transport/httplib2.py
```diff
@@ -29,6 +29,9 @@
 _MAX_REFRESH_ATTEMPTS = 2
 
 
+exceptions = (httplib2.HttpLib2Error,)
```
dhermes: I immediately regret requesting it go in the transport. Can you instead just define `exceptions` as `None` on `TransportComplianceTests` and make the subclasses update the value?

This is a burden that shouldn't be public.

(I promise this is my last request.)

dhermes (changes requested):

oauth2client/transport/httplib2.py
theacodes: Done, I'm 100% for this.

jonparrott:
@nathanielmanistaatgoogle can you weigh in on the contentious topic on making `transport.request` private?

jonparrott:
@dhermes I've made request public again. I'm going to try to make the entire transport module private.

dhermes (approved):

PR #660, opened 2016-09-27
Make transport private, remove usage of get_http_object

theacodes:
It's broken into 3 separate commits on purpose. The first makes transport private and removes get_http_object, the next two fix everything that breaks when that happens.


oauth2client/_transport/__init__.py
```diff
@@ -16,29 +16,49 @@
...
+    from oauth2client._transport import httplib2
+except ImportError:  # pragma: NO COVER
+    httplib2 = None
+try:
+    from oauth2client._transport import urllib3
+except ImportError:  # pragma: NO COVER
+    urllib3 = None
+
+# The available transports, in order of priority.
+_TRANSPORTS = tuple(filter(None, (urllib3, httplib2,)))
```
dhermes: :-1: to the `filter`. Why not just conditionally `_TRANSPORTS += (foo,)` when the imports succeed?

oauth2client/_transport/__init__.py
```diff
@@ -16,29 +16,49 @@
...
-
-
+try:
+    from oauth2client._transport import httplib2
+except ImportError:  # pragma: NO COVER
+    httplib2 = None
+try:
+    from oauth2client._transport import urllib3
+except ImportError:  # pragma: NO COVER
+    urllib3 = None
```
dhermes: It'd be nice to do something else here, maybe even just punting the imports to their own file, and also do the `if not _TRANSPORTS` there as well.

oauth2client/_transport/__init__.py
```diff
@@ -16,29 +16,49 @@
...
-    """Gets the global default transport."""
-    return _SETTINGS.default_transport
+    Returns:
+        module: A transport module.
 
+    Raises:
+        ValueError: if no transport can be found for the object.
+    """
+    for transport in _TRANSPORTS:
+        if isinstance(http_object, transport.HTTP_OBJECT_CLASSES):
```
dhermes: 1. Isn't `HTTP_OBJECT_CLASSES` new?
2. Couldn't you have a mutable dictionary in this module and let implementations add classes to it? Then you could just use `transports_dict[http_object.__class__]`.

oauth2client/_transport/__init__.py
```diff
@@ -86,5 +106,11 @@ def request(http_object, uri, method='GET', body=None, headers=None,
         * headers: dict, the HTTP response headers.
         * data: bytes, the HTTP response body.
     """
-    return get_default_transport().request(
-        http_object, uri, method=method, body=body, headers=headers, **kwargs)
+    if http_object is None:
+        http_object = _get_preferred_http_object()
+
+    transport = _transport_for_http_object(http_object)
```
dhermes: Could be a little more efficient:

``` python
if http_object is None:
    http_object = _get_preferred_http_object()
    transport = _PREFERRED_TRANSPORT
else:
    transport = _transport_for_http_object(http_object)
```

oauth2client/_transport/httplib2.py
```diff
@@ -157,7 +170,18 @@ def request(http, uri, method='GET', body=None, headers=None, **kwargs):
...
+    # it.
+    original_timeout = http.timeout
+    if timeout is not None:
+        http.timeout = timeout
+
+    try:
+        response, data = http.request(
+            uri, method=method, body=body, headers=headers, **kwargs)
+    finally:
+        http.timeout = original_timeout
```
dhermes, line 185: Aren't you a fan of using `contextlib.contextmanager` in these situations?

oauth2client/_transport/httplib2.py
```diff
@@ -112,18 +114,28 @@ def request(self, uri, method='GET', body=None, headers=None,
...
+        return self.http.timeout
+
+    @timeout.setter
+    def timeout(self, value):
+        """Proxy to self.http."""
+        self.http.timeout = value
+
 
-def make_authorized_http(credentials, http, refresh_status_codes):
+def make_authorized_http(http, credentials, refresh_status_codes):
```
dhermes, line 128: Why the swap?

oauth2client/_transport/urllib3.py
```diff
@@ -163,13 +166,19 @@ def request(http_object, uri, method='GET', body=None, headers=None, **kwargs):
...
-        headers (Mapping): Key-value pairs of request headers.
+        body (bytes): The payload / body in HTTP request.
+        headers (Mapping): Request headers.
     Returns:
         urllib3.response.HTTPResponse: The HTTP response.
     """
+
+    # Urllib3 uses a sentinel default value for timeout, so only set it if
+    # specified.
+    if timeout is not None:
```
dhermes, line 183: Do you not want to use a sentinel as well? Like what if a user or internal caller pass an explicit `None`?

tests/transport/test_transport.py
```diff
@@ -12,16 +12,31 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import unittest
+import httplib2
+import pytest
+import urllib3
```
dhermes: No `__future__ import absolute_import`?

tests/transport/test_transport.py
```diff
@@ -12,16 +12,31 @@
...
-import mock
+from oauth2client import _transport
+import oauth2client._transport.httplib2
+import oauth2client._transport.urllib3
 
-from oauth2client import transport
 
+def test__get_preferred_http_object():
+    assert isinstance(
+        _transport._get_preferred_http_object(), urllib3.PoolManager)
```
dhermes: Shouldn't you patch here to make sure `urllib3` is the default?

oauth2client/client.py
```diff
@@ -2056,10 +2048,7 @@ def step2_exchange(self, code=None, http=None, device_flow_info=None):
         if self.user_agent is not None:
             headers['user-agent'] = self.user_agent
 
-        if http is None:
-            http = transport.get_http_object()
```
dhermes, line 2060: What happens if `http` is `None` here? Just an error? Should we add a test for this?

oauth2client/client.py
```diff
@@ -984,10 +982,10 @@ def _detect_gce_environment():
     #       could lead to false negatives in the event that we are on GCE, but
     #       the metadata resolution was particularly slow. The latter case is
     #       "unlikely".
-    http = transport.get_http_object(timeout=GCE_METADATA_TIMEOUT)
     try:
-        response = transport.request(
-            http, _GCE_METADATA_URI, headers=_GCE_HEADERS)
+        response = _transport.request(
+            None, _GCE_METADATA_URI, headers=_GCE_HEADERS,
```
dhermes, line 987: Whoa. `None`? Huh?

oauth2client/contrib/appengine.py
```diff
@@ -727,19 +726,14 @@ def authorize_url(self):
...
         from within an @oauth_aware decorated method where has_credentials()
         returns True.
-
-        Args:
-            *args: Positional arguments passed to httplib2.Http constructor.
-            **kwargs: Positional arguments passed to httplib2.Http constructor.
         """
-        return self.credentials.authorize(
-            transport.get_http_object(*args, **kwargs))
+        return self.credentials.authorize(http)
```
dhermes, line 736: :+1: Here

oauth2client/contrib/appengine.py
```diff
@@ -727,19 +726,14 @@ def authorize_url(self):
...
 
-    def http(self, *args, **kwargs):
+    def http(self, http):
         """Returns an authorized http instance.
 
         Must only be called from within an @oauth_required decorated method, or
         from within an @oauth_aware decorated method where has_credentials()
         returns True.
-
-        Args:
```
dhermes, line 737: Shouldn't `Args` be put back?

oauth2client/contrib/flask_util.py
```diff
@@ -536,20 +535,15 @@ def required_wrapper(*args, **kwargs):
...
             return curry_wrapper
 
-    def http(self, *args, **kwargs):
+    def http(self, http):
         """Returns an authorized http instance.
 
         Can only be called if there are valid credentials for the user, such
         as inside of a view that is decorated with @required.
 
-        Args:
```
dhermes, line 545: Shouldn't `Args` be put back?

tests/conftest.py
```diff
@@ -17,11 +17,19 @@
...
+from tests import http_mock
 
 
 logging.basicConfig(level=logging.INFO)
 
 
+# Make HttpMock instances use the httplib2 transport.
+httplib2.HTTP_OBJECT_CLASSES = (
+    (http_mock.HttpMock, http_mock.HttpMockSequence) +
+    httplib2.HTTP_OBJECT_CLASSES)
```
dhermes, line 30: This feels icky. Why not just do it with `mock.patch` when needed?

oauth2client/contrib/django_util/__init__.py
```diff
@@ -481,9 +480,8 @@ def credentials(self):
         """Gets the authorized credentials for this flow, if they exist."""
         return _credentials_from_request(self.request)
 
-    @property
```
dhermes, line 484: Not sure this will fly.

oauth2client/_transport/__init__.py
theacodes: Not sure what you mean?

oauth2client/_transport/__init__.py
theacodes: 1. Yes.
2. I don't like mutable globals.

oauth2client/_transport/httplib2.py
theacodes: Seemed more appropriate.

oauth2client/_transport/httplib2.py
theacodes: Sure, but there isn't a context manager specifically for this (other than `mock.patch` which shouldn't be used in real code. :P)

oauth2client/_transport/urllib3.py
theacodes: I don't think so. The options are:
1. Pass an integer - set the timeout.
2. Pass None - accept the default timeout.

Transport doesn't provide the idea of an `infinite timeout`. We could, but I'm on the fence about it. (there's no clear use case for it in this library)

tests/transport/test_transport.py
theacodes: Not needed here.

tests/transport/test_transport.py
theacodes: No? This tests makes sure that our preferred http object is urllib3 even when all transports are available.

oauth2client/client.py
theacodes: See the logic in `_transport.request`, `http` can now explicitly be `None` and the preferred transport will be used. This is reflected in the docstring.

oauth2client/client.py
theacodes: Nope - see comment above - the preferred transport will be used.

tests/conftest.py
theacodes: Because there are so many uses of HttpMock.

oauth2client/contrib/django_util/__init__.py
theacodes: @waprin: thoughts?

oauth2client/_transport/urllib3.py
theacodes: I will update the docstring to reflect that behavior.

oauth2client/_transport/__init__.py
dhermes: Create a standalone module with:

``` python
_TRANSPORTS = ()

try:
    from oauth2client._transport import urllib3
    _TRANSPORTS += (urllib3,)
except ImportError:  # pragma: NO COVER
    urllib3 = None

try:
    from oauth2client._transport import httplib2
    _TRANSPORTS += (httplib2,)
except ImportError:  # pragma: NO COVER
    httplib2 = None

if not _TRANSPORTS:
    raise ImportError('Oops')
```

oauth2client/_transport/__init__.py
dhermes: > I don't like mutable globals.

But do you like `O(n)` lookups consistently throughout the life of the application instead of `O(1)` lookup?

oauth2client/_transport/httplib2.py
dhermes: Yeah true

oauth2client/_transport/httplib2.py
dhermes: Right. Could write a decorator, but probably not worth it.

tests/transport/test_transport.py
dhermes: OK.

oauth2client/client.py
dhermes: :cry: That means we're building things in our castle.

tests/conftest.py
dhermes: Does `py.test` respect `setUpModule` and `tearDownModule`?

oauth2client/_transport/__init__.py
theacodes: Done

oauth2client/_transport/__init__.py
theacodes: Done

oauth2client/_transport/__init__.py
theacodes: Discussed offline - a map won't work because of the use of `isinstance`. After exhausting alternatives, we're going to stick with this route and add a strong warning for future changes.

oauth2client/_transport/__init__.py
theacodes: Done

oauth2client/_transport/urllib3.py
theacodes: Done.

oauth2client/contrib/flask_util.py
theacodes: Done

tests/conftest.py
theacodes: (resolved offline.. I think?)

dhermes:
LGTM

jonparrott:
Merging with admission of the following:

1. I really want to remove any uses of patching `_get_preferred_http_object` except for the gce check. HTTP should be required *everywhere*.
2. The outcome of the conversation over at https://github.com/jonparrott/goth/issues/1 will likely have some bearing as to where I take this.

PR #662, opened 2016-10-06
Try to revoke token with POST when getting a 405

LiraNuna:
The OAuth spec does not specify the HTTP verb explicitly but it does hint that POST is the correct verb.
When using the client library with other OAuth services that implement revocation token via a POST, revoking the token will fail.

This commit adds the ability to re-try the revocation process if we get a 405 with the POST verb.


LiraNuna:
I signed it!

jonparrott:
Hi @LiraNuna! This change looks good, but you'll need to add another test to cover this so that travis passed.

LiraNuna:
@jonparrott Just noticed the failure as well - I'll get the tests soon.

jonparrott:
Sounds good, just ping me when you're ready.

(Also, small world - I remember you from NDS dev stuff. 😄 )

LiraNuna:
@jonparrott Tests fixed - I had to reshuffle stuff around to get the desired behavior correctly.

jonparrott:
Thanks for the contribution, @LiraNuna!

LiraNuna:
@jonparrott how soon should I expect this to show up on PyPI? I've noticed the current code is a bit different than `3.0.0`

jonparrott:
I'm not sure.

PR #670, opened 2016-11-03
Prettify our JSON dumps

### LINKED ISSUE #669: Prettify stored credentials
Currently, credentials are stored as minimal JSON. This makes it a challenge when a developer needs to peek at credentials stored to a file in order to find their project_id, client_id, refresh_token or some other detail.

We should be able to pretty print the JSON before storage to make it easier for human eyes to read.

jay0lee:
Fixes #669.


jonparrott:
Hmm, I'm apprehensive about this change as it causes the size of our json dumps to grow significantly.

dhermes:
I'm also 👎  on this

jay0lee:
For my app (which uses about 28 scopes making credentials pretty large), a file created without the patch is 3,926 bytes. A file created with the patch is 4,404 bytes.

So growth is ~12.1%. What if we made this a non-default argument to to_json()? Would that be preferable?

FWIW, service account credentials downloaded from console.developers.google.com and generated via API are pretty printed.

Issue #672, opened 2016-11-11
Unpickle problems when upgrading from Python 2.7 to 3.5


chripede:
I've been using the Django contrib CredentialsField to store authentications in my database. This all happened in Python 2.7

I'm trying to upgrade to 3.5, but the pickled CredentialsField is giving me some problems. It seems impossible to unpickle an Python 2.7 object in Python 3 that contains both strings and a datetime. See http://bugs.python.org/issue22005

Is there any "best practice" way to do this, or will I need to convert everything to JSON and the pickle it again?

jonparrott:
Unfortunately I don't know if there's anything we can do to solve the cross-major-version datetime pickle problem in this library. If I had to recommend anything, it would be to use json to convert.

chripede:
@jonparrott how would you feel about a pull request that replaces pickle with jsonpickle? I see the views.py in the django contrib started using it.

In that case all that's needed is loading and re-saving every object to get them converted and future proofed.

jonparrott:
I'd prefer not to change anything major like that at the moment. Though, @bjwatson might disagree.

bjwatson:
I second what @jonparrott says. I'd prefer to do major changes to https://github.com/GoogleCloudPlatform/google-auth-library-python instead, unless something is really broken in this repo.

chripede:
I'd argue that this isn't a major change.

1. It's in the django contrib module only
2. It will still work with pickled models
3. If jsonpickle is not installed it can just fall back to pickle, it could even be a config switch.

jonparrott:
If the change only applies to contrib.django I'm okay with it.

bjwatson:
I defer to your judgement @jonparrott.

waprin:
I already made jsonpickle a dependency of the module so I am in favor of this change, thanks for the issue and kicking off the PR.

theacodes:
Thank you for creating this issue, however, this project is deprecatedand we will only be addressing critical security issues. You can read moreabout this deprecation [here](http://google-auth.readthedocs.io/en/latest/oauth2client-deprecation.html).

If you need support or help using this library, we recommend that you ask yourquestion on StackOverflow.

If you still think this issue is relevant and should be addressed, pleasecomment and let us know!

PR #676, opened 2016-11-16
Use jsonpickle in django contrib

chripede:
This change will try to import jsonpickle and use that to pickle and unpickle the token stored in database.

Discussed in #672

### DIFF
--- a/oauth2client/contrib/devshell.py
+++ b/oauth2client/contrib/devshell.py
@@ -37,6 +37,7 @@ class CommunicationError(Error):
 class NoDevshellServer(Error):
     """Error when no Developer Shell server can be contacted."""
 
+
 # The request for credential information to the Developer Shell client socket
 # is always an empty PBLite-formatted JSON object, so just define it as a
 # constant.
--- a/oauth2client/contrib/django_util/models.py
+++ b/oauth2client/contrib/django_util/models.py
@@ -19,6 +19,7 @@ import pickle
 
 from django.db import models
 from django.utils import encoding
+import jsonpickle
 
 import oauth2client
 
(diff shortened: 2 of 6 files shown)


oauth2client/contrib/django_util/models.py
```diff
@@ -16,6 +16,12 @@
 
 import base64
 import pickle
+import sys
+
+try:
+    import jsonpickle
+except ImportError:
+    pass
```
waprin: I think I already made jsonpickle a dependency of this contrib package? Have you tested it with regular pickle? Think there are some subtle issues...maybe best to just require jsonpickle.

chripede: jsonpickle is not installed with the package, but I'd prefer to do that if it's okay.

waprin: It's used in `contrib.django_util.views.py`. It's not installed with oauth2client, but necessary to use the django contrib.

What we really want is something alone the lines of making the Django contrib package an optional dependency, which brings in jsonpickle with it. For the purpose of this PR I would just assume it's installed.

chripede: That's great. Cleaned up the code.

waprin:
Looks like you have some flake errors

chripede:
Not really sure why, as I never touched the files that fails. I've fixed them and everything seems fine now

oauth2client/contrib/devshell.py
```diff
@@ -37,6 +37,7 @@ class CommunicationError(Error):
 class NoDevshellServer(Error):
     """Error when no Developer Shell server can be contacted."""
 
+
```
waprin, line 40: was this needed to pass flake?

tests/test_service_account.py
```diff
@@ -369,6 +369,7 @@ def test_access_token(self, utcnow):
 
         self.assertEqual(credentials.access_token, token2)
 
+
```
waprin, line 372: was this needed to pass flake?

waprin:
Ok I see, yeah I don't think we pin the linter so sometimes lint errors pop up underneath us.

waprin:
Thanks for PR.

Issue #678, opened 2016-11-21
Store the original id_token JWT in OAuth2Credentials


clancychilds:
Currently, the credentials that are returned as a JWT in an identity token are extracted and stored in their extracted state, but the JWT itself is not retained. In some flows (like using the [Auth0 Delegation endpoint](https://auth0.com/docs/api/authentication#!#post--delegation)) having the original JWT is required.
I would like to add the full JWT to the id_token dict in OAuth2Credentials. Probably as 'jwt'.

jonparrott:
I'm open to this in general, but we're avoiding making large changes to this library unless totally necessary. Are you willing to send a PR to see what the impact will be?

clancychilds:
Yes. Working on one now.

clancychilds:
As discussed, I have submitted a PR that would address this. 
https://github.com/google/oauth2client/issues/680

PR #679, opened 2016-11-22
Store original encoded and signed identity JWT in OAuth2Credentials

clancychilds:
For some authorization flows, including the Auth0 Delegation flow, the endpoint requires the original signed identity JWT to be sent. To support this, it is necessary to store the JWT alongside the decoded content.

This PR adds `id_token_jwt` as an element in the OAuth2Credentials. It also adds the logic to populate that element in parts of the flows that receive the identity token.

Resolves: #678


clancychilds:
Argh. Closing this because I had the wrong credentials on the commits.

PR #680, opened 2016-11-23
Store original encoded and signed identity JWT in OAuth2Credentials

### LINKED ISSUE #678: Store the original id_token JWT in OAuth2Credentials

clancychilds:
For some authorization flows, including the Auth0 Delegation flow, the endpoint requires the original signed identity JWT to be sent. To support this, it is necessary to store the JWT alongside the decoded content.

This PR adds id_token_jwt as an element in the OAuth2Credentials. It also adds the logic to populate that element in parts of the flows that receive the identity token.

(This is the second PR for this as a previous PR had bad commit email addresses and was closed.)

Resolves: #678

### DIFF
--- a/oauth2client/client.py
+++ b/oauth2client/client.py
@@ -451,7 +451,7 @@ class OAuth2Credentials(Credentials):
     def __init__(self, access_token, client_id, client_secret, refresh_token,
                  token_expiry, token_uri, user_agent, revoke_uri=None,
                  id_token=None, token_response=None, scopes=None,
-                 token_info_uri=None):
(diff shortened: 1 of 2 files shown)


jonparrott:
Thanks for this, @clancychilds.

PR #683, opened 2016-12-06
Remove b64 padding from PKCE values, per RFC7636

### LINKED ISSUE #674: Incorrect Base64 encoding of PKCE challenge
Applies to: 4.0.0

In `_pkce.py` we are currently seeing the following to generate the code challenge from the code verifier:

`base64.urlsafe_b64encode(hashlib.sha256(verifier).digest())`

Unfortunately, as per [RFC 7636](https://tools.ietf.org/html/rfc7636#section-4.2) the generated challenge is not valid, as [it can and will](https://docs.python.org/3/library/base64.html#base64.urlsafe_b64encode) contain a `=` padding character, while the allowed characters are:

```
code-challenge = 43*128unreserved
unreserved = ALPHA / DIGIT / "-" / "." / "_" / "~"
ALPHA = %x41-5A / %x61-7A
DIGIT = %x30-39
```

It has been observed that Google's OAuth endpoints reject a code challenge from `oauthclient2` 4.0.0 with an error message:

```
Error: invalid_request

Code Challenge must be base64 encoded.
```

A fix (briefly tested) would simply remove the padding, e.g.

```
base64.urlsafe_b64encode(hashlib.sha256(verifier).digest()).rstrip(b'=')
```

bjmc:
Fixes #674

### DIFF
--- a/oauth2client/_pkce.py
+++ b/oauth2client/_pkce.py
@@ -38,7 +38,7 @@ def code_verifier(n_bytes=64):
     Returns:
         Bytestring, representing urlsafe base64-encoded random data.
     """
-    verifier = base64.urlsafe_b64encode(os.urandom(n_bytes))
+    verifier = base64.urlsafe_b64encode(os.urandom(n_bytes)).rstrip(b'=')
     # https://tools.ietf.org/html/rfc7636#section-4.1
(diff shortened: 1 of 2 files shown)


jonparrott:
Thanks, @bjmc.

PR #691, opened 2017-01-08
master

TZubiri:
Making scope POST parameter optional when using flow.step2_exchange.


TZubiri:
Silly oversight of mine is causing checks to fail. I hope it's ok if I use Travis to perform tests instead of setting a local instance myself. Let me know if it's bad ettiquette.
Will commit again and hopefully it'll pass.

jonparrott:
As discussed over at #692, I'd rather not make any changes to this library, especially those that change behavior.

PR #699, opened 2017-02-15
Add support for reauth in oauth2

svalentin:
When the library refreshes an access token, if it fails because a rapt
token is required or invalid, we try to refresh the rapt token and then
retry the access token refresh.

To refresh the rapt token, we interact with the rapt service to get one.
This implies getting an access token for just REAUTH_SCOPE (which
shouldn't require a rapt token) then passing that to the reauth service
and answering challenges. Challenges could either be PASSWORD or
SECURITY_KEY. Different accounts can have different policies. Some
accounts might need only password, some can require both.
Once all challenges are answered, we get back a rapt token, which we use
to get an access token with all scopes, and we return the request as if
the original access token refresh completed successfully.

Access tokens can have a lifespan shorter than a rapt token, and getting
a rapt token requires going through the whole reauth flow, where the
user enters their password/gnubby. An example is 1h expiry time for
access token, and 24h for rapt token.

Communication with the gnubby is done via the pyu2f libary (all python)
https://github.com/google/pyu2f/

The idea is clients (like gcloud) could come with that library included,
but if it's not, we can fail elegantly and just tell users we couldn't
refresh access token.
We can also include the library in oauth2 if it's appropriate.


jonparrott:
In general we will not be making any more releases of this library unless there is a critical bug. In short- we aren't adding new features to this library.

[google-auth](google-auth.readthedocs.io) will be the permanent replacement for this library once it reaches 1.0. Since this feature is focused on user auth (3lo), I'd recommend contributing it to oauthlib so that we can leverage google-auth's integration with oauthlib.

jonparrott:
@svalentin I'm leaving this open, but we still need a lot of work on this as we agreed when we discussed deprecation. Let me know if you need any help.

theacodes:
Thank you for creating this pull request, however, this project is deprecated
and we will only be accepting PRs to address critical security issues. If you
still think this PR should be merged, please comment and let us know!

PR #703, opened 2017-03-23
Custom Key can be specified for StorageByKeyName

sgabello:
The default storage class uses StorageByKeyName with no option to specify a custom KeyName (it defaults to the userId). This cause problems when multiple decorators are created in order to deal with different OAuth providers: without separate KeyNames, all the decorators would share the same entity. By providing a custom storage_key for each decorator, collision is avoided.


sgabello:
I signed it!

jonparrott:
I'm closing this PR because this project is now deprecated. Please see the [release notes for v4.1.0](https://github.com/google/oauth2client/releases/tag/v4.1.0) for more details.

PR #704, opened 2017-03-24
Allow customizing the GCE metadata service address via an env var.

craigcitro:
The goal here is to make it possible for a user of a binary that depends on
this library (eg the google cloud SDK) to be able to customize where it looks
for the GCE metadata service. (An adventurous user can already customize the
GCE metadata service location via the existing global vars in this library.)

The only bit of awkwardness here is really the test: since this is a top-level
statement, reloading is the only way to ensure it works.

PTAL @jonparrott @dhermes


craigcitro:
(Travis issues seem pre-existing, but I think that's well-known. 😛 )

craigcitro:
(small update squashed in: `reload` is apparently no longer top-level in python3!)

craigcitro:
(and another for some 2-space vs. 4-space indentation)

craigcitro:
(rebased so i can feel good about seeing green)

PR #710, opened 2017-04-19
Escape callback error code in flask_util

### LINKED ISSUE #709: oauth2callback endpoint provided by flask extension has an XSS vulnerability.
Due to a lack of sanitisation in the error parameter, it is possible for arbitrary code to be reflected on the oauth2callback endpoint (this vulnerability is referenced as OTG-INPVAL-001 in the OWASP Testing Guide). 

Using the default setup of the contrib flask extension:
```oauth2 = oauth2client.contrib.flask_util.UserOAuth2(app)```

The /oauth2callback endpoint then allows for an XSS attack like so:
```
/oauth2callback?error=a%3C!%27/*!%22/*!\%27/*\%22/*--!%3E%3C/Title/%3C/script/%3E%3CInput/Autofocus/%0D*/Onfocus=confirm(document.domain)//%3E%3Csvg%3E
```

We had this reported as a security issue on all our services that make use of oauth2client, requiring an urgent switch to a different authentication method.

theacodes:
Resolves #709

### DIFF
--- a/oauth2client/contrib/flask_util.py
+++ b/oauth2client/contrib/flask_util.py
@@ -176,6 +176,7 @@ try:
     from flask import request
     from flask import session
     from flask import url_for
+    import markupsafe
 except ImportError:  # pragma: NO COVER
     raise ImportError('The flask utilities require flask 0.9 or newer.')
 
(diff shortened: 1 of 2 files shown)


jonparrott:
@jkrukoff can you confirm this fixes the issue?

dhermes (approved):

PR #713, opened 2017-05-09
Fix system tests

theacodes:


tox.ini
```diff
@@ -71,7 +71,7 @@ deps =
     pycrypto>=2.6
     cryptography>=1.0
     pyopenssl>=0.14
-passenv = GOOGLE_* OAUTH2CLIENT_* TRAVIS*
+passenv = GOOGLE_* OAUTH2CLIENT_* TRAVIS* encrypted_*
```
dhermes, line 74: Seems you've obsoleted the `OAUTH2CLIENT_*` vars?

theacodes: No, I just changed the decrypt command for that one file.

PR #731, opened 2017-11-23
Don't use socket.recv with socket.MSG_WAITALL

petrhosek:
socket.MSG_WAITALL is not available on Windows so the call to
socket.recv with this flag will fail there. Provide a helper method
that tries to receive an exact number of bytes and use it instead.


theacodes:
Thank you for creating this pull request, however, this project is deprecated
and we will only be accepting PRs to address critical security issues. If you
still think this PR should be merged, please comment and let us know!

PR #746, opened 2018-10-31
Alternative library that is closer to oauth2client API

timworx:
**Note**: oauth2client is now deprecated. As such, it is unlikely that we will
review or merge to your pull request. We recommend you use
[google-auth](https://google-auth.readthedocs.io) and [oauthlib](http://oauthlib.readthedocs.io/).

I've recently removed oauth2client from a project. It was a struggle to determine how to go about it without having to do a ton of extra work. It took a bit of digging to find `google-auth-oauthlib` and ultimately it drops in fairly nicely, with a lot less work than using google-auth and oauthlib directly .

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ This is a client library for accessing resources protected by OAuth 2.0.
 
 **Note**: oauth2client is now deprecated. No more features will be added to the
 libraries and the core team is turning down support. We recommend you use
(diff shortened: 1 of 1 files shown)


timworx:
i signed it!

theacodes:
Thanks for your PR, but as this library is deprecated we won't be merging any PRs here. It might make sense to send a PR instead to google-auth's documentation page about oauth2client deprecation to better call out google-auth-oauthlib?

